PR #25, opened 2020-06-29
Include "Go to Definition" keyboard shortcut

### LINKED ISSUE #23: Idea: keyboard shortcut as Cmd+Click alternative
I think it would be really cool to have a keyboard shortcut to open links instead of Cmd+Click. Like many VSCode users, I don't leave my keyboard that much (vim compatibility) so the 'click' interaction in the middle of typing is awkward. 

For now I can use the normal file jump commands but if I have many files named similarly, a keyboard shortcut would be slicker.

Georift:
Also has a quick one-liner about changing the mapping.

Closes #23


Georift:
Oh woops mispelt cursor. One second

jevakallio:
Thanks @Georift! 🙏

PR #31, opened 2020-06-29
Fix contribution-guide links

coffenbacher:


jevakallio:
Thanks again @coffenbacher! 

As you can see, I wrote this documentation in a mad rush as users were already onboarding, so it's full of small mistakes like this. Appreciate the fixes!

Issue #30, opened 2020-06-29
Idea: use built-in Git extension for now


coffenbacher:
This still isn't ideal [1], but IMO a slightly slicker default than requiring the Git Automator extension at the moment is:

.vscode/settings.json
```
  "git.enableSmartCommit": true,
  "git.postCommitCommand": "sync",
```

To save & sync to Github or any remote, it's just `Git: Commit All`. This doesn't require the additional extension and automatically does the sync.

[1] (I think it would be a cooler default and more like Roam to do this automatically when saving files, but I didn't immediately see how to do that...)

jevakallio:
@coffenbacher that's a great idea! I was looking for commands like these in the default integration and couldn't find them, so I went with Git Automator.

I'll take this to use in my local workspace, test it out for a day or so, and if it works implement it into the template!

jevakallio:
Update: Tested this locally, and it's nice. 

The downside of this approach is that it's not possible to automatically set up a keybinding for `Git: Commit All`, so there's one more setup step for the users. But, still worth it IMO.

One way around this would be to add the key bindings to foam-vscode...

coffenbacher:
Great! Nice turnaround time on testing :rocket: 

> One way around this would be to add the key bindings to foam-vscode

That seems like a good approach from a user's perspective. I don't think the average user needs to know how nless they want to move away from the default.

I think I am becoming a big enough Foam fan that I'll try to get my feet wet with PRs here in a bit, thanks!

coffenbacher:
https://github.com/foambubble/foam/pull/32
https://github.com/foambubble/foam-template/pull/6

jevakallio:
Thank you very much @coffenbacher! Merged both 🥳

PR #32, opened 2020-06-29
Use built-in git instead of Git Automator extension

coffenbacher:
This commit updates the git instructions
To use the built-in VS Code git integration instead of Git Automator.

- Automatically stage all changes
  "git.enableSmartCommit": true,

- Automatically push & pull changes to the remote after committing
  "git.postCommitCommand": "sync"

Updates the settings in this project to dogfood the new defaults.

This feature could be improved further
By setting a default key-binding to Git: Commit All,
And/or by automatically committing on save or focus change
For the most SaaS-like experience.

Part of issue #30 along with a PR to foambubble/foam-template.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -11,10 +11,6 @@
     // GitLens for seeing version history inline
     "eamodio.gitlens",
 
-    // Git Automator for quick commit and push
-    // (Could be improved)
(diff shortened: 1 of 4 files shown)


jevakallio:
LGTM! :shipit:

PR #35, opened 2020-06-30
[feat] add chinese README

xieyezi:
Hi😄   , thanks you this inspire project,it's great!!
There, I translate English README to Chinese,so that more chinese developer to use Foam !


jevakallio:
Thanks for the translation! I'd love for Foam to be internationalised, but since the project is so new and moving so fast, realistically it's too early for us to take on translations I can't myself update and maintain as things go out of sync.

Thanks again! If you want to publish a Foam starter guide in Chinese in your own repo, I'd be happy to link to it!

PR #82, opened 2020-07-09
link ref definition proposal

jojanaho:
Proposal to improve upon link reference definitions + related changes in other doc files


docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+- This will work on all platforms.
+
+**Cons:**
+
+- It will break the Markdown Notes [[backlinking]] support
+- Less convenient to write
+
+### Finding certain words clutter the VS Code search results
+
+Since link reference definitions have `[//begin]` and `[//end]` guards with explanatory text that use certain words, these words (like "generate") appear in VS Code search results if you happen to search matching strings from the workspace.
```
jevakallio, line 48: Good point! These were a hack to ensure I didn't remove any user-generated content in the MVP. The comment blocks are not necessary and must go.

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Improvement Proposal
+
+Problem space in essence:
```
jevakallio, line 52: Stellar job summarising this!

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+    - get out of date if user uses a tool that doesn't autogenerate them
+    - may clutter the search results
+- During build-time (when converting markdown to html for publishing purposes)
+  - link reference definitions are needed, if the files are published via such tools (or to such platforms) that don't understand wikilinks
+  - link reference definitions might have to be in different formats depending on the publish target (e.g. Github pages vs Github UI)
+
+The potential solution:
+
+- For edit-time
+  - Make edit-time link reference definition generation optional (via user settings and maybe even "off" by default, so that the user wouldn't typically see the auto-generated section in the file at all if it's really not needed)
```
jevakallio, line 67: If there has to be a default, I'd make it "on, but generating .md". There are two sets of tools we want to support:
- Recommended extensions
- The much larger subset of all markdown tools

Agree that we should be able to turn this off, though.

```suggestion
  - Make edit-time link reference definition generation optional via user settings. They should be on by default, and generating valid markdown links with a relative path to a `.md` file.
```

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+  - Make format of the link reference definition configurable (whether to include '.md' or not)
```
jevakallio, line 68: 👍

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+  - Out of recommended extensions, currently only "markdown links" doesn't support them (?). However even its [code](https://github.com/tchayen/markdown-links/blob/master/src/parsing.ts#L25) seems to include wikilink parser, so it might just be a bug?
+  - If user really want's to use an extension that requires "plain links", they can set the setting "on" and work as in current Foam
```
jevakallio, line 70: Is "plain links" links without .md?

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+- For build-time
+  - Have publish targets defined in settings, that support both turning the link reference definitions on/off and defining their format (.md or not)
+  - With Foam repo, just use edit-time link reference definitions with '.md' extension - this makes the links work in the Github UI
```
jevakallio, line 73: 👍

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+  - Have publish target defined for Github pages, that doesn't use '.md' extension, but still has the link reference definitions. Generate the output into gh-pages branch (or separate repo) with automation.
```
jevakallio, line 74: They should either go to `gh-pages` or `master/docs`, since those are the two supported GH pages default sources.

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+    - This naturally requires first removing the existing link reference definitions during the build
+- Other
+  - To clean up the search results, remove link reference definition section guards (assuming that these are not defined by the markdown spec). Use unifiedjs parse trees to identify if there's missing (or surplus) definitions (check if they are identified properly by the library), and just add the needed definitions to the bottom of the file (without guards) AND remove them if they are not needed (anywhere from the file).
```
jevakallio, line 77: Agreed. Nuke 'em from orbit

docs/wiki-links.md
```diff
@@ -16,72 +16,15 @@ Foam enables you to Link pages together using `[[file-name]]` annotations.
...
-If you look at link references the bottom of any Foam workspace file that uses wiki-links, you should see an automatically generated list of references that look as follows:
-
-```markdown
-[wiki-links]: wiki-links "Wiki Links"
-[other-page]: other-page "Other Page"
-```
-
-These exist to make `[[wiki-links]]` compatible with Markdown-consuming tools such as static site generators, VS Code plugins etc. 
-
-## Why don't `[[wiki-links]]` work on GitHub
```
jevakallio, line 30: Note to self: When this change lands to docs, go and update a bunch of issues where this document is referenced.

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+  - link reference definitions might have to be in different formats depending on the publish target (e.g. Github pages vs Github UI)
+
+The potential solution:
+
+- For edit-time
+  - Make edit-time link reference definition generation optional (via user settings and maybe even "off" by default, so that the user wouldn't typically see the auto-generated section in the file at all if it's really not needed)
+  - Make format of the link reference definition configurable (whether to include '.md' or not)
+  - Out of recommended extensions, currently only "markdown links" doesn't support them (?). However even its [code](https://github.com/tchayen/markdown-links/blob/master/src/parsing.ts#L25) seems to include wikilink parser, so it might just be a bug?
+  - If user really want's to use an extension that requires "plain links", they can set the setting "on" and work as in current Foam
+- For build-time
```
jevakallio, line 71: Could we add a bullet below this describing what "build-time" is, since that's currently not a concept in Foam.

```suggestion
- For build-time
  - To satisfy mutually incompatible constraints between GitHub UI, VSCode UI, and GitHub Pages, we should add a pre-processing/build step for pushing to GitHub Pages. 
    - This would be a GitHub action (or a local script, ran via foam-cli) that outputs publish-friendly markdown format for static site generators and other publishing tools
    - This build step should be pluggable, so that other transformations could be ran during it
```

docs/link-reference-definition-improvements.md
jevakallio: If we are doing a build step anyway, I wonder if we should even use link ref defs, or if we should just nuke 'em, and "transpile" wiki-links to relative markdown links inline? I think the latter, but could have missed a reason why not.

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+
+Problem space in essence:
+
+- During edit-time (when modifying the markdown files in an editor)
+  - link reference definitions are needed if user uses editor extensions that don't understand wikilinks
+  - link reference definitions may be annoying since they 
+    - add content to files that the user hasn't typed in by themselves
+    - get out of date if user uses a tool that doesn't autogenerate them
+    - may clutter the search results
+- During build-time (when converting markdown to html for publishing purposes)
```
jevakallio, line 60: I think the build-time you describe below is not converting markdown to html, it's converting markdown to markdown (with GH Pages-friendly links) and letting GitHub pages still do the HTML generation.

Both approaches should be possible. It's not clear to me yet which should be the default for Foam:
- Have a "foam-ssg" static site generator. Powerful, publish platform agnostic, way more powerful
  - Running code means potential bugs, and if people encounter bugs in their Foam repo, they might get blocked from publishing, and have no recourse to fix it.
  - I want to build this anyway, but not sure if it should be the default
- Continue to customise the default GH Pages Jekyll/Primer with template/css/js modifications

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,88 @@
...
+
+The potential solution:
+
+- For edit-time
+  - Make edit-time link reference definition generation optional (via user settings and maybe even "off" by default, so that the user wouldn't typically see the auto-generated section in the file at all if it's really not needed)
+  - Make format of the link reference definition configurable (whether to include '.md' or not)
+  - Out of recommended extensions, currently only "markdown links" doesn't support them (?). However even its [code](https://github.com/tchayen/markdown-links/blob/master/src/parsing.ts#L25) seems to include wikilink parser, so it might just be a bug?
+  - If user really want's to use an extension that requires "plain links", they can set the setting "on" and work as in current Foam
+- For build-time
+  - Have publish targets defined in settings, that support both turning the link reference definitions on/off and defining their format (.md or not)
```
jevakallio, line 72: What would these publish targets in settings look like? Just a setting e.g. `foam.publishTarget: "GitHub Pages"`, or something more involved?

jevakallio:
@jojanaho when thinking about the settings of whether to output link ref defs, let's keep in mind that we'll probably have other similar settings, such as whether to output backlinks, link lists and other metadata into the source directory or only for publishing.

For simplicity, we might want to consider having a unified approach for these settings, e.g. having a master setting for whether you want in-place modifications, build-time modifications, or both, and then more granular settings for which transforms to actually apply.

docs/link-reference-definition-improvements.md
jojanaho: plain links here referred on non-wikilinks. With the other changes, this line becomes irrelevant (removing)

docs/link-reference-definition-improvements.md
jojanaho: edit-time use of extensions not supporting wikilinks would still need the link ref defs, right? Running those against separate build-directory might be something that users wouldn't want, especially if they don't have a need to build (meaning they are not planning to publish their notes). If you mean that we would transpile the "live" md-files from wikilinks to normal md-links, that would IMO result uglier md content and reduce the whole editing experience (so wouldn't do that either).

docs/link-reference-definition-improvements.md
jojanaho: in general build-time means just converting the original md-files to something else. You are right that the description mostly concerned md to md transformation. Build-time as a term can certainly be somewhat confusing, especially if we consider watch-like functionality and combining that e.g. with edit-time visible changes in graphs or something like that. Hopefully in the context of this description it however brought more clarity than confusion :).

docs/link-reference-definition-improvements.md
jojanaho: Will add to next commit

docs/link-reference-definition-improvements.md
jevakallio: I meant that if we have a structure:

```
/
/file.md
/.build
/.build/file.md
```

Then is there reason to generate link reference definitions into `.build/file.md`, or could we transpile `[[other]]` to `[Other](other.md)` there?

While still generating link reference definitions in `file.md` while editing.

I'm honestly not sure which is better. I feel like transpiling inline would make the output more consumable and portable, since the Markdown spec(s) ecosystem is messy, and not every tool supports link reference definitions but they all support standard links.

Thoughts?

jojanaho: If the generated site would contain "see source files" or "download editable copy" (with a requirement of having nice-to-look-at wikilinks in place), having refdefs in files would make publishing this blob easier. Otherwise the publisher would have to make also the original files separately available. Maybe also some metrics could change due to inline transpilation (e.g. amount of characters in an article) which might be confusing when comparing the results between edit-time and after build. How useful / critical these use cases are, I'm not quite sure at this point.

jojanaho: took this into consideration in the latest commit

docs/link-reference-definition-improvements.md
jojanaho: added

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,128 @@
...
+- For build-time
+  - To satisfy mutually incompatible constraints between GitHub UI, VSCode UI, and GitHub Pages, we should add a pre-processing/build step for pushing to GitHub Pages. 
+    - This would be a GitHub action (or a local script, ran via foam-cli) that outputs publish-friendly markdown format for static site generators and other publishing tools
+    - This build step should be pluggable, so that other transformations could be ran during it
+  - Have publish targets defined in settings, that support both turning the link reference definitions on/off and defining their format (.md or not). Example draft (including also edit-time aspect):
+    ```typescript
+    // settings json:
+    {
+      "foam": {
+        "publish": [
```
jevakallio, line 79: Is there a reason for this to be an array of single-keyed objects? For me it would feel more natural either as object `{ [target: string]: { wikiLinks, linkRefDefs } }`, or as array of `{ target, wikiLinks, linkRefDefs }[]`

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,128 @@
...
+          {"Gitlab Mirror": {
+            "wikiLinks": true,
```
jevakallio, line 81: How about instead of "wikiLinks" and "linkRefDefs", we could do:

```
linkFormat: mediawiki|markdown|html
definitionFormat: withExtensions|withoutExtensions|off
```

Not sure we'll need additional link formats (html is an example), but I could see a scenario where e.g. we would want to generate straight up HTML links for fancy things like deep links, if some markdown parser didn't understand tham

docs/link-reference-definition-improvements.md
```diff
@@ -0,0 +1,128 @@
...
+  - Make edit-time link reference definition generation optional via user settings. They should be on by default, and generating valid markdown links with a relative path to a `.md` file.
+  - Make format of the link reference definition configurable (whether to include '.md' or not)
+  - Out of recommended extensions, currently only "markdown links" doesn't support them (?). However even its [code](https://github.com/tchayen/markdown-links/blob/master/src/parsing.ts#L25) seems to include wikilink parser, so it might just be a bug?
+- For build-time
+  - To satisfy mutually incompatible constraints between GitHub UI, VSCode UI, and GitHub Pages, we should add a pre-processing/build step for pushing to GitHub Pages. 
+    - This would be a GitHub action (or a local script, ran via foam-cli) that outputs publish-friendly markdown format for static site generators and other publishing tools
+    - This build step should be pluggable, so that other transformations could be ran during it
+  - Have publish targets defined in settings, that support both turning the link reference definitions on/off and defining their format (.md or not). Example draft (including also edit-time aspect):
```
jevakallio, line 74: How does the publish target get picked? When is the build ran for e.g. GitHub pages?

docs/link-reference-definition-improvements.md
jojanaho: You are absolutely correct, somehow this morphed into single-key beast which doesn't make any sense. The intent was to model it according to vscode launch configurations, which uses array of objects identified by the name attribute. The reason for array is the explicit ordering, which might get lost on keyed objects depending on the execution environment. Will fix in the following commit.

docs/link-reference-definition-improvements.md
jojanaho: Will be addressed in the next commit, essentially something like this:

UI-wise, the publish targets could be picked in some similar fashion as the run/debug targets in vscode by implementing a separate panel, or maybe through command execution (CTRL+SHIFT+P) - not yet defined at this point.

docs/link-reference-definition-improvements.md
jojanaho: Good thinking! Changed this somewhat in the oncoming commit. However I'd like them to be more explicit:
- Instead of linkFormat, I'd call it linkTranspilation, since that's what it's describing. The html-related enum I'd leave out until it's really planned to be implemented, and otherwise I'd call them "mediawiki" -> "Off" (i.e. no transpilation) and "markdown" -> "WikiLinksToMarkdown", again to really describe the intent of the values.
- Instead of definitionFormat, I'd call it "linkRefDefinitionFormat" (or something else between "linkRefDefFmt" and"linkReferenceDefinitionFormat"), since definitionFormat might raise a question "definition of what"

jojanaho: Still final tweak for the oncoming commit, a name change from linkRefDefinitionFormat to linkReferenceDefinitions, since value "off" for "format" doesn't sound quite logical

jevakallio:
@jojanaho thanks you for this! Merged, will include this in the refined roadmap I'll work on next!

PR #83, opened 2020-07-09
Factored out graph operations and moved some code into core

riccardoferretti:
I will comment in-line to guide the reasoning


packages/foam-workspace-manager/src/core.ts
```diff
@@ -0,0 +1,90 @@
...
+    return this.links.filter(link => link.from === this.id)
+  }
+
+  public getBacklinks(): BubbleLink[] {
+    return this.links.filter(link => link.from === this.id)
+  }
+}
+
+
+export class Foam {
```
riccardoferretti, line 36: this is basically the interface to the graph of md notes. 
- it gives access to `Bubbles` and their connections
- it would be the object emanating events (I left that out for simplicity so far)
- it would manage refactoring operations (e.g. file rename)
- it is queried at higher levels (e.g. for backlinks)
- it can be validated for consistency (e.g. orphaned links, isolated nodes, ...)

packages/foam-workspace-manager/src/utils/utils.ts
```diff
@@ -1,23 +1,103 @@
...
 // @ts-expect-error
 export function readWorkspaceFile(filename: string): string {
   throw new Error('Not implemented');
 }
 
+// pipeline cache
+let processor: unified.Processor | null = null;
+
+function parse(markdown: string): Node {
+  processor =
```
riccardoferretti, line 18: this code I copied from @jojanaho's PR - unfortunately because the PR is on a fork I couldn't (or don't know how) create this branch off of it

packages/foam-workspace-manager/src/utils/utils.ts
```diff
@@ -1,23 +1,103 @@
...
+      .use(wikiLinkPlugin);
+  return processor.parse(markdown);
+}
+
+export function createBubbleFromMarkdown(uri: string, markdown: string): Bubble {
+  const filename = path.basename(uri);
+  const id = path.parse(filename).name;
+  const tree = parse(markdown)
+  let title = id
+  visit(tree, node => {
```
riccardoferretti, line 31: this is also copied from the other PR, I just put it inline because I need the parsing only once

packages/foam-workspace-manager/src/utils/utils.ts
```diff
@@ -1,23 +1,103 @@
...
+  return new Bubble(id, title, links, uri, markdown)
+}
+
+interface MarkdownReference {
+  linkText: string
+  wikiLink: string
+  pageTitle: string
+}
+
+export function createMarkdownReferences(foam: Foam, bubbleId: string): MarkdownReference[] {
```
riccardoferretti, line 56: I moved this here from the extension, this way it can also be used e.g. by the cli

packages/foam-workspace-manager/src/utils/utils.ts
```diff
@@ -1,23 +1,103 @@
...
+    .sort()
+}
+
+function dropExtension(path: string): string {
+  const parts = path.split(".");
+  parts.pop();
+  return parts.join(".");
+}
+
 export function parseNoteTitleFromMarkdown(markdown: string): string | null {
```
riccardoferretti, line 80: this can be ignored. I am no longer using `parseNoteTitleFromMarkdown` and `parseNoteLinksFromMarkdown` - these are just here because I haven't removed the old code yet. Again, this is just the other PR's code copied over

packages/foam-vscode/src/extension.ts
```diff
@@ -126,41 +128,10 @@ async function generateReferenceList(doc: TextDocument): Promise<string[]> {
...
-  }
-
-  references.sort()
-
-  // for (const backlink of note.backlinks) {
-  //   references.push(
-  //     `[backlink:${backlink.id}]: ${backlink.filename} "${backlink.title}"`
-  //   );
-  // }
+  const references = uniq(createMarkdownReferences(foam, id)
```
riccardoferretti, line 133: now the only thing the extension does is converting the md references for the editor (looking at it even the string creation should go in the core)

jojanaho: This is somehow related to the [link reference definition proposal](https://github.com/foambubble/foam/blob/a4277776f8596b12b915231ac2e1d81a67f54f1b/docs/link-reference-definition-improvements.md#improvement-proposal), currently I'm not quite sure in which order it would be beneficial to implement these things

packages/foam-workspace-manager/src/core.ts
jojanaho: Personally I'd call this "NoteGraph" or something similar to make terminology obvious for newcomers (or maybe I'm just dull :)). Somehow the term "foam" in my head belongs to "branding layer" of the app, at the core it's still about notes, graphs etc. Especially in (maybe unlikely) case of a project name change from Foam to something else, things would be quite confusing for a while in the codebase.

packages/foam-workspace-manager/package.json
```diff
@@ -40,10 +40,13 @@
     "typescript": "^3.9.5"
   },
   "dependencies": {
+    "@types/graphlib": "^2.1.6",
     "@types/lodash": "^4.14.157",
+    "graphlib": "^2.1.8",
```
jojanaho, line 45: Is it possible to store metadata about nodes & edges with this library or should they live in a separate index? E.g. links should probably contain e.g. line and column numbers + text length to support proper highlighting in the UI.

packages/foam-vscode/src/extension.ts
riccardoferretti: yes tbh I am not fully clear on how the link reference would work best, in this case I was reimplementing the existing behavior, but I agree this needs a bit more clarity

packages/foam-workspace-manager/package.json
riccardoferretti: yes good point - it's possible to add metadata to node/edges. I am currently storing the `Bubble` object in the node, and just the text of the link in the edge (but I should probably store an object there instead, to more easily prepare for extra metadata)

packages/foam-workspace-manager/src/core.ts
riccardoferretti: yeah I totally agree, this was me playing with ideas :) but I realize it can get confusing quickly

packages/foam-workspace-manager/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+
+
+export class Bubble {
+  public id: ID
+  public title: string
+  public source: string
+  public path: string
+  public links: OutLink[]
+
+  constructor(id: ID, title: string, links: OutLink[], path: string, source: string) {
```
jevakallio: Personally, for long argument lists made mainly of primitive data types, I prefer a named input object instead of positional arguments, since it's too easy to accidentally e.g. swap path and source order at call site

packages/foam-workspace-manager/src/core.ts
riccardoferretti: I changed the name in the latest commit, so I will resolve this conversation

packages/foam-vscode/src/extension.ts
jevakallio: @riccardoferretti is the purpose of `uniq` here to dedup the list in cases where there were multiple links to the same document? Unless those references are triple equal to each other, this will still leave duplicates as [lodash#uniq uses SameValueZero comparison](https://lodash.com/docs/3.10.1#uniq), which for object compares reference equality, not deep equality.

packages/foam-cli/src/commands/hello.ts
```diff
@@ -23,7 +23,7 @@ hello world from ./src/hello.ts!
     const {args, flags} = this.parse(Hello)
     const name = flags.name ?? 'world'
 
-    const wm = new WorkspaceManager('./foo');
+    const wm = new NoteGraph();
```
jojanaho, line 26: would make sense to call this "ng" - or maybe just "g" :)

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+  - Positional data from AST?
+- `tags`
+
+### Markdown layer
+
+- `source`: raw markdown (rename?)
+- `ast`: raw markdown ast
+- `checksum`: if we do caching
+
+### Link text
```
jevakallio, line 70: Reviewers, please ignore this, needs cleaning up -- committed working scratchpad here accidentally

packages/foam-cli/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "foam-cli",
+  "name": "@foam/cli",
```
jevakallio, line 2: Is the namespace/scope required for the TypeScript scoping? We don't own the `@foam` scope on npm so we can't publish the CLI by this name.

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+### Getting started
+
+- Land work to master
+  - Create a foam-core package
+  -
+
+### Open questions
+
+- How should writing to files work
+  - What if affected notes have unsaved changes
```
jojanaho, line 21: Good question! Writing to files (in the context of VS Code) should probably be done via something like applyEdit API (search from https://code.visualstudio.com/api/references/vscode-api#workspace)

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- get all
+- search by
+  - tag
+  - free text
+  - [[todo]]: how do vs code search editors work? are they pluggable? what do they need?
+- find dead links
+  - for linters
+- serialize/toJSON (for visualizers)
+- subscribe to changes
+
```
jojanaho, line 34: find if a link exists (and which link) in a given row / column position + return it's start and end position - this would probably be needed e.g. to CTRL-hovering to work properly

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+### Node methods
+
+- rename
```
jojanaho, line 37: This would probably rename both the node and all links to it? Or maybe that's in second phase, and in the first phase these would just be reported as invalid links. How about the other way around when refactor action (rename) would be applied to a link?

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- get links
+  - forward links (for link lists)
+  - backlinks (with surrounding context)
+
+### Node definition
+
+What do we need the node (and edge metadata) to contain:
```
jojanaho, line 44: How about the graph metadata, e.g. "root folder of the graph"?

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+
+- `id`: tbd
+  - should be unique, needs some kind of unique gen function
```
jojanaho, line 47: Would path in relation to the graph root folder work as unique id?

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+  - should be reconstructable even if links are not updated every time
+  - what happens during rename? is reparenting the graph going to be hard?
```
jojanaho, line 49: What does the reparenting of the graph mean in this context?

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+  - do id's need to be persistent, or can we create them per in-memory session, keep them stable despite renames, and then next session generate a new id?
+  - Ideally should be a path to file, so it's easy to look up from the graph by id for renaming
+- `type`: Note | Image | etc
```
jojanaho, line 52: Maybe MIME-type?

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- `title`: can be read from markdown title or frontmatter metadata
+- `path`: full path to file, relative to workspace (graph) root
```
jojanaho, line 54: The relation of this to id?

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- `links`:
+  - `id`: File to link to
+  - `text`: The link label
+  - `type` markdown | mediawiki | image | http
+  - `section`
```
jojanaho, line 59: section and block are a bit vague in this description :)

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+
+Do we apply any constraints:
+
+- `[[file-name-without-extension]]`
+- `[[file-name-with-extension.md]]`
+- `[[Title Cased File Name]]`
+
+Not supported by Markdown Notes:
+
+- `[[path/to/file-name.md]]` - Just use markdown links
```
jojanaho, line 104: IMO this is absolutely needed (better yet in format of [[path/to/file-name]]). Foldering of notes would probably be pretty typical use case (is at least to me). It's not particularly hard to implement in comparison to flat notes, and looks like we are already building the base for functionality that's currently in Markdown Notes...

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- `[[Target Note Title]]`
+
+Issues:
+
+- Name clashes in directories
```
jojanaho, line 109: Not an issue if root-based wikilinks are supported

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- Name clashes between extensions
```
jojanaho, line 110: This could probably be handled with linting

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+- Renaming
+  - Change filename/title needs to reflect everywhere
+- Orphaning
+
+- If we can't rely on in-memory process to rename things correctly while changes happen (e.g. file is renamed, moved, deleted, or titled) <ref id="1" />
```
jojanaho, line 115: Getting this to work properly probably requires using VS code APIs for rename operations (e.g. https://code.visualstudio.com/api/references/vscode-api#RenameProvider)

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+
+Solving this issue is necessarily heuristic. We could try to write smart solutions, plus a linter for orphans
+
+How others solve this:
+
+- Unique ids -- could support optionally as part of file name or front matter metadata. Should not be required.
```
jojanaho, line 121: For renames outside VS code file hash could maybe do, inside VS code probably through APIs(?)

packages/foam-core/src/core.ts
```diff
@@ -0,0 +1,89 @@
+import { Graph, Edge } from 'graphlib'
```
jojanaho, line 1: Maybe call this file (core.ts) as notegraph.ts?

packages/foam-core/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+export interface Link {
+  from: ID
+  to: ID
+  text: string
+}
+
+export interface NoteLink {
+  to: ID
+  text: string
+}
```
jojanaho, line 15: Maybe add fields from this? https://code.visualstudio.com/api/references/vscode-api#Position

packages/foam-core/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+
+
+export class Note {
+  public id: ID
+  public title: string
+  public source: string
+  public path: string
```
jojanaho, line 22: here id vs path

packages/foam-core/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+  public links: NoteLink[]
+
+  constructor(id: ID, title: string, links: NoteLink[], path: string, source: string) {
```
jojanaho, line 25: maybe use Partials here? https://stackoverflow.com/a/37682352

packages/foam-core/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.title = title
+    this.source = source
+    this.path = path
+    this.links = links
+  }
+}
+
+
+export class NoteGraph {
+  private graph: Graph
```
jojanaho, line 36: Should this contain other fields like https://github.com/foambubble/foam/blob/master/packages/foam-workspace-manager/src/WorkspaceManager.ts#L50

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: @jojanaho#3766 these are my scribbles, I pushed these to @riccardoferretti's branch while we were hacking on a call

File path relative to workspace root is exactly what we landed on.

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: The same, if the id is a path, so either we drop path, or we keep both in case we have to change the id function later

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: I believe my thinking here was that if we use file path as id, and the file's path changes, is it possible to update the node's id, or must the node be deleted, new Note node be created, and then all in/outEdges be re-routed to it. 

If id is mutable, then this is no issue.

If it's not, it's merely an inconvenience.

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: Haha! Yes, apologies, these were very rough notes. Should be

```suggestion
  - `section`: Anchor link to a heading in target note, if we want to add support for linking to sections
```

docs/meeting-notes/foam-core-2020-07-11.md
```diff
@@ -0,0 +1,125 @@
...
+  - Ideally should be a path to file, so it's easy to look up from the graph by id for renaming
+- `type`: Note | Image | etc
+- `title`: can be read from markdown title or frontmatter metadata
+- `path`: full path to file, relative to workspace (graph) root
+- `links`:
+  - `id`: File to link to
+  - `text`: The link label
+  - `type` markdown | mediawiki | image | http
+  - `section`
+  - `block` (ref)
```
jevakallio, line 60: @jojanaho this is in reference to a vague plan I have about building Roam's [block references and block embeds](https://www.roamtips.com/home/what-is-block-roam-research) into Foam. This is not something we need to think of now, more of a future consideration.

The idea here was to annotate blocks of text (paragraph, bullet, line, etc.) using some method (HTML tags were discussed), and then being able to link to that section from another document such that in VS Code that block is brought as an editable embed to that document. Additionally, this linking would be supported by a hypothetical Foam static site generator, that would be able to construct published pages from blocks in other documents.

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: Agreed

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: @jojanaho could you expand on this?

packages/foam-cli/src/commands/hello.ts
jevakallio: Or, gosh, even `graph` to be painfully explicit :)

(This is in the dummy code in CLI, however, so not mega important, since this file won't live through first real commit to the CLI project)

packages/foam-core/src/core.ts
jevakallio: 👍 -- I'd say `note-graph.ts` since capital G makes it a compound word, and therefore should be separated.

packages/foam-core/src/core.ts
jevakallio: Mentioned this on previous comment, but the fact that the ID happens to be a path is an implementation detail and we shouldn't rely on the fact. If we do need a file path (and we do), I think it would be good to keep it in `path`

packages/foam-core/src/core.ts
jevakallio: Currently the old WorkspaceManager.path was never used, but it may be useful considering that in the graph nodes we want to keep relative urls, and it may be useful to have the information at hand to construct absolute file URIs

tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
+{
+  "include": ["./packages/*/src"],
+  "compilerOptions": {
+    "declaration": true,
+    "declarationMap": true,
+    "baseUrl": ".",
+    "paths": {
+      "@foam/core": ["./packages/foam-core/src"],
```
jevakallio, line 8: If the `@scope/package` naming convention is not mandatory, perhaps we should drop it? Alternatively we'll need another scope, since we don't own the npm user/org `@foam`. We could do `@foambubble`, but that makes it slightly confusing as folks might think it's somehow coupled to the GitHub org.

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: Can we add @jojanaho's commend to the list here, so we don't forget it?

```suggestion
  - find if a link exists (and which link) in a given row / column position + return it's start and end position - this would probably be needed e.g. to CTRL-hovering to work properly
```

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: As per comment later on, adding this would not hurt, so that implementers will have access to it if needed

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: It would have to rename node and all links. If we are to support `[[Page Title]]` links, there's also a special case here where the rename may have to not only regenerate the link reference definitions, but replace the link text inside the document itself.

```suggestion
- rename node and all links to that node
```

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: We could implement a Node TS type that's a union of different types (Note | Image | Etc)

packages/foam-vscode/src/extension.ts
riccardoferretti: yeah, that's why I am doing the `uniq` on the resulting string, which is kinda of a hack, this should be done at the `createMarkdownReferences` layer (on the object as you hinted at), I was a bit lazy about doing the right comparison. I should really move it there..

packages/foam-workspace-manager/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+export class Foam {
+  private graph: Graph
+
+  constructor() {
+    this.graph = new Graph()
+  }
+
+  public setBubble(bubble: Bubble) {
+    if (this.graph.hasNode(bubble.id)) {
+      (this.graph.outEdges(bubble.id) || []).forEach(edge => {
```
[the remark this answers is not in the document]
riccardoferretti: basically this cleans the old edges before adding the new ones, otherwise the old links would still exist - I might actually move that out and have a `removeBubble` (which we would need when removing a file anyways)

packages/foam-workspace-manager/src/core.ts
```diff
@@ -0,0 +1,89 @@
...
+        this.graph.removeEdge(edge)
+      })
+    }
+    this.graph.setNode(bubble.id, bubble)
+    bubble.links.forEach(link => {
+      this.graph.setEdge(bubble.id, link.to, link.text)
+    })
+  }
+
+  public getBubbles(): Bubble[] {
```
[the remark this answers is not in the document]
riccardoferretti: This one returns all the notes in the graph, whether or not they are linked to others, that's why I kept the name bare, what do you think?

Somewhat related, I tend to use the following conventions: 
- `getNote(id): Note` expects the `id` to exist and will throw if it doesn't find it
- `findNote(id): Note | null` for those situations when it's unsure whether the `id` exists, but don't wanna fail (alternative would be to have `hasNote(id)`, I slightly prefer the `find` syntax)
- `listNotes(): Note[]` when dealing with arrays

If you like them we could also use those conventions, I am super easy

packages/foam-workspace-manager/src/core.ts
riccardoferretti: I agree 100%.

In fact, for `Note`, I was thinking of just making it a type/interface, as I have removed all the methods I initially added here

packages/foam-cli/src/commands/hello.ts
riccardoferretti: good point, here I was just fixing the object used, wasn't paying attention to the name of the variable

as it's just a dummy I will leave it that way

packages/foam-cli/package.json
riccardoferretti: good point I didn't think of that. will use `foam-core`, `foam-cli` and `foam-vscode`

packages/foam-cli/src/commands/hello.ts
riccardoferretti: nevermind, there was a compilation error afterwards, so I just commented this part of the function :)

docs/meeting-notes/foam-core-2020-07-11.md
riccardoferretti: I started to look into this, there are a couple of open questions that we can chat about. I will not include it in this PR though so we can merge it in the meantime

docs/meeting-notes/foam-core-2020-07-11.md
riccardoferretti: I was leaning towards having the ID being the URI of the file, basically case n.2
I don't see it as a big issue, and I think it might be cleaner, but I am open to the idea of having another ID for the note (the question would be how to generate it/store it/....)

docs/meeting-notes/foam-core-2020-07-11.md
riccardoferretti: yeah, there are some considerations here similar to the ones I have encountered for the links. we should have a chat about these

docs/meeting-notes/foam-core-2020-07-11.md
riccardoferretti: I started to look into this, but it won't probably make it in this PR as it turns out a couple of things change around wikilinks as a result

docs/meeting-notes/foam-core-2020-07-11.md
riccardoferretti: what do you mean? I am not familiar with root-based wikilinks

packages/foam-core/src/core.ts
riccardoferretti: yeah this could be interesting, but I want to better understand the use case (do we need to update the info on every edit, or only on save? depends on what we need the info for I think)

packages/foam-core/src/core.ts
riccardoferretti: I was wondering whether this needs to be a class in the first place, or just a type/interface - what do you think?

packages/foam-core/src/core.ts
riccardoferretti: yeah I removed it because it wasn't used, and I was debating whether this info should live in the config or the graph. currently we store the absolute paths in the graph, having a base path would allow us to make those relative. why do you prefer relative urls @jevakallio ?

packages/foam-workspace-manager/src/core.ts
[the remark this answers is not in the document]
jevakallio: I'm fine with the current convention in the latest iteration of this PR 👍

packages/foam-workspace-manager/src/core.ts
jevakallio: @riccardoferretti yes, all for using types over classes whenever we can!

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: Perfect

docs/meeting-notes/foam-core-2020-07-11.md
jevakallio: @riccardoferretti I think @jojanaho meant the line he commented on above, e.g. `[[path/to/file-name.md]]`.

Sounds like we need to write a further document on aligning how we'll approach this. @jojanaho on our call with Riccardo yesterday we brain stormed a few ideas (in the scratchpad section above) but didn't make any decisions or arrive in firm conclusions

packages/foam-core/src/core.ts
jevakallio: @riccardoferretti that's a good question. 

I guess I prefer them for readability -- the paths will be closer to what's being used in links and generated into link ref defs -- but that's not a strong preference.

Another argument for relative paths is that if we ever persist the graph, relative links continue to work when e.g. cloning the repo to different computer, moving it to different folders, or even as logical paths in a GitHub repo / published site.

If there's no good reason for absolute paths in `Note.path`, I'd rather keep it relative, but can be convinced otherwise.

Issue #80, opened 2020-07-09
Publish Foams from a private repo


juanfrank77:
Hey @jevakallio I'm opening this issue to report my findings on publishing Foams from a private repo without 
using GH pages _and_ Jekyll.

I tried two ways, using eleventy as the SSG and then hosting it on Netlify and using Next.js as SSG and hosting it on Vercel. 

After some work and figuring out stuff I was able to make both of them work. I created two repos based on the foam template which are private to try out if publishing from a private repo was possible. However, since you mentioned that the intention is not lock-in the user into a specific set of tools, I think the better approach is to be as less extra config as possible. 

Doing the process of conversion to a static site and then hosting it online is quite cumbersome with Next.js because it requires config both on the next project and also on a Vercel account for it to be properly displayed. On the other hand, doing that process with eleventy requires much less config and Netlify does the deployment process quite easy regardless of private or public repo. 
 
I guess the next step is for me to make a foam eleventy template public so you can look at it and let me know if that's one way in which people can publish their private Foams while having control of which bubbles they can make public and which ones to keep as a secret (maybe for privacy or for being WIP's).

Thanks mate.

SanketDG:
I would love to take a look at the 11ty template!

How did you get all the [[linking]] and backlinking features?

jevakallio:
@juanfrank77 thank you for doing this work! Once you have the template ready, let's get it into the docs along with the foam-gatsby-template. Maybe a new "Templates" section!

juanfrank77:
Hey guys.

I think I'm done setting up the template. I found out the hard way that markdown documents with a Front Matter header doesn't work yet and would have to do too much fancy stuff to get it working so nah.

Also, I couldn't decide on what visual theme to put on the pages, so I just left it with default barebones styling.

Here it is the [template](https://github.com/juanfrank77/foam-eleventy-template)

PR #95, opened 2020-07-11
Add recipe for publishing with netlify

juanfrank77:
Hey @jevakallio 

I added the recipe for publishing foam to Netlify using Eleventy as per #80. Not sure if this is the format you intended or putting in a 'Templates' section like you mentioned. 

Anyway, let me know which way would be better. 

Thanks mate


jevakallio:
@all-contributors add @juanfrank77 for doc

PR #96, opened 2020-07-12
Update Contributor Image Link 🧐 in Readme

sauravkhdoolia:
Hey guys, 

I saw that the contributor's image link wasn't getting rendered for some reason. I tried to get it to work in md, but in vain.

So, Finally i inserted an Anchor tag with img tag in it, instead of using md img links. 

Also, The readme has two contributors images one with 0 count and another one with count 16. So, i removed one, hoping that it was the right decision. ** That fixes it! ** 🙃


jevakallio:
@all-contributors please add @sauravkhdoolia for doc

PR #104, opened 2020-07-13
Init foam-workspace-janitor

jevakallio:
Setting up some space for us to work on the Foam CLI / Workspace Janitor stuff

### DIFF
--- a/packages/foam-workspace-janitor/.github/workflows/main.yml
+++ b/packages/foam-workspace-janitor/.github/workflows/main.yml
@@ -1,0 +1,42 @@
+name: CI
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Begin CI...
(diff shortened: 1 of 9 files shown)


jevakallio:
I'm closing this as per Discord feedback from @riccardoferretti:
https://discordapp.com/channels/729975036148056075/732125304646270985/732151729436753942

> if we can manage I would keep the janitor as part of the foam-core package purely for convenience, but within it as it's own subdir to make it easier spotting the dependencies

> we could also put it into its own module, not sure though whether it's worth complicating the dep graph

Agreed, it's not worth complicating the dependency graph. I see a few ways where we'd get into issues trying to keep versions up to date.

Self-closing this PR. @anku255 @chirag-singhal for now, put all the janitorial files to foam-core/janitor, and any file system related features to foam-cli. If we find true middle ground that fits neither, we can do a quick refactor.

PR #105, opened 2020-07-13
Proposals: deep links and block embeds

jevakallio:
By @riccardoferretti and @jevakallio

We discussed about some potential near future enhancements to the linking experience, and agreed that there are a couple of additional features that would be cool:

- **Deep links**: Support and autocompletion for linking inside a section in another note
  - This could involve defining custom sections other than just markdown headings (we have a few different syntax proposals to consider -- new ideas welcome)
  - Main reason for custom sections is block embeds aka. block transclusion, 
- **Block embeds**: The idea where instead of simply linking to a section, we bring it inline to the other document as VS Code Peek Definition pane, and also in the generated HTML side (this presumes build step).

These are in addition to the work @digiguru is doing with supporting [[Loose Matching Links]], and the spike I'm doing to autocomplete links based on File Headings instead of file-names.


mathieudutour:
Just going to chim in with how Roam is doing it:
- each bullet (or in Foam case, paragraphs) has an id
- a block is referenced with `((block-id))`. It is replaced by the block content. Writing `((` opens the autocomplete to find the block to reference

Each block is actually a page on its own, with a link to its parent at the top.

jlongster:
@mathieudutour How is the block-id generated in Roam? The first couple words of the block?

I'd love block links, don't care as much about block embeds, if it makes sense to just do linking first.

mathieudutour:
No it’s an opaque uid which stays constant regardless of the content (it’s the id of the node in the db)

jevakallio:
Closing in favour of https://github.com/foambubble/rfcs/pull/3

PR #107, opened 2020-07-13
Spike: Link and Completion Providers

jevakallio:
I'm playing around with link and completion providers. This is not ready for review yet (and won't be for a while, as it will conflict with Markdown Notes autocompletions).


jevakallio:
Closing this, as we won't be proceeding with the VSCode provider approach for now.

PR #108, opened 2020-07-13
Add NoteLink.position to keep track of link ranges

jevakallio:
Cherry-picked this off from #107, as it's likely we'll need this for other core/janitor features before the spike can land.

- Adds `NoteLink.position` field to attach Markdown AST positional information to links, to be used by VS Code providers. 
- Uses the unist types directly as they are just plain interfaces.

### DIFF
--- a/packages/foam-core/src/markdown-provider.ts
+++ b/packages/foam-core/src/markdown-provider.ts
@@ -4,7 +4,7 @@ import wikiLinkPlugin from 'remark-wiki-link';
 import visit, { CONTINUE, EXIT } from 'unist-util-visit';
 import { Node, Parent } from 'unist';
 import * as path from 'path';
-import { Link, Note, NoteGraph } from './note-graph';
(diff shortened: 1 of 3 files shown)


packages/foam-core/src/markdown-provider.ts
```diff
@@ -29,13 +29,13 @@ export function createNoteFromMarkdown(uri: string, markdown: string): Note {
...
   });
-  const links: Link[] = [];
+  const links: NoteLink[] = [];
   visit(tree, node => {
     if (node.type === 'wikiLink') {
       links.push({
-        from: id,
         to: node.value as string,
         text: node.value as string,
+        position: node.position!
```
jevakallio, line 38: Technically node.position may be null, but this should only be the case when the links have been generated/appended to the tree and not yet committed. The nodes here are always in markdown, and therefore should always have positions.

Issue #94, opened 2020-07-11
Saving a file generates a new link reference section every time


ianjsikes:
Possibly related to #20

When I save a file that contains [[wiki-links]], the generated markdown references appear at the bottom of the file. If I save the file again (even if there are no changes, just pressing `cmd+s`), _another_ identical generated section appears below the first one. [Here is a video][video].

If I `> Save without formatting`, no link references are generated.

**Also**, if I delete the actual [[wiki-links]] from the page and then save, this still happens (with references generated for the links that are no longer there). I have to quit and re-open VS Code to make it realize there are no links there anymore, at which point saving does not generate more references.

### System Info
macOS 10.14.6
VS Code 1.47.0
Foam for VSCode 0.1.7
Markdown All in One 3.1.0
Markdown Links 0.6.0
Markdown Notes 0.0.11
Prettier 5.1.3

[video]: https://share.getcloudapp.com/Z4uY999w

jevakallio:
Wow, that's broken! I'm on same OS, VS Code and Foam versions, and I'm not
seeing this issue.

The link syntax you're using, using double square brackets [label][file] is
not something that we officially support. The way you'd do it right now
would be either as wiki links `[[file]]`, or as a standard markdown link
`[label](file.md)`. Wiki links should generate reference definitions, and
the markdown link should not (since it's already markdown).

Could you test if the [[file]] syntax works?

Otherwise, let me try to land https://github.com/foambubble/foam/pull/83 today
and release a new version to see if that fixes the problem.

Cheers,
Jani

ianjsikes:
Ah woops I used a bad example file. The wiki link in the video is the one that says `[[how-to-run-a-dungeon]]` at the top. [Here is a smaller example][vid] with _only_ wiki links.

I figured out the cause. I have the `files.insertFinalNewline` VS Code setting enabled to automatically add an empty newline to the end of files that don't have one [because Git][newline]. So when I save the file, it seems like:
1. VS Code adds a newline if there isn't one
2. Foam checks if the last lines of the file are the generated references (I am assuming this is how it works, haven't looked at the code yet).
3. Generate new reference if there isn't one at the end of the file.

But on subsequent saves, VS Code will add the newline _after_ the references, which seems to cause Foam to not find them and generate new ones.

I disabled the setting in my workspace settings and the problem is gone.

[vid]: https://share.getcloudapp.com/lluJAezQ
[newline]: https://thoughtbot.com/blog/no-newline-at-end-of-file

[...]

PR #112, opened 2020-07-14
Implement foam-cli janitor command

anku255:
This Pull Request adds the `foam-cli janitor` command.

It does the following: 

- Adds heading to the files if they don't exist
- Adds/Update/Removes link references
- kebab cases file names if they are not already

**Why this can't be merged?**

1. Needs refactoring.
2. The command needs better documentation.

Update: Tests for generateHeading has been added!


packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const graph = await initializeNoteGraph(foamWorkspaceDir);
+
+    const notes = graph.getNotes();
+
+    const fileWritePromises = notes.map(note => {
+      // Generate Heading 
+      const heading = generateHeading(note);
+
+      const fileWithHeading = heading ? applyTextEdit(note.source, heading) : note.source;
```
jevakallio, line 39: Usually variable reassignment is bad, but in this case I think it would improve readability to reassign to same instead of keep coming up with names.

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      // Generate Link References
+      const linkRefTextEdits = generateLinkReferences(note, graph);
```
jevakallio, line 42: For readability, I'd like to see a bit code grouping here. Since these operations are not dependent on each other, you could group them like:
```
  // get edits
  const heading = ...
  const definitions = ...

  //apply edits
  let file = note.source;
  file = heading ? applyTextEdit(file, heading) : file;
  file = definitions ? applyTextEdit(file, definitions) : file;
```

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,73 @@
...
+
+
+    // TODO: Add a loader
+
+    await Promise.all(fileWritePromises);
+
+
+    // Improve the message (also show changed files??)
+    // Use Chalk
+    this.log('Succfully generated link references and heading!')
```
jevakallio, line 71:
```suggestion
    this.log('Successfully generated link references and heading!')
```

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,73 @@
...
+      return null;
+    })
+
+    // Kebab case file names
+    await Promise.all(notes.map(note => {
+      const kebabCasedFileName = getKebabCaseFileName(note.title);
+      if (kebabCasedFileName) {
+        return renameFile(note.path, getKebabCaseFileName(note.title));
+      }
+      return null;
```
jevakallio, line 59: Does Promise.all allow nulls?

packages/foam-cli/src/utils/apply-text-edit.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @returns {string} text with the applied textEdit
+ */
+export const applyTextEdit = (text: string, textEdit: TextEdit): string => {
+  const characters = text.split('');
+  const startOffset = textEdit.range.start.offset || 0;
+  const endOffset = textEdit.range.end.offset || 0;
+  const deleteCount = endOffset - startOffset;
+
+  const textToAppend = `${textEdit.newText}`;
+  characters.splice(startOffset, deleteCount, textToAppend);
```
jevakallio, line 16: So simple, so beautiful 😍

packages/foam-cli/src/utils/rename-file.ts
```diff
@@ -0,0 +1,11 @@
+import * as fs from 'fs'
+
```
jevakallio, line 2: Remove excess whitespace

packages/foam-cli/src/utils/rename-file.ts
```diff
@@ -0,0 +1,11 @@
...
+
+
+export const renameFile = async (fileUri: string, newFileName: string) => {
+  const fileName = fileUri.split('/').pop();
```
jevakallio, line 6: This may not work on Windows -- would depend whether URIs are normalised before passing. There's probably a node x-plat method for doing this

packages/foam-cli/src/utils/rename-file.ts
```diff
@@ -0,0 +1,11 @@
...
+  const extension = fileName?.split('.').pop();
+  const newFileUri = fileUri.replace(`${fileName}`, `${newFileName}.${extension}`);
```
jevakallio, line 8: This would break if the file name is anywhere in the directory, e.g. `markdown/markdown.md`.

packages/foam-cli/src/utils/rename-file.ts
```diff
@@ -0,0 +1,11 @@
+import * as fs from 'fs'
+
+
+
+export const renameFile = async (fileUri: string, newFileName: string) => {
```
jevakallio, line 5: There's probably better ways of implementing this using Node `path` APIs

packages/foam-cli/src/utils/write-file-to-disk.ts
```diff
@@ -0,0 +1,8 @@
+import * as fs from 'fs';
+
+export const writeFileToDisk = async (fileUri: string, data: string): Promise<Boolean> => {
+  return fs.promises.writeFile(fileUri, data).then(() => true).catch(err => {
```
jevakallio, line 4: Why do we catch and log here instead of letting the Error bubble up to the caller?

I'm not opposed to it, just wondering if it's necessary

packages/foam-cli/test/write-file-to-disk.test.ts
```diff
@@ -0,0 +1,19 @@
...
+describe('writeFileToDisk', () => {
+  it('should overrwrite existing file in the disk with the new data', async () => {
+    const existingData = `content in the existing file`;
+    const newData = `content in the new file`;
+    const fileUri = path.join(__dirname, 'test-file.md');
+    fs.writeFileSync(fileUri, existingData);
+
+    await writeFileToDisk(fileUri, newData);
+
+    const actual = fs.readFileSync(fileUri, { encoding: 'utf8' });
```
jevakallio, line 15: No need to make this sync. The test is already async, just await on readFile to avoid blocking on concurrent tests

packages/foam-cli/test/write-file-to-disk.test.ts
```diff
@@ -0,0 +1,19 @@
+
+import { writeFileToDisk } from '../src/utils/write-file-to-disk'
+import * as fs from 'fs';
+import * as path from 'path';
+
+describe('writeFileToDisk', () => {
+  it('should overrwrite existing file in the disk with the new data', async () => {
```
jevakallio, line 7: This is more of an integration test than a unit test (since it does I/O). I'm not sure how I feel about it -- the proper way of testing this would be to mock `fs` and test writes against mocks.

If that's too time consuming, I'm happy to keep this test and refactor after the fact.

packages/foam-cli/test/write-file-to-disk.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    const existingData = `content in the existing file`;
+    const newData = `content in the new file`;
+    const fileUri = path.join(__dirname, 'test-file.md');
```
jevakallio, line 10: If we don't mock the file system operations, at minimum we should write to a temp directory, so we don't pollute the active git workspace in case cleanup fails.

packages/foam-cli/test/write-file-to-disk.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    fs.writeFileSync(fileUri, existingData);
+
+    await writeFileToDisk(fileUri, newData);
+
+    const actual = fs.readFileSync(fileUri, { encoding: 'utf8' });
+    expect(actual).toBe(newData);
+    fs.unlinkSync(fileUri);
```
jevakallio, line 17: If we don't do mocking, can we move the cleanup to `after` or `afterEach` so that if the test throws, at least the files get cleared.

packages/foam-core/src/janitor/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    stringifyMarkdownLinkReferenceDefinition
+  ).join('\n');
+
+  if (note.definitions.length === 0) {
+    if (newReferences.length === 0) {
+      return null;
+    }
+
+    // @todo: how to abstract new line?
+    const padding = note.end.column === 1 ? '\n' : '\n\n';
```
jevakallio, line 26: Note @anku255 @chirag-singhal. We still need to figure out how/where to normalise newlines. Usually newline format depends on platform but in some cases (e.g. when using git), people's git configs can check out linux new lines on windows and vice versa -- not sure what the best way is.

packages/foam-core/src/note-graph.ts
```diff
@@ -15,17 +15,28 @@ export interface NoteLink {
...
+  title?: string;
+  position?: Position;
+}
+
 export class Note {
   public id: ID;
   public title: string;
   public source: string;
   public path: string;
+  public end: Point;
```
jevakallio, line 30: FYI @riccardoferretti @jojanaho I added Note.end (the line/col/offset at end of file) and Note.definitions (the existing parsed link ref defs and their positions) here. 

Ideally, there would be a cleaner way of doing this. e.g. some kind of metadata container on the graph node (and arguably, the definitions should live on the edges, maybe?). I'm open to refactoring the naming/storage of these, but any conceptual issues with this?

packages/foam-core/src/utils.ts
```diff
@@ -1,5 +1,43 @@
...
+/**
+ * 
+ * @param filename 
+ * @returns title cased heading after removing special characters
+ */
+export const getHeadingFromFileName = (filename: string): string => {
+  return titleCase(filename.replace(/[^\w\s]/gi, ' '));
+}
+
+export const initializeNoteGraph = async (aboluteDir: string) => {
```
jevakallio, line 25: Not sure this belongs to utils -- maybe its own file?

packages/foam-core/types/utils.d.ts
```diff
@@ -1 +1,3 @@
 declare module 'remark-wiki-link';
+
+declare module 'dashify';
```
jevakallio, line 3: Does this mean there's no typescript definitions for this module?

packages/foam-cli/src/commands/janitor.ts
anku255: It does! Not sure if there is a better way to handle this.

packages/foam-core/types/utils.d.ts
anku255: There is! I have added the `@types/dashify` as dep.

packages/foam-core/src/utils.ts
anku255: Done!

packages/foam-cli/test/write-file-to-disk.test.ts
anku255: I have moved it to `afterAll` for now but I will try to mock `fs` (added a TODO).

packages/foam-cli/src/commands/janitor.ts
anku255: Awesome! That does make the code more readable. I have pushed the change.

packages/foam-core/src/janitor/index.ts
anku255: Will look into it.

packages/foam-core/src/note-graph.ts
jevakallio: It's the definitions that already exist in the file. `end` is the line/col/offset at end of file -- essentially where the markdown AST root node ends, and new content may be appended

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,78 @@
...
+  static args = [{ name: 'workspacePath' }]
+
+  async run() {
+    const spinner = ora('Reading Files').start();
+
+    const { args, flags } = this.parse(Janitor)
+
+    const { workspacePath = './' } = args;
+
+    const foamWorkspaceDir = path.join(__dirname, workspacePath);
```
jevakallio, line 32: I don't think we need to join here? Just `workspacePath` will do?

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,78 @@
...
+    })
+
+    spinner.succeed();
+    spinner.text = 'Renaming files';
+
+    // Kebab case file names
+    await Promise.all(notes.map(note => {
+      const kebabCasedFileName = getKebabCaseFileName(note.title);
+      if (kebabCasedFileName) {
+        return renameFile(note.path, getKebabCaseFileName(note.title));
```
jevakallio, line 67: Should be, I think:

```suggestion
        return renameFile(note.path, kebabCasedFileName);
```

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const { args, flags } = this.parse(Janitor)
+
+    const { workspacePath = './' } = args;
+
+    const foamWorkspaceDir = path.join(__dirname, workspacePath);
+
+    const graph = await initializeNoteGraph(foamWorkspaceDir);
+
+    const notes = graph.getNotes();
```
jevakallio, line 36: Let's log how many files we found, and if it's 0, exit early. When I was testing this, it said transformation succeeded when I provided a path that didn't match anything

packages/foam-cli/src/commands/janitor.ts
jevakallio: Let's also check that the directory exists, and exit if it doesn't.

packages/foam-core/package.json
```diff
@@ -15,18 +15,23 @@
     "prepare": "tsdx build"
   },
   "devDependencies": {
+    "@types/dashify": "^1.0.0",
     "@types/graphlib": "^2.1.6",
     "@types/lodash": "^4.14.157",
+    "glob": "^7.1.6",
```
jevakallio, line 21: `glob` needs to be a dependency, not a devDependency

packages/foam-cli/package.json
```diff
@@ -11,17 +11,23 @@
     "@oclif/command": "^1",
     "@oclif/config": "^1",
     "@oclif/plugin-help": "^3",
```
jevakallio, line 13: We need to add foam-core into dependencies as well for npm publishing

packages/foam-cli/src/commands/janitor.ts
chirag-singhal: Yeah @jevakallio workspacePath works, changing it.

packages/foam-core/package.json
chirag-singhal: @jevakallio fixed this

packages/foam-core/src/note-graph.ts
riccardoferretti: @jevakallio if I understand correctly, you are now taking into account the existing definitions within the file (I guess so that we don't double-generate them?).
Do you think we could just treat our generated code as regular definitions? What I mean is that instead of having a defined (code-lensed) section for the definition, we simply generate definitions at the bottom of the file for any missing wikilink.

- the advantage is that it integrates the 2 use cases
- might even automatically solve #94 
- possible drawback is that it might be a bit spooky to generate text in the file without properly flagging it
- (meta: this might not be the best place to discuss, happy to take it to a different PR/issue)

packages/foam-core/src/janitor/index.ts
```diff
@@ -0,0 +1,79 @@
...
+        start: first.position!.start,
+        end: last.position!.end,
+      },
+    };
+  }
+};
+
+export const generateHeading = (note: Note): TextEdit | null => {
+  // Note: This may not work if the heading is same as the file name
+  if (note.title !== note.id) {
```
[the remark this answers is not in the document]
anku255: Hey @riccardoferretti!

We wanted to check if the file already has a heading or not.  If `note.id` and `note.title` are the same then this probably means that the note didn't have a heading tag and we generate a new one.

Not sure how else can we do this without traversing the AST again.

packages/foam-core/src/note-graph.ts
```diff
@@ -15,25 +15,41 @@ export interface NoteLink {
...
+  public eol: string;
```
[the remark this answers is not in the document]
anku255: As per our discussion with @jevakallio , we decided to store it on the note level. This might be useful when the same workspace is edited from different platforms.

packages/foam-core/package.json
```diff
@@ -23,10 +24,14 @@
     "typescript": "^3.9.5"
   },
   "dependencies": {
+    "dashify": "^2.0.0",
```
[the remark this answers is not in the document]
anku255: We are going with `github-slugger` as it might be useful later on. :)

packages/foam-cli/package.json
anku255: done!

packages/foam-cli/src/utils/rename-file.ts
chirag-singhal: Done!

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const { workspacePath = './' } = args;
+
+    if (fs.existsSync(workspacePath) && fs.lstatSync(workspacePath).isDirectory()) {
+      const graph = await initializeNoteGraph(workspacePath);
+
+      const notes = graph.getNotes();
+
+      spinner.succeed();
+      spinner.text = `${notes.filter(note => note !== undefined).length} files found`;
```
jevakallio, line 38: Instead of filtering here and then checking again inside notes.map, why not filter once and reuse the same array?

packages/foam-cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,97 @@
+import { Command, flags } from '@oclif/command';
+import * as ora from 'ora';
+import { initializeNoteGraph, generateLinkReferences, generateHeading, getKebabCaseFileName } from 'foam-core';
+import { applyTextEdit } from '../utils/apply-text-edit';
+import { writeFileToDisk } from '../utils/write-file-to-disk';
+import { renameFile } from '../utils/rename-file';
+import * as fs from 'fs'
+
+
+export default class Migrate extends Command {
```
jevakallio, line 10: Bit of repetition here between `janitor` and `migrate`. Ideally I'd like to have these refactored a bit, but since we will soon change how the reference link definition generation works, we can do that as a second pass then.

Let's leave a `@todo` here at least.

packages/foam-core/src/note-graph.ts
jevakallio: Note: above conversation led to proposal https://github.com/foambubble/foam/pull/122

packages/foam-core/src/initializeNoteGraph.ts
```diff
@@ -0,0 +1,30 @@
...
+import fs from 'fs';
+import os from 'os';
+import detectNewline from 'detect-newline';
+import { NoteGraph } from './note-graph';
+import { createNoteFromMarkdown } from './markdown-provider';
+
+const findAllFiles = promisify(glob);
+
+export const initializeNoteGraph = async (aboluteDir: string) => {
+  // remove trailing slash from aboluteDir if exists
```
jevakallio, line 13:
```suggestion
  // remove trailing slash from absoluteDir if exists
```

packages/foam-core/src/initializeNoteGraph.ts
```diff
@@ -0,0 +1,30 @@
...
+import { promisify } from 'util';
+import fs from 'fs';
+import os from 'os';
+import detectNewline from 'detect-newline';
+import { NoteGraph } from './note-graph';
+import { createNoteFromMarkdown } from './markdown-provider';
+
+const findAllFiles = promisify(glob);
+
+export const initializeNoteGraph = async (aboluteDir: string) => {
```
jevakallio, line 12: Typo: Rename variable to `absoluteDir`

packages/foam-core/src/janitor/index.ts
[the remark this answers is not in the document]
jevakallio: This relies on `let title = id;` in `createNoteFromMarkdown`. It works, but agree with Riccardo it's a little sketchy. A better solution would be to make `title` nullable and not populate it with id by default. 

That change will likely break something else however, so if we do it, we'll have to understand where the title is used.

packages/foam-core/src/note-graph.ts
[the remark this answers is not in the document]
jevakallio: @riccardoferretti we don't *expect* them to be, but realistically they may be, so might as well do the right thing here and avoid a weird bug down the line.

jevakallio:
Wow, this is one chonky pr boi 👀  

I tested this locally and it seems to be working, but there are still a few code issues to look at. @chirag-singhal @anku255 can you look through the unresolved review comments and address the ones that are possible to address in a reasonable time?

For things that aren't easily addressable, let's collect a list of TODOs and merge this, since the foam-core changes are needed for other work, and the foam-cli is not yet and there are no released versions, so we're not breaking anything live)

jevakallio:
Oops, hit close by accident!

packages/foam-core/src/janitor/index.ts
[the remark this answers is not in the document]
riccardoferretti: Oh, I see what you mean.
That would work now, but be aware that you are making a fairly big assumption about both what's in the `note.id` and _how_ the `note.title` is computed. The check is using an implementation detail of `Note`, that is, the reason why you know this check works is because you have seen the code that provides the default value to `title`.

As a workaround you could add a test that would fail in this case, so we can catch it shall we make changes to `Note` (which I think are coming). 

Alternatively we could explore whether we really need `title` to default to `id`, or whether we could leave it `null` if there is no title in the note. This wold also solve the `Note: ` you put in the line above your check.
@jevakallio might also have an idea here.
 
Something to think about :)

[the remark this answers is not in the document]
riccardoferretti: @jevakallio our comments overlapped, but it seems we agree the `null` route is the right way forward.

packages/foam-core/src/initializeNoteGraph.ts
anku255: Renamed it to `workspacaPath` since that made more sense than` absoluteDir`.

packages/foam-cli/src/commands/janitor.ts
anku255: Done! Added the code for exiting early in case no files are found.

packages/foam-cli/src/commands/migrate.ts
anku255: I agree. I did some refactoring but still, there is a lot that can be done there. I have added a @todo for now.

packages/foam-core/src/janitor/index.ts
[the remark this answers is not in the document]
anku255: We needed this PR to get merged quicky. Hence, I have created #131. Will fix it soon :)

packages/foam-cli/src/commands/janitor.ts
chirag-singhal: Done by @anku255

packages/foam-cli/test/write-file-to-disk.test.ts
anku255: Added `mock-fs`.

packages/foam-cli/src/commands/janitor.ts
anku255: Done!

packages/foam-cli/test/write-file-to-disk.test.ts
anku255: Done!

packages/foam-cli/src/utils/write-file-to-disk.ts
anku255: Done!

packages/foam-cli/src/utils/rename-file.ts
anku255: Done!

packages/foam-cli/src/utils/rename-file.ts
anku255: It should be fixed now.

packages/foam-cli/test/write-file-to-disk.test.ts
anku255: I am mocking the fs now. But maybe the tests can still be improved.

jevakallio:
@all-contributors add @chirag-singhal for code 🎉

PR #122, opened 2020-07-15
Proposal: Link Definition and Disambiguation

jevakallio:
Masterstroke idea by @riccardoferretti! Super excited about this.

This is fully compatible with the existing [Link Reference Definition Improvements](https://foambubble.github.io/foam/link-reference-definition-improvements) proposal, with the exception that with this plan, it's no longer feasible to set `LinkReferenceDefinitions: Off`

### DIFF
--- a/docs/proposal-link-definition-and-disambiguation.md
+++ b/docs/proposal-link-definition-and-disambiguation.md
@@ -1,0 +1,51 @@
+# Proposal: Link Definition and Disambiguation
+
+## Current status
+
+- We generate link ref defs from wiki-links inline
+- Wiki links are expected to be the full source of truth, but we don't have enough infomration to conclusively disambiguate
+
+## Solution
+
+- Don't wipe out link ref defs. Only add/remove as needed. This can be our metadata for disambiguation.
+
+Let's say we have two documents in the workspace with the same title "Some Title"
+
+- We would have autocompletion base on title of the document
+  - This generates a link (typically a title, but could also have e.g. `:2` to disambiguate multiple links in the same file)
+  - The label doesn't matter, it could be used as a display label. We'll just auto-generate something sensible. Could even be a user setting.
+- They would look like this:
+  ```
+  [[Some File]]
+  [[Some File:2]]
+  ```
+- The link ref definitions become equivalent to import statements in a programming language, telling us where we are importing and what:
+  `[Some File]: some-file.md: "Some file in root directory"`
+  `[Some File:2]: subdirectory/some-file.md: "Some file in a subdirectory"`
+- Here the title component of the link ref definition becomes optional since display name can be controlled by the label inside link.   
+- Note that we are adding .md here by default, but can omit that based on user preference as per @jojanaho's proposal
+  - Fixes GitHub UI rendering
+  - Breaks GitHub Pages links, but we'll override links in JavaScript as we do titles right now
+  - Going forward, Gatsby or other build tool can consume these directly
+- This will also help us with anchor links, because we don't need to put them inline in the label anymore:
+  - `[Some File With Anchor Section]: some-file.md#anchor`
+- This would also allow us to be more permissive of file names
+  - Markdown labels, the target needs to be a valid URI so we may needs to URI encode
+  - `[Some File With Anchor Section]: Some%20File%20With%20Anchor%20Section.md`
+- For migrating into this structure, we'll need to do a bit of guesswork and/or manual editing, but we can build tooling to make this easier (e.g. in `foam-cli janitor/migrate`) and ensure lint consistency with `foam-cli lint`
+  - In VS Code we can surface any issues in the Problems/Diagnostic pane
+  - We could also show syntax errors (red squigglies) inline, with a `Quick Fix...` menu
+- This all means that we will have to severely change Markdown Notes or replace it
+- Take into account @digiguru's work on supporting accents etc.
+
+## Plan
+
+- CLI team is finishing first iteration of foam-cli janitor/migrate/lint
+  - This will not affect their work, and we'll ensure their work can be used in the new structure
+  - Link ref def parsing is already on their branch, let's try to land that ASAP
+- Fix the .md generation @riccardo
+  - Change the way the links are generated in vs code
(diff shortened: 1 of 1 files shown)


docs/proposal-link-definition-and-disambiguation.md
```diff
@@ -0,0 +1,51 @@
+# Proposal: Link Definition and Disambiguation
+
+## Current status
+
+- We generate link ref defs from wiki-links inline
+- Wiki links are expected to be the full source of truth, but we don't have enough infomration to conclusively disambiguate
```
mathieudutour, line 6:
```suggestion
- Wiki links are expected to be the full source of truth, but we don't have enough information to conclusively disambiguate
```

docs/proposal-link-definition-and-disambiguation.md
```diff
@@ -0,0 +1,51 @@
+# Proposal: Link Definition and Disambiguation
+
+## Current status
+
+- We generate link ref defs from wiki-links inline
+- Wiki links are expected to be the full source of truth, but we don't have enough information to conclusively disambiguate
```
jojanaho, line 6: Slapping comments here just by reading this from top to bottom, at this point got a question "disabiguate what"?

docs/proposal-link-definition-and-disambiguation.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Solution
+
+- Don't wipe out link ref defs. Only add/remove as needed. This can be our metadata for disambiguation.
+
+Let's say we have two documents in the workspace with the same title "Some Title"
+
+- We would have autocompletion base on title of the document
```
jojanaho, line 14: could autocomplete always be done with a combination of title and file path within the workspace - in a style of "My title - foo/bar/my-title.md"

docs/proposal-link-definition-and-disambiguation.md
jevakallio: Pushed a disambiguation to this disambiguation :)

docs/proposal-link-definition-and-disambiguation.md
```diff
@@ -0,0 +1,51 @@
...
+  - This generates a link (typically a title, but could also have e.g. `:2` to disambiguate multiple links in the same file)
```
jojanaho, line 15: Personally I would prefer the local links to use style [[foo/bar/my-title]] - would this be configurable?

jevakallio: @jojanaho, to clarify, do you mean you prefer `foo/bar/my-title` for just the disambiguated links in case of title conflicts, or always?

It could be configurable, since it's an arbitrary label. Options could include:
- file-name.md
- Title Cased File Name
- Document Title
- Arbitrary Label you write yourself

jojanaho: Always, since in my case it would improve readability - e.g [[book-summaries/war-and-peace]]

nstafie:
If links are being revamped let's explicitly support aliasing too, something like `[[some really long link|alias]]`

docs/proposal-link-definition-and-disambiguation.md
jojanaho: And to be exact, we have the freedom for arbitrary labels only when link ref defs are in use :)

jojanaho:
> If links are being revamped let's explicitly support aliasing too, something like `[[some really long link|alias]]`

Wouldn't note aliases be in the note frontmatter? Or do you mean alternative title for a link?

nstafie:
> Wouldn't note aliases be in the note frontmatter? Or do you mean alternative title for a link?

Alternative title for a link is a good way of putting it. For example, say I have a file called `integration-methods.md` and I want to use it in a sentence, I would do `To solve this problem, you must first [[integration-methods|integrate]] this equation`.

It's easier to read and allows the text to flow more naturally.

jevakallio:
> I would do `To solve this problem, you must first [[integration-methods|integrate]] this equation`.
> 
> It's easier to read and allows the text to flow more naturally.

I'm not very opinionated either way, but if we're talking about reading-friendliness, I'd rather put the bit that's meant to be read up-front: "you must first [[integrate:integration-methods]] this equation" -- that way my eyes can jump to the end of the block once I hit the delimiter. 

Same could be achieved with syntax highlighting too.

Just thinking out loud.

jevakallio:
@jojanaho and I had a discussion about this. I'm dropping our rough notes here, will integrate these as part of the proposal when I have time, but for now just to get @riccardoferretti up to speed.

Ideally, we would like to be able to support both style links
- [[Note title]]
  - In this case, disambiguation is necessary via link definitions
- [[literal-reference-to-file]], or [[path/to/literal-reference-to-file]] 
  - In this case, disambiguation is not necessary
  - As per [Link Reference Definition Improvements](https://foambubble.github.io/foam/link-reference-definition-improvements) spec, we should aim to support a "linkReferenceDefinitions: Off" approach

Now, I don't want to play the "please everyone" game, but I feel like these are both completely valid use cases, and I think actually not very complicated to implement:

## Write time
- Always show autocomplete menu items in form of: **Title** *path/to/file*
  - The autocomplete list can be filtered by matching parts of either title or file name
- When autocomplete item is selected, expand the link label to either [[Title]] or [[path/to/file]] depending  based on user preference
- Generate link definitions (or not) based on user preference

## Read time

This same logic should work when constructing a graph, and when resolving the link target (e.g. Cmd+Click following the link):

- Check if the link has a corresponding reference definition
  - If it does, use the "target" from there
  - If not, treat the link label as literal path (extension optional)
  - If not found, warn the user (via linting)

---

I think we need to do a bit more thinking about this, but this would be an ideal outcome, if it doesn't introduce too much complexity

jojanaho:
> I think we need to write a **specification**  ... but if anyone else wants to take a stab of pulling a draft of such a specification document together, that would be fantastic and I can provide inputs for it

I started to work on this, just first trying to write down all the aspects to consider

jevakallio:
Closing in favour of https://github.com/foambubble/rfcs/pull/3

PR #123, opened 2020-07-15
Clarify how the repository should be opened in VS Code

nstafie:
See https://github.com/foambubble/foam-template/pull/10 for more details.

Figured the change should be reflected in both places because consistency is key :)


jevakallio:
@all-contributors add @nstafie for doc 🎉

PR #132, opened 2020-07-16
Add recipe for capturing notes from Drafts app on iOS

synesthesia:
This recipe shows how to use the popular Drafts app on iPhone to capture notes and push to user's Foam repo.


jevakallio:
@all-contributors add @synesthesia for docs! 🥳

PR #133, opened 2020-07-16
Paths in link definition section now include file extension

riccardoferretti:
PR looks big because I also updated the docs with the new section.
First commit contains the code changes
Second commit contains the updated link definition sections for all docs

2 changes:
- link definitions now include `.md` extension => this allows wiki navigation within github
- `.md` extension is stripped via JS in github pages site => this allows regular (html => html) navigation within github published page site (not removing the extension would have resulted in accessing the original, uncompiled markdown file)

For more context see https://foambubble.github.io/foam/link-reference-definition-improvements#wikilinks-dont-work-on-github

@jojanaho this should be parallel to what we have been chatting in the `foam-core` channel, but I want to be sure you are in the loop in case I am missing something


riccardoferretti:
note: I have tested locally the github generated site (based on https://docs.github.com/en/enterprise/2.14/user/articles/setting-up-your-github-pages-site-locally-with-jekyll) but haven't tested the navigation in github itself

docs/_layouts/home.html
```diff
@@ -12,6 +12,9 @@
     .querySelectorAll(".markdown-body a[title]:not([href^=http])")
     .forEach((a) => {
       a.innerText = a.title;
+      if (a.href.endsWith('.md') || a.href.endsWith('.markdown')) {
```
jojanaho, line 15: Markdown file extensions would probably be a good candidate to include into user config? Default values could maybe resemble this https://superuser.com/a/285878 or some subset of them?

docs/_layouts/page.html
```diff
@@ -10,6 +10,9 @@
     .querySelectorAll(".markdown-body a[title]:not([href^=http])")
     .forEach((a) => {
       a.innerText = a.title;
+      if (a.href.endsWith('.md') || a.href.endsWith('.markdown')) {
```
jojanaho, line 13: the same as above; maybe create an utility function to check if file is (or at least suspected to be) markdown?

jojanaho:
Maybe the following (or something like them) could be added into user settings

```json
"foam.defaultMarkdownExtension": {
  "type": "string",
  "default": "md",
  "description": "When creating new notes, this extension is appended to the filename."
},
"foam.acceptedMarkdownExtensions": {
  "type": "array",
  "items": {
    "type": "string"
  },
  "default": ["markdown", "mdown", "mkdn", "md", "mkd", "mdwn", "mdtxt", "mdtext"],
  "description": "List of accepted markdown file extensions. Only files having one of these file extensions are indexed by Foam."
}
```

These could then drive the implementation.

docs/_layouts/home.html
jevakallio: If we are going towards allow-listing markdown extensions, `.mdx` should definitely be there.

I know some of the other extensions in the ecosystem don't support this yet, but we want to be future compatible.

riccardoferretti: ok, pushed a couple of changes to address those concerns. TIL about Jekyll templates :)

docs/_layouts/foam.html
```diff
@@ -0,0 +1,32 @@
+---
+layout: default
+---
+
+<script type="text/javascript">
+var MD_EXT = ['.md', '.markdown', '.mdown', '.mkdn', '.mkd', '.mdwn', '.mdtxt', '.mdtext', '.text', '.Rmd'];
```
jojanaho: Oh yes, this was in a html-file used for publishing, so cannot really easily use package.json. In perfect world we would have tooling to read this out from the package.json and output it here (or similar mechanics). For now I think this is quite good. One thing we still should change - not necessarely in this PR though - is the file extension use in the vscode-extension. I was able to find at least this line:

```typescript
(f) => f.scheme === "file" && f.path.match(/\.(md|mdx|markdown)/i)
```

that should be aligned as well. I would expect that line to be constructed based on the user settings. For this MD_EXT fragment I'd add a disclaimer to synchronize the file extensions with the other places in the codebase.

riccardoferretti: sounds good, will add the disclaimer

jevakallio:
Just a quick note, overall this PR looks good, but we need to be strategic about how/when to deploy this, as it requires a the bit of javascript to be added to peoples' project templates, and may break other existing custom static site generation setups.

We should fix this sooner rather than later since more people are initing projects with the old template, which means more people will inherit this legacy approach. At the same time, I want to make this type of breaking change exactly once (if I can help it), so we need to do it properly.

As a first stab, we should:
- Upon upgrade to first version that has this feature, we should automatically offer to upgrade people's project templates if they're on vanilla setups (check _layouts files to ensure there are no local modifcations, and then override them)
- If they are on non-standard setups, link them to a document that describes what has changed

Possibly, I'd like to do this simultaneously with some of the other [link ref def improvements](https://foambubble.github.io/foam/link-reference-definition-improvements) such as removing the [[//begin]] and [[//end]] blocks.

Let's not merge this until we've figured out how to do this right.

jevakallio:
Here' my proposal to get this landed:

1. Put the file extension generation behind a user preference as per @jojanaho's [proposal](https://foambubble.github.io/foam/link-reference-definition-improvements):
    ```
    foam.edit.linkReferenceDefinitions: "withExtensions"
    ``` 
    And presume that if not defined, the default is
    ```
    foam.edit.linkReferenceDefinitions: "withoutExtensions"
    ```   
    To preserve the current behaviour until the user migrates.

2.  Add instructions on how to migrate to the release notes, and circulate it in the Foam docs/discord to make sure people ara aware of it.

3. Build an automated "codemod" that is able to:
    - Turn on the `withExtensions` setting
    - Regenerate all link reference definitions (this functionality already exists in the [janitor PR](https://github.com/foambubble/foam/pull/112)
    - Update the `_layouts/*.html` templates if they have not been edited (i.e. they correspond to one of the. historical versions in Foam template)

4. Offer to run this codemod automatically the first time Foam is used after updating to new version of the extension.

5. If we are not able to run codemod (e.g. people are on a non-standard template), link them to the release notes that explains how  they can enable the feature manually and what needs to be fixed.

packages/foam-vscode/src/features/wikilink-reference-generation.ts
```diff
@@ -97,13 +97,24 @@ async function updateReferenceList(foam: NoteGraph) {
...
+  withoutExtensions = "withoutExtensions",
+}
+
 async function generateReferenceList(foam: NoteGraph, doc: TextDocument): Promise<string[]> {
   const filePath = doc.fileName;
 
   const id = dropExtension(basename(filePath));
 
+  const linkDefinitionSetting: LinkReferenceDefinitionsSetting = workspace
+    .getConfiguration("foam.edit")
```
jevakallio, line 111: What is `foam.edit` is not defined? Will this return null and crash, or return an empty config object?

riccardoferretti: good point, I am not sure how it works in vscode. it should never be undefined, as we are defining it in the `package.json`, but will look into how configuration works in vscode a bit more closely

PR #136, opened 2020-07-18
Initialise note.title only if heading exists

### LINKED ISSUE #131: Initialise note title only if heading exists
From https://github.com/foambubble/foam/pull/112#discussion_r455400557

We need #112 to merge soon. Hence, I decided to create a new issue for this change. 

I am self assigning this.

anku255:
fixes #131 

As discussed in [#112](https://github.com/foambubble/foam/pull/112#discussion_r455403651)  I have created this PR.

**Note:**. #112 will have to be updated if this gets merged.


package.json
```diff
@@ -10,7 +10,10 @@
     "packages/*"
   ],
   "scripts": {
-    "watch": "yarn workspace foam-vscode watch"
+    "watch:vscode": "yarn workspace foam-vscode watch",
```
anku255, line 13: I added these as it was easier to run scripts from the root directory. Happy to remove if this is not a good practice.

packages/foam-core/src/markdown-provider.ts
```diff
@@ -91,7 +91,7 @@ export function createMarkdownReferences(
       return {
         linkText: link.to,
         wikiLink: pathToNote,
-        pageTitle: target.title,
+        pageTitle: target.title || target.id,
```
anku255, line 94: This is the only place where `note.title` was used directly.  `pageTitle` is initialized by `note.id` if heading doesn't exist.

packages/foam-core/test/markdown-provider.test.ts
```diff
@@ -49,6 +53,24 @@ describe('Markdown loader', () => {
...
+    const graph = new NoteGraph();
+    graph.setNote(createNoteFromMarkdown('page-a', pageA));
+
+    const pageANoteTitle = (graph.getNote('page-a') as Note).title;
+    expect(pageANoteTitle).toBe('Page A');
+  });
+
+  it('should not initialize note title if heading does not exist', () => {
+    const graph = new NoteGraph();
+    graph.setNote(createNoteFromMarkdown('page-d', pageD));
```
riccardoferretti, line 67: do you need to add the note to the graph to run this test? (not adding it might also spare you from the casting below)

or is there something else that you are trying to test as well?

package.json
riccardoferretti: looks good to me!

packages/foam-core/test/markdown-provider.test.ts
jevakallio: I guess it's not strictly necessary, but I don't mind doing a little integration testing here to exercise the setNote/getNote to make sure they work on notes without titles, so I'm fine with this.

PR #144, opened 2020-07-21
DX: repo-wide build tasks

riccardoferretti:
Adding a few build tasks across packages and standardizing on naming.
```
clean: remove output from build process
build: build the package
watch: build and watch for changes
test: run tests
lint: run lint
```

That means we can run any of those tasks from the root and have them executed everywhere. (I found it especially useful with `watch` to monitor both changes in `foam-core` and `foam-vscode`).

Note:
- `watch` is not defined in `foam-cli` as I don't think it's needed there
- `lint` has a placeholder in `foam-cli` because it's not there yet, but probably should
- `test` has a placeholder in `foam-vscode` because it's not there yet, but probably should

p.s. while testing the setup it found a small compilation error, so I fixed that too

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,12 @@
     "watch:vscode": "yarn workspace foam-vscode watch",
     "build:core": "yarn workspace foam-core build",
     "watch:core": "yarn workspace foam-core start",
-    "test:core": "yarn workspace foam-core test"
+    "test:core": "yarn workspace foam-core test",
+    "clean": "lerna run clean",
+    "build": "lerna run build",
+    "test": "lerna run test",
+    "lint": "lerna run lint",
+    "watch": "lerna run watch --concurrency 20 --stream"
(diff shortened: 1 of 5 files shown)


jevakallio:
LGTM! Thanks @riccardoferretti! 

Just a note that the downside of using `lerna run` to run multi-project tests is that it loses colors, formatting etc. and may cause confusing errors if multiple projects error out. 

At some point might be worth looking at [Jest multi-project runner](https://jestjs.io/blog/2017/05/06/jest-20-delightful-testing-multi-project-runner#multi-project-runner--configuration-overhaul) (or whatever the latest way of doing the same may be), but for now the lerna approach is suitable, and much improved compared to the current experience.

Thank you!

riccardoferretti:
I wasn't aware of jest multi-project, good to know!

PR #145, opened 2020-07-21
add GH workflows

mathieudutour:
- [x] Add a few GitHub actions to run when the packages changes. Each workflow is only run when the package associated changes.


jevakallio:
@mathieudutour this looks good, thanks for doing this! I don't have much experience with GitHub actions, but just wondering, should the have ran as part of this PR (or does the config need to exist on master first)? I'm not sure how to test this PR without first landing it.

I'm tempted to just use CircleCI since it's what I'm most familiar with, but would be nice to learn more about GH actions so happy to test it out!

Also, is moving CONTRIBUTING.md to .github necessary? I thought it's valid and detectable in project root too?

mathieudutour:
> should the have ran as part of this PR (or does the config need to exist on master first)?

Not this PR since it doesn't change any of the packages

> I'm not sure how to test this PR without first landing it

I could remove the limitation around package changes and then roll it back?

> is moving CONTRIBUTING.md to .github necessary? I thought it's valid and detectable in project root too?

Oh you're right, habits. I usually try to put as few as possible files at the root to quickly get to the readme

.github/workflows/foam-vscode.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Lint foam-vscode
+        run: yarn workspace foam-vscode lint
+      # - name: Test foam-vscode
+      #   run: yarn workspace foam-vscode test
+      # - name: Publish foam-vscode
+      #   if: github.ref == 'refs/heads/master'
+      #   run: yarn workspace foam-vscode publish-extension
+      #   with:
+      #     vsce_token: ${{ secrets.VSCE_TOKEN }}
```
jsjoeio, line 29: I assume this is commented out, but left in intentionally because we'll soon be using it?

mathieudutour: Yeah I'm thinking it could be a good idea to publish it automatically when it lands on master

jsjoeio: That would be amazing! (I've been waiting for the Daily Notes feature 😅)

jevakallio: @mathieudutour @jsjoeio agreed, auto deploy from master would be dope. I'm not 100% sure how I feel about that without having any kind of vs code e2e test coverage -- maybe we could autopublish to an alpha channel for the brave (under different extension id), and then do stable releases ~weekly?

mathieudutour: Yeah that sounds good

jsjoeio: Sounds good to me too

PR #155, opened 2020-07-24
Implement janitor command to extension

chirag-singhal:
This Pull Request adds the `vs-code janitor` command to extension

It does the following: 

- Adds heading to the files if they don't exist
- Adds/Update/Removes link references


packages/foam-vscode/src/features/janitor.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+  } catch (e) {
+    window.showErrorMessage(`Foam Janitor attempted to clean your workspace but ran into an error. Please check that we didn't break anything before committing any changes to version control, and pass the following error message to the Foam team on GitHub issues:
+    ${e.message}
+    ${e.stackTrace}`);
+  }
+}
+
+async function runJanitor() {
+  const foam = await bootstrap(getConfig());
```
riccardoferretti, line 54: `foam` is passed in the `activate` function of the feature, so that extensions don't need to bootstrap individually. in the `activate` function we should wait for the promise, then pass `foam` as a param in this function

jevakallio: @riccardoferretti ha yeah, I noticed this myself and already fixed it sneakily on master:
https://github.com/foambubble/foam/commit/7c6874d52844cffc9936efd056848f7e9dd7ce9c

Issue #160, opened 2020-07-25
Daily note not resolved by VSCode


MehraAkshay:
For some reason, using Alt + D or "Open Daily Note" with 0.3.0 leads to the following error:
```
d:%5CFoamWorkspaces%5CFirstFoamWorkspace%5C2020-07-25.md. Detail: Unable to resolve resource d:%5CFoamWorkspaces%5CFirstFoamWorkspace%5C2020-07-25.md
```
Note that the file _is_ created but that VSCode cannot locate it. In settings, I've made the path "." to no avail:
```
"foam.openDailyNote.directory": "."
```

ricalamino:
Same problem here.

jevakallio:
@MehraAkshay weird! Can you provide system information (VS Code version, operating system)? Looks like you're on Windows, and there's something funky going on with the way we resolve paths (the error message has URI encoded path separators, which doesn't look normal).

@ricalamino could you also provide system info.

If that is indeed the case, I don't have a Windows box at hand so can't test myself right now, but hopefully one of our other contributors can!

MehraAkshay:
@jevakallio Definitely! 

Here's the partial "about" output from VSCode: 

Version: 1.47.2 (user setup)
OS: Windows_NT x64 10.0.18363

jevakallio:
Merged #167 and released 0.3.1 https://github.com/foambubble/foam/releases/tag/v0.3.1

FYI @MehraAkshay @ricalamino @d2s  -- can you please test. I'll close this once the fix has been verified!

MehraAkshay:
@jevakallio Just installed 0.3.1 on my windows VSCode and Alt+D/"Open Daily Note" works as expected! Same for Mac. I can't speak to operation on a Linux machine.

Thank you!

ricalamino:
Working perfectly now in the new version!

Thank you!

vitaly-pevgonen:
Can confirm on Windows box: didn't work on v0.3.0 (it could create note, but could not open it; in error message backslashes in file path were replaced with %5C), works on v0.3.1.

jevakallio:
Thanks for reporting and fixing @MehraAkshay 🙇

PR #167, opened 2020-07-26
fixed issue with opening daily note on windows

MehraAkshay:
Fixed issue with opening daily note on windows ( #160 )

Seems like `uri.parse` screws up file paths on windows. `uri.file` fixes this issue.


jevakallio:
@all-contributors add @MehraAkshay for code

PR #186, opened 2020-07-28
test: use npx

SheetJSDev:
Running `npm run test` or `lerna exec npm run test` from a fresh clone fails because the foam-core package.json assumes a global install of `tsdx`:

```
yarn run v1.22.4
$ tsdx test
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run test stderr:
/bin/sh: tsdx: command not found
error Command failed with exit code 127.

lerna ERR! yarn run test exited 127 in 'foam-core'
```

Using `npx` obviates the issue.

### DIFF
--- a/packages/foam-core/package.json
+++ b/packages/foam-core/package.json
@@ -9,11 +9,11 @@
   ],
   "scripts": {
     "clean": "rimraf dist",
-    "build": "tsdx build",
-    "test": "tsdx test",
-    "lint": "tsdx lint",
-    "watch": "tsdx watch",
-    "prepare": "tsdx build"
+    "build": "npx tsdx build",
+    "test": "npx tsdx test",
+    "lint": "npx tsdx lint",
+    "watch": "npx tsdx watch",
+    "prepare": "npx tsdx build"
   },
   "devDependencies": {
     "@types/github-slugger": "^1.3.0",


riccardoferretti:
Hi @SheetJSDev, good job spotting the issue. I am not super familiar with using `npx`, I was just wondering whether we could fix the problem by adding `tsdx` to the `devDependencies`, similarly to how e.g. we do for `typescript`. 
That would also ensure that the same version is used across systems (not sure how that would work via `npx`).
What do you think?

SheetJSDev:
That line already exists https://github.com/foambubble/foam/blob/master/packages/foam-core/package.json#L23

`npx` is cross-platform, integrated with `npm` starting with npm 5.2.0 (from 2017), and referenced from [`webpack`](https://webpack.js.org/guides/getting-started/) documentation (so it's not a crazy change)

That said, there may be something else at play (possibly related to lerna)

riccardoferretti:
Mmm... that's a good point, I wonder why this is happening for tsdx and not typescript. 
I can do a test once I am at my laptop, could you please also share the steps to reproduce the issue?
Was this happening after running yarn?

SheetJSDev:
This is funny.  The issue is reproducible using `npm` from a fresh clone on a system without a global `tsdx` install:

```
git clone --depth=1 https://github.com/foambubble/foam
cd foam
npm i
npm run test
```

However using `yarn` works:

```
git clone --depth=1 https://github.com/foambubble/foam
cd foam
yarn
yarn test
```

Making this work with `npm` directly is probably not worth the effort at this point.

PR #187, opened 2020-07-29
revise vscode task- and launch-configs

jojanaho:
- Removed commands from sub-package:
  same can be achieved from the root
- Added build and no-build variants.
  In case separate watcher is already running,
  no need to build.
- Changed watch to build (no hot-reload in vscode by default)
- Show the build in console

Didn't update yarn.lock, seems to want to add
foam-core@0.3.0-alpha.0, but not really
related to this change

Note: the extension tests are failing due to
`Cannot find module '...foam-vscode/out/test/suite/index'`.
However that shouldn't be related to this change either.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,10 +6,9 @@
 	"version": "0.2.0",
 	"configurations": [
 		{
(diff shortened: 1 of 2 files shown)


jevakallio:
Commits merged via #193

PR #191, opened 2020-07-29
Parse frontmatter and use frontmatter.title as heading (fixes #174)

### LINKED ISSUE #174: foam-vscode: Janitor command adds the heading above the frontmatter
**Actual:**  The `Run: Janitor (Experimental)` command adds the heading above the frontmatter.

**Expected:** The generated headings should be below the frontmatter.

**Demo GIF:**

anku255:
## Summary

fixes #174 

This PR does the following things  - 

- [x] Add frontmatter metadata to the NoteGraph node.
- [x] Parse heading from the frontmatter and give it preference over the `#` or `=` heading. 
- [x] If the frontmatter exists but the title (heading) doesn't then generate the heading below the frontmatter.
- [x] Show a progress bar (`window.withProgress`) while the Janitor runs.

## Test Plan

I have tested with the following types of markdown notes.

**Input:**

```markdown
---
title: Note Title From Heading
---

Note content
```

**Output:** Doesn't generate the heading and uses `frontmatter.title` as `note.title`.

 **Input:**

`inbox.md`
```markdown
---
other: some other data
---

Note content
```

**Output:** 

```markdown
---
other: some other data
---

# Inbox

Note content
```

 **Input:**

`inbox.md`
```markdown
---
other: some other data
---
note content starts without a newline
```

**Output:**

```markdown
---
other: some other data
---

# Inbox

note content starts without a newline
```


packages/foam-core/src/markdown-provider.ts
```diff
@@ -27,15 +30,29 @@ export function createNoteFromMarkdown(
...
     return title === null ? CONTINUE : EXIT;
   });
+
   const links: NoteLink[] = [];
   const linkDefinitions: NoteLinkDefinition[] = [];
+  let frontmatter: any = null;
+  let start: Point = { line: 1, column: 1, offset: 0 }; // start position of the note
   visit(tree, node => {
+    if (node.type === 'yaml') {
+      frontmatter = parseYAML(node.value as string);
```
anku255, line 47: I am using the `yaml` package for parsing the frontmatter. Is that okay?

jevakallio: The only downside is that it's 27Kb minzipped, and theoretically foam-core could be used in a browser environment, so it adds a bit of chonk to the bundle.

I don't really see a good alternative right now, since I don't want to implement a partial yaml subset parser

riccardoferretti: yeah, I'd say it's probably ok for now, we might want to revisit for optimizations once as part of the browser release

packages/foam-core/src/note-graph.ts
```diff
@@ -26,30 +26,36 @@ export interface NoteLinkDefinition {
 
 export class Note {
   public id: ID;
+  public frontmatter: object | null;
```
riccardoferretti, line 29: is there a reason to have `null` instead of simply an empty object when there are no properties? dealing with `null` is always a bit of a pain :)

packages/foam-vscode/src/features/janitor.ts
```diff
@@ -80,11 +91,9 @@ async function runJanitor(foam: Foam) {
   // Apply Text Edits to Non Dirty Notes using fs module just like CLI
 
   const fileWritePromises = nonDirtyNotes.map(note => {
-    processedFileCount += 1;
-
     let heading = generateHeading(note);
     if (heading) {
-      console.log("fs.write heading " + note.path + " " + note.title);
+      console.log("fs.write heading " + note.path + " " + note.id);
```
riccardoferretti, line 96: what's this `console` statement needed for? debug purposes, info, or something else?

packages/foam-vscode/src/features/janitor.ts
```diff
@@ -149,14 +156,14 @@ async function runJanitor(foam: Foam) {
...
 
         if (heading) {
-          console.log("editor.write heading " + note.title);
+          console.log("editor.write heading " + note.id);
 
           updatedHeadingCount += 1;
           const start = new Position(
-            heading.range.start.line,
-            heading.range.start.column
+            heading.range.start.line - 1,
```
riccardoferretti, line 163: out of curiosity, what has changed that caused this?

anku255: The AST node positions (line and column) are `1-indexed` while VS Code APIs are `0-indexed`.

To maintain this consistency I am initializing `note.start` by 1-indexed values - 

https://github.com/foambubble/foam/blob/59eb48abd7ec086dbed1a600c3f76fde0fe4d5e6/packages/foam-core/src/markdown-provider.ts#L44

packages/foam-vscode/src/features/janitor.ts
anku255: It was already there. I am happy to remove them. I don't see why they need to be there permanently.

packages/foam-core/src/note-graph.ts
anku255: Will change it :)

anku255:
> @anku255 I was testing this and managed to break this when the frontmatter wasn't fully valid yaml:
> ...
> ...



@jevakallio  I don't think the invalid `yaml` is the issue here.  If the `yaml` is invalid then you should see this error - 

```
Command 'Foam: Run Janitor (Experimental)' resulted in an error (Implicit map keys need to be followed by map values)
```

I think this problem is related to inconsistent `NoteGraph`.  For some reason, the file changes are not reflected properly while running the Janitor. 

The Janitor command works fine on a freshly loaded VS Code window but if you make changes to some files (be it dirty or nondirty), Janitor may produce unexpected results.


To reproduce this you can try the following steps: -

1. Have a file with invalid frontmatter. For example

```markdown
---
title: Note Title
foo
----
```

2. Launch Extension Host (F5).

3. Run Janitor. (It should show the invalid frontmatter error).

4. Make the frontmatter valid and (save the file). 

```markdown
---
title: Note Title
---
```

5. Run Janitor again. (It still shows the invalid frontmatter error).




Do you have any idea why this might be happening? My initial understanding is that the `NoteGraph` is not updated after every file change. 

The `createNoteFromMarkdown` is only called on the first execution of `Janitor Command` but not on the subsequent runs. I am not sure if this is what's happening but my `console.log` debugging proves this.

packages/foam-vscode/src/features/janitor.ts
riccardoferretti: we could add a small utility function called `astPositionToVsCodePosition` which does this simple operation. 
it would both explicitly call out the intent of the operation, centralize the logic, and clean up the code on the client (caller) side.
Basically the code above would become:
```
const start = astPostionToVsCodePosition(heading.range.start)
```
(if necessary you can also create the reverse `vsCodePositionToAstPosition`, for the same reasons)

wdyt?

packages/foam-core/test/markdown-provider.test.ts
```diff
@@ -72,6 +81,31 @@ describe('Note Title', () => {
...
+    const graph = new NoteGraph();
+    graph.setNote(createNoteFromMarkdown('page-e', pageE, '\n'));
+
+    const pageENoteTitle = (graph.getNote('page-e') as Note).title;
+    expect(pageENoteTitle).toBe('Note Title');
+  });
+});
+
+describe('frontmatter', () => {
+  it('should parse yaml frontmatter', () => {
```
riccardoferretti, line 95: we should probably have a case for the empty front matter and the invalid front matter (I don't quite know how to deal with the latter, that test will force us to answer that question :) )

packages/foam-core/test/core.test.ts
```diff
@@ -1,24 +1,58 @@
 import { NoteGraph, Note } from '../src/note-graph';
 
 const position = {
-  start: { line: 0, column: 0 },
+  start: { line: 1, column: 1 },
   end: { line: 0, column: 0 },
```
riccardoferretti, line 5: should this be ` { line: 1, column: 1 }`?

anku255:
> Thanks @anku255 for the explanation. Seems like the root cause for this issue is already on master and needs to be addressed separately

Thank you for your approval. But don't merge it just yet. I will be doing the changes suggested by @riccardoferretti soon.

riccardoferretti:
> > Thanks @anku255 for the explanation. Seems like the root cause for this issue is already on master and needs to be addressed separately
> 
> Thank you for your approval. But don't merge it just yet. I will be doing the changes suggested by @riccardoferretti soon.

hey @anku255 we ended up merging this, but feel free to make the changes on another branch, or also directly on `refactor/note-id` if you think you'd run into conflicts, after all it's all minor things

PR #193, opened 2020-07-29
Fix VSCode Launch Task and Monorepo builds

jevakallio:
Builds upon and supercedes @jojanaho's work in #187. If this approach works, the other PR can be closed as this contains the commits from it.

The big change here is that foam-vscode build also now buils the foam-core package, so that we no longer need to run a separate watcher process for foam-core while developing.

This relies on the TSC "build mode", and "references" field, which we already had in place. (https://github.com/foambubble/foam/commit/eb22cd3055b360ec46d36bc7d85e4c974967a666)
https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript

This will necessitate us running the core in commonjs build mode that generates a commonjs entry point file instead of a `.cjs.development.js` bundle as before, which is fine for testing and deployment to node environments (such as vs code), but won't work in the browser, so foam-core needs a separate build config for UMD builds (https://github.com/foambubble/foam/commit/d64273829e3cc6885527d278ecf932fa7ebf26ea)

### First use per VSCode session

1. Open this repository in VS Code.
2. Run > Start Debugging
    - Expected outcome: Terminal shows "Starting compilation in watch mode..."
    - Expected outcome: Extension host process starts
3. Make modifications in foam-core files
    - Expected outcome: Terminal watcher shows that file modification was made
4. Restart extension host (Recycle button in debug menu)
    - Expected outcome: Changes made in step 3 were reflected

### Subsequent uses per VSCode session

1. Stop previously running Extension Host (Stop button in debug menu)
    - Expected outcome: Watcher process is still running in terminal
2. Run > Start debugging
    - Expected outcome: Extension host process starts immediately (no build needed)
3. Make modifications in foam-core files
    - Expected outcome: Terminal watcher shows that file modification was made
4. Restart extension host (Recycle button in debug menu)
    - Expected outcome: Changes made in step 3 were reflected


.vscode/launch.json
```diff
@@ -34,16 +34,7 @@
 			"outFiles": [
 				"${workspaceFolder}/packages/foam-vscode/out/test/**/*.js"
 			],
-			"preLaunchTask": "Build foam-vscode"
-		},
-		{
-			"type": "node",
-			"request": "launch",
-			"name": "Workspace Manager tests",
```
jojanaho, line 42: Keeping some kind of entry point to debug the core tests would be nice? We could put this also into foam-core launch.json, but it would probably be somewhat annoying to need to load that folder into separate vscode instance. Thus I'd still keep this here (and name it e.g. to "foam-core unit tests" or something similar.

jevakallio: Ah yes, I didn't think of debugging tests, I was making the assumption that we would just run tests from the command line but that is clearly insufficient!

jevakallio:
@jojanaho I've removed the duplicated .vscode directory entirely from packages/foam-vscode as not to encourage its use in workspace root.

I've also re-added the launch task for core tests, but they're not runnign. When are they expected to be ran? At the same time as the extension host? If so, I'm not seeing the tests execute, possibly because the watch mode terminal is foregrounded, and the test task actually runs in the background?

Feel free to try to fix that or merge now, and we can address it later. The default test task can still be ran, and works.

jojanaho:
> I've also re-added the launch task for core tests, but they're not running. When are they expected to be ran? At the same time as the extension host? If so, I'm not seeing the tests execute, possibly because the watch mode terminal is foregrounded, and the test task actually runs in the background?

I guess the main task for the foam-core launch target is to be able to debug the core unit tests; now I can set a breakpoint and land there, so that's great! What's not so great is that on debugger exit, we get:
```
...
illegal access
FATAL ERROR: 
node::inspector::Agent::ToggleAsyncHook Cannot toggle Inspector's AsyncHook, please report this.
...
```

With a quick search, I was able to find [this](https://github.com/nodejs/node/issues/27261#issuecomment-662615093) - so it might be related to a bug in Node. However that's not too critical, since we still can inspect the actual results of the test by running `yarn run test:core` from root, so I would just let this go for now. In case you want to see the crash message in vscode terminal, you can add the following into the "Test Core"-launch target:
```
"console": "integratedTerminal"
```

PR #204, opened 2020-08-02
feat(foam-vscode): add jest for testing

jsjoeio:
This adds all the necessary pieces for Jest to work properly in `foam-vscode`. It also enables testing to run as part of the GitHub workflow.


jsjoeio:
@jevakallio  Those are good points! The intention behind this was to have the bare minimal of Jest set up so that we could at least write unit tests.

> This means that any module that imports vscode can not be imported into a jest test without mocking the vscode module first.

Right! Thanks for bringing that up. My thinking was to add support for those types of integration and e2e tests later. But if you want that to all land as one unit in this PR, happy to do some investigation (thanks for the resources btw!).

jsjoeio:
After doing a quick investigation, it seems like mocking the `vscode` module is the most common approach. There is an [ongoing discussion](https://github.com/microsoft/vscode/issues/94746) about making this easier for folks. 

- Red Hat's [[vs-code]] [extension tester](https://github.com/redhat-developer/vscode-extension-tester)
- example [[vs-code]] [mock](https://github.com/gkalpak/aio-docs-utils/blob/master/src/test/helpers/vscode.mock.ts)
- end-to-end [[testing]] with [[vs-code]] [link](https://medium.com/@soloydenko/end-to-end-testing-vs-code-extensions-via-jest-828e5edfeb75)
- [open issue](https://github.com/microsoft/vscode-test/issues/37)
- [mock with jest](https://www.richardkotze.com/coding/unit-test-mock-vs-code-extension-api-jest)
- [another link](https://github.com/Almenon/AREPL-vscode/issues/295) with some resources

For now, I'm going to work on adding a mock then I'll re-request your review.

packages/foam-vscode/__mocks__/vscode.ts
```diff
@@ -0,0 +1,16 @@
...
+
+combined with advice from this GitHub issue comment:
+https://github.com/microsoft/vscode-test/issues/37#issuecomment-584744386
+*/
+
+const vscode = {
+    // Add values and methods as needed for tests
+};
+
+module.exports = vscode;
```
jsjoeio, line 16: This was enough to get the test you pushed to pass. Thoughts?

PR #208, opened 2020-08-03
Decoupled note ID from note slug

riccardoferretti:
With this change we can support files that have the same name (and therefore slug) across multiple directories.
(note that slugs are non-unique, but as we introduce a proper concept of ID that won't be an issue)


packages/foam-core/src/utils.ts
```diff
@@ -1,4 +1,27 @@
...
+  const filename = noteUri.split('/').slice(-1)[0].split('.').slice(0, -1).join('.');
+  return GithubSlugger.slug(filename);
+}
+
+export const hashURI = (uri: URI): ID => {
+  return hash(path.normalize(uri));
+}
+
+export const getUriViaRelative = (reference: URI, relativeSlug: string): URI => {
+  const slug = relativeSlug.endsWith('.md') // TODO hack
```
riccardoferretti, line 32: this is a hack, I am assuming an `md` extension, which of course we can't.
probably need to add a proper extension check, will do that while you take a look at the approach as a whole

packages/foam-core/src/markdown-provider.ts
```diff
@@ -167,9 +177,9 @@ export function createMarkdownReferences(
 
       // [wiki-link-text]: path/to/file.md "Page title"
       return {
-        label: link.text,
+        label: link.link.text,
         url: pathToNote,
-        title: target.title || target.id,
+        title: target.title || target.slug,
```
riccardoferretti, line 182: using the `slug` instead of the `id`

packages/foam-core/src/utils.ts
```diff
@@ -1,4 +1,39 @@
...
+    .slice(0, -1)
+    .join('.');
+  return GithubSlugger.slug(filename);
+};
+
+export const hashURI = (uri: URI): ID => {
+  return hash(path.normalize(uri));
+};
+
+export const getUriViaRelative = (
```
jevakallio, line 28: This method probably needs a better name, since on a first read I'm struggling to understand what it does. Will probably become clear on second read.

packages/foam-core/src/janitor/index.ts
```diff
@@ -87,22 +87,22 @@ export const generateHeading = (note: Note): TextEdit | null => {
...
+  const paddingStart = frontmatterExists ? note.source.eol : '';
   const paddingEnd = newLineExistsAfterFrontmatter
-    ? note.eol
-    : `${note.eol}${note.eol}`;
+    ? note.source.eol
+    : `${note.source.eol}${note.source.eol}`;
 
   return {
-    newText: `${paddingStart}# ${getHeadingFromFileName(note.id)}${paddingEnd}`,
+    newText: `${paddingStart}# ${getHeadingFromFileName(note.slug)}${paddingEnd}`,
```
jevakallio, line 102: GitHub actions still showing a few prettier-eslint issues here, nbd just fyi

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,162 @@
...
+  public getNoteId(query: NoteQuery): ID | null {
+    if (query.id) {
+      return query.id;
+    }
+    const searchFn = query.uri
+      ? (note: Note | null) => note?.source.uri === query.uri
+      : query.slug
+      ? (note: Note | null) => note?.slug === query.slug
+      : query.title
+      ? (note: Note | null) => note?.title === query.title
```
jevakallio, line 118: Are titles unique in the workspace? This seems like it could have clashes?

(Sorry if you saw and dismissed this comment earlier. I left it before but I think it got nuked as outdated by a push)

riccardoferretti: `id` and `uri` are unique. `slug` and `title` could have 0+ results. in case there is > 1 result, the first one is returned.
I needed something for `slug` and I thought it could be done is such a way as to allow to query the graph.
some comments:
- if we want to call out the difference between picking in the graph or scanning it, we could have 2 fns. I don't think that's necessary as it feels like an implementation detail (and we could improve the performance of various searches by adding maps)
- we could have a `listNotes(query: NoteQuery)` that returns an array for those situations where we actually want all the results (e.g. autocomplete)

of course we could also simply have `getByUri` and `getBySlug` (or `listBySlug` as they are not unique)

packages/foam-core/src/note-graph.ts
```diff
@@ -24,98 +49,112 @@ export interface NoteLinkDefinition {
...
+
 export class NoteGraph {
   private graph: Graph;
   private events: EventEmitter;
+  private createIdFromURI: (uri: URI) => ID;
 
   constructor() {
     this.graph = new Graph();
     this.events = new EventEmitter();
+    this.createIdFromURI = hashURI;
```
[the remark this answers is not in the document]
jojanaho: Any benefits in hashURI vs UUID? Just thinking that UUID would be more future proof when merging multiple foams and maybe even reusing their cache files. With hashURI uids might collide, with UUID most probably not.

packages/foam-core/src/utils.ts
```diff
@@ -1,4 +1,34 @@
+import path from 'path';
+import crypto from 'crypto';
 import { titleCase } from 'title-case';
+import GithubSlugger from 'github-slugger';
+import { URI, ID } from 'types';
+
+export const hash = (text: string) =>
+  crypto
+    .createHash('md5')
```
jojanaho, line 9: in case we keep using path hashes, might make sense to use sha1: https://medium.com/@chris_72272/what-is-the-fastest-node-js-hashing-algorithm-c15c1a0e164e . Also git uses sha1 IIRC.

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,162 @@
...
+  end: Point;
+  eol: string;
 }
 
-export interface NoteLink {
-  to: ID;
+export interface WikiLink {
+  type: 'wikilink';
   text: string;
+  slug: string;
```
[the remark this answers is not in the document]
riccardoferretti: yes that is a good point. in fact at the moment it is not, but the situation where this would be the case is:
`this is a [[the-greatest-day|great day]] for things`.
In the old code I was supporting this case too but I didn't port the feature to minimize scope creep. I forgot to fix the type, happy to do that if you guys think it might be better

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,162 @@
...
   position: Position;
 }
 
+export type NoteLink = WikiLink;
```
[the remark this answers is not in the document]
riccardoferretti: you are correct on all fronts, and probably it is an over-optimization which I am happy to roll back if it makes things easier. Basically the idea is that a `NoteLink` could be a `WikiLink`, a `RegularLink` (`[Duck Duck Go](https://duckduckgo.com)`), and possibly an `AttachmentLink` (`[An old rock in the desert`).
The `type` would help in the case of the union on `NoteLink` (kinda like actions in redux), the closest thing I know of pattern matching in js/ts.

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,162 @@
...
+
+export interface NoteInfo {
```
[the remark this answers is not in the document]
riccardoferretti: yes, there are two main things:
- `NoteInfo` is concerned with the parsing of the document. it is not aware of the graph and shouldn't make assumptions about it (e.g. in its links, it knows there is a destination, it doesn't know whether it actually exists)
- `Note` is the "connected document" within the graph. It is an elaboration/augmentation of `NoteInfo` in the context of the rest of the foam.

I actually think that the inheritance relationship might not be the best one, and instead they should have two possibly separate structure (even if as of today they only differ in the `id` and `slug`).

I was considering pulling the `NoteInfo` and related types in its own file. Let me point out that the relationship should only be `Note -> NoteInfo` and never the other way around (`NoteInfo` is unaware of the graph, that's how we can parse files in isolation)

I agree I should add some comments in the file.

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,162 @@
...
-    this.definitions = definitions;
-    this.start = start;
-    this.end = end;
-    this.eol = eol;
-  }
-}
-
 export type NoteGraphEventHandler = (e: { note: Note }) => void;
 
+export type NoteQuery = {
```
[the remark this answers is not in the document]
riccardoferretti: hehe you are right. the logic here is that there is a precedence: `id`, `uri`, `slug`, `title`.
atm in case of multiple parameters the above precedence is used to pick the criteria.

I think your idea of having `type` and `value` is interesting, will think about that more.
if I understand correctly it would look like: `getNote(by: <'id' | 'slug' | 'uri' | 'title'>, value: string)`

re more complex queries I would probably go the object approach, but I wonder if that would live in this fn or another (I am ambivalent about this, still need to think about it)

packages/foam-core/src/note-graph.ts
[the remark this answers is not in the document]
riccardoferretti: the major advantage of basing the ID of the URI, instead of using a UUID, is shown in line 90 of this file.
```
const targetId = this.createIdFromURI(targetPath);
```
If we use the UUID we wouldn't be able to generate a connection at this stage, unless the target has already been added to the graph and we would still need to use the URI to identify it.

Even in that case, given that we must assume that docs will have links to other docs that have not been added to the graph yet, we would need to add an explicit linking phase.

I don't have a huge issue around this, but it does make the loading of the graph more complicated, so I am thinking of waiting to make such a change until there is a tangible advantage in doing that

packages/foam-core/src/utils.ts
riccardoferretti: I am super easy about that, happy to change it unless @jevakallio has any objection

packages/foam-core/src/utils.ts
riccardoferretti: haha I totally agree :) feel free to suggest better names, will come up with something tomorrow

packages/foam-core/src/markdown-provider.ts
```diff
@@ -44,7 +48,7 @@ export function createNoteFromMarkdown(
   let start: Point = { line: 1, column: 1, offset: 0 }; // start position of the note
   visit(tree, node => {
     if (node.type === 'yaml') {
-      frontmatter = parseYAML(node.value as string);
+      frontmatter = parseYAML(node.value as string) ?? {}; // parseYAML returns null if the frontmatter is emtpy
```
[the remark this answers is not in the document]
riccardoferretti: cc @anku255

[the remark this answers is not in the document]
anku255: fixed!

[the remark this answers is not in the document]
riccardoferretti: Fun story:
I kept looking at the comment, and the code, and the comment.. and not understand what it was about.
Then I looked at it again, comment and code, and still nothing.
So I thought "let's see if Ankit understands this" because I clearly don't! (so I cc'ed him)
After he implemented the fix, I kept looking at the code, the comment, the code - it took me 3 more times to finally understand there was a typo 🤣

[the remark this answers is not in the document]
anku255: 🤣🤣🤣🤣

packages/foam-core/src/utils.ts
jevakallio: Feel free to change, I don't have any issues as these don't need to be secure hashes and I have never experienced collisions with sha.

Keeping in mind, this is in foam-core, so technically we might want to support using it in a browser based environment -- though I think we might have leaked a bunch of other node APIs in here as well. If we wanted to be good about this, could consider murmurhash.

packages/foam-core/src/note-graph.ts
jevakallio: The bit I don't like here is that it's too easy to accidentally link to the wrong note due to slug/title collision. Enshrining their non-uniqueness to the API would help avoid these accidents from future implementers.

I'm not prissy about the naming.

packages/foam-core/src/note-graph.ts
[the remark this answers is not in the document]
riccardoferretti: I removed `text` for now, we were not using and was causing confusion

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,138 @@
...
+  properties: object;
+  // sections: NoteSection[]
+  // tags: NoteTag[]
+  links: NoteLink[];
+  definitions: NoteLinkDefinition[];
+  source: NoteSource;
 }
 
-export type NoteGraphEventHandler = (e: { note: Note }) => void;
+export type GraphNote = Note & {
```
riccardoferretti, line 41: I have actually fixed this to make it clearer (unfortunately the previous comment got lost).

But basically now we have `GraphNote` and `GraphConnection`.
Should be easier to understand than before (`NoteInfo`, `Note` and `Connection`)

packages/foam-core/src/note-graph.ts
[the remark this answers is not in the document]
riccardoferretti: added comment

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,137 @@
...
+
+export interface GraphConnection {
+  from: ID;
+  to: ID;
+  link: NoteLink;
+}
+
+export type NoteGraphEventHandler = (e: { note: GraphNote }) => void;
+
+export type NotesQuery = { slug: string } | { title: string };
```
riccardoferretti, line 53: now the query is mutually exclusive (and will be used only in `getNotes`, see below)

packages/foam-core/src/note-graph.ts
```diff
@@ -1,121 +1,137 @@
...
     });
-
-    this.events.emit(noteExists ? 'update' : 'add', { note });
+    this.events.emit(noteExists ? 'update' : 'add', { note: graphNote });
+    return graphNote;
   }
 
-  public getNotes(): Note[] {
-    return this.graph.nodes().map(id => this.graph.node(id));
+  public getNotes(query?: NotesQuery): GraphNote[] {
```
riccardoferretti, line 95: we can add an optional query. it's backwards compatible, clear about the list semantics, and possibly extensible for the future

packages/foam-core/src/utils.ts
```diff
@@ -1,4 +1,34 @@
...
+
+export const uriToSlug = (noteUri: URI): string => {
+  return GithubSlugger.slug(path.parse(noteUri).name);
+};
+
+export const hashURI = (uri: URI): ID => {
+  return hash(path.normalize(uri));
+};
+
+export const computeRelativeUri = (
```
riccardoferretti, line 21: this is the name I came up with (unfortunately the previous comment was removed)

packages/foam-core/src/markdown-provider.ts
```diff
@@ -145,8 +146,17 @@ export function createMarkdownReferences(
   return graph
     .getForwardLinks(noteId)
     .map(link => {
-      const target = graph.getNote(link.to);
-
+      let target = graph.getNote(link.to);
+      // if we don't find the target by ID we search the graph by slug
+      if (!target) {
+        const candidates = graph.getNotes({ slug: link.link.slug });
```
riccardoferretti, line 152: we are now explicitly finding out how many notes exist with the given slug and can act on it (atm logging to console for consistency with the rest of the reporting in this fn)

packages/foam-core/src/note-graph.ts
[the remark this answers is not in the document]
jojanaho: maybe this is good enough for now; just want to point out that this makes ids now non-opaque (?), since we have semantics associated with them (the requirement we discussed earlier). In the long run I'd really like to see distinct linking phase after the full workspace load.

jojanaho:
Good job; last nitpick - I think there was still a few `is defined but never used` warnings in output

packages/foam-core/src/markdown-provider.ts
```diff
@@ -55,8 +54,8 @@ export function createNoteFromMarkdown(
 
     if (node.type === 'wikiLink') {
       links.push({
-        to: node.value as string,
-        text: node.value as string,
+        type: 'wikilink',
+        slug: node.value as string,
```
zbw8388, line 58: I'm wondering why there's only slug for each wikiLink? Currently slug is basically equivalent to the file name, meaning wikiLinks that are titles are not going to get identified. I think a small adjustment of this PR (namely `getNotes` (`note-graph.ts:95`)) will enable Foam to allow both [[Title Case]] and [[file-case]] links, as mentioned in this comment: https://github.com/foambubble/foam/pull/103#issuecomment-657257591

riccardoferretti: The way I approached it is to have different type of links for different situations we encounter in the md document. 
For instance, if we find a `[[my-note]]` vs `[[My Note]]` vs `[text](http://link)` link, we actually model them differently (as they are different things: the first one a standard wikilink, the second a title-based-wikilink (for lack of a better name), the last one a regular url link.
At this phase of the parsing we are just modeling the document (atm `Note` and `GraphNote` are basically the same as we don't need extra stuff, that might change in the future).

Of course other ways of doing this are possible (e.g. `wikilink` type could include both with some sort of qualifier), but atm I believe it's better to be more explicit. To be clear, I haven't run this with the rest of the team (except via the PR review), so it's not a "project strategy" more than it is my personal approach.

Related to #103 , another field that might be added in this specific type is `text` or `label` to support wikilinks in the format `[[my-note|this is the label]]`, we removed that from the original implementation because we don't parse it at the moment and it was a bit confusing. Same would apply for an anchor (`[[my-note#my-section]]`).

PR #218, opened 2020-08-09
Fixed `Note ${noteId} was not added to NoteGraph`

zbw8388:
After #208, id for each note in `NoteGraph` is hashed. However, in `generateReferenceList` in file `wikilink-reference-generation.ts`, the id is still given by the file path, which compromised the core functionality. This PR fixed this bug.

### DIFF
--- a/docs/principles.md
+++ b/docs/principles.md
@@ -36,7 +36,7 @@ This principle may seem like it contradicts [Foam wants you to own your thoughts
 ## Foam allows people to collaborate in discovering better ways to work, together.
 
 - **Foam is a collection of ideas.** Foam was released to the public not to share the few good ideas in it, but to learn many good ideas from others. As you improve your own workflow, share your work on your own Foam blog.
-- **Foam is open for contributions.** If you use a tool or workflow that you like that fits these principles, please contribute them back to the Foam template as [[recipes]], [[recommended-extensions]] or documentation in [this workspace](httpsL//github,com/foambubble/foam). See also: [[roadmap]] and [[contribution-guide]].
+- **Foam is open for contributions.** If you use a tool or workflow that you like that fits these principles, please contribute them back to the Foam template as [[recipes]], [[recommended-extensions]] or documentation in [this workspace](https://github.com/foambubble/foam). See also: [[roadmap]] and [[contribution-guide]].
 - **Foam is open source.** Feel free to fork it, improve it and remix it. Just don't sell it, as per our [license](license).
 - **Foam is not Roam.** This project was inspired by Roam Research, but we're not limited by what Roam does. No idea is too big (though if it doesn't fit with Foam's core workflow, we might make it a [[recipes]] page instead).
 


packages/foam-core/src/markdown-provider.ts
```diff
@@ -129,20 +134,25 @@ export function stringifyMarkdownLinkReferenceDefinition(
 }
 export function createMarkdownReferences(
   graph: NoteGraph,
-  noteId: ID,
+  inputNote: GraphNote,
```
riccardoferretti, line 137: was wondering, why are you passing the `Note` object instead of its `ID`?

packages/foam-vscode/src/features/wikilink-reference-generation.ts
```diff
@@ -135,10 +135,10 @@ async function generateReferenceList(
 ): Promise<string[]> {
   const filePath = doc.fileName;
 
-  const id = dropExtension(basename(filePath));
```
riccardoferretti, line 138: wow nice catch! I totally missed this in my code! Odd that TS didn't notice the type mismatch

packages/foam-vscode/src/features/wikilink-reference-generation.ts
```diff
@@ -135,10 +135,10 @@ async function generateReferenceList(
...
+  const note = foam.getNoteByURI(filePath);
 
   const references = uniq(
-    createMarkdownReferences(foam, id, includeExtensions()).map(
+    createMarkdownReferences(foam, note, includeExtensions()).map(
```
riccardoferretti, line 141: speaking of TS missing typing errors, if I am not mistaking `note` here is `Note | null` but the parameter needs to be `Note`.
I honestly don't believe it could ever be `null` in this code path, so I don't expect this to fail (but we should probably still check to be sure) - what I am concerned about is TS not reporting a straightforward type error

packages/foam-core/src/markdown-provider.ts
zbw8388: It seems like `createIdFromURI` in class `NoteGraph` is a private funtion, meaning the ability of turning `URI` to `ID` is restricted to `NoteGraph`. Instead of dealing with (private) hash value, I think it would be better to just pass the entire node object. An alternate solution would be to make the translation function public.

packages/foam-vscode/src/features/wikilink-reference-generation.ts
zbw8388: I found this on stackoverflow: https://stackoverflow.com/questions/51577959/is-it-possible-to-type-check-string-aliases-in-typescript

riccardoferretti: ahh got it!

packages/foam-core/src/markdown-provider.ts
riccardoferretti: yes you are right on both fronts regarding `createIdFromURI`, I think adding the function you did is a good approach.
What I was referring to is that I saw you create a `Note` in `generateReferenceList` (`wikilink-reference-generation.ts:138`). 
I think that you could pass `note.id` when calling this function a few lines later (line 141), but I saw you pass the whole `Note` object, so was wondering why.
I don't have a strong preference, was mostly curious to understand :)

zbw8388: Woops... I did not think about that... Thank you for point it out :) I will revert back to the original solution in the new PR.

PR #219, opened 2020-08-09
Alternative way to fix #218

zbw8388:
Fixed #218

### DIFF
--- a/packages/foam-vscode/src/features/wikilink-reference-generation.ts
+++ b/packages/foam-vscode/src/features/wikilink-reference-generation.ts
@@ -135,7 +135,7 @@ async function generateReferenceList(
 ): Promise<string[]> {
   const filePath = doc.fileName;
 
-  const id = dropExtension(basename(filePath));
+  const id = foam.createIdFromURI(filePath);
 
   const references = uniq(
     createMarkdownReferences(foam, id, includeExtensions()).map(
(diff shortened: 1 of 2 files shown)


packages/foam-vscode/src/features/wikilink-reference-generation.ts
```diff
@@ -128,14 +128,15 @@ async function updateReferenceList(foam: NoteGraph) {
...
 async function generateReferenceList(
   foam: NoteGraph,
   doc: TextDocument
 ): Promise<string[]> {
   const filePath = doc.fileName;
 
-  const id = dropExtension(basename(filePath));
+  const note = foam.getNoteByURI(filePath);
+
+  const id = note?.id ?? "";
```
riccardoferretti, line 139: (This is a minor semantic comment. Not a blocker to merge the PR :) )

If I follow the reasoning here, when the note is not found, you set the `id` to `""`, and you rely on `createMarkdownReferences` to trigger the error condition (note not found).

I would suggest you handle that case here (basically if the note doesn't exist, don't even call the function).
The reason I feel this is a more robust approach is that you can't rely on the error condition inside the function (e.g. you don't know whether it would throw or simply log an error), as that is not part of its API. 
By looking at the API of the function one would expect a valid ID to be passed it, so if we know beforehand we don't have one, we probably shouldn't call it.
Also, here you have more context for a better error message. Compare "Can't find note with ID ''", with "Can't find note for URI ${filePath}". So why not take advantage of it :)

Again, just a nitpick/opinion, PR is approved a good to be merged!

zbw8388: Thank you for pointing that out! It's really helpful for me.

PR #221, opened 2020-08-11
Improve wikilink styling

jevakallio:
Builds on #220 (Improve Wikilink detection)

Now that we have the power to detect wiki links more accurately, we can also improve their styling. We could remove the brackets altogether, but I believe it's useful to differentiate between in-brain wikilinks and external links.

This PR:
- Removes the surrounding brackets around wikilinks https://github.com/foambubble/foam/commit/97644056782502c9e8dc6cabb7dff0d5462b2b06
- Adds them back as styleable CSS pseudoelements https://github.com/foambubble/foam/commit/d68a751e2069b86d1cf0aacdb60c15e5bcbea315

Before:
https://foambubble.github.io/foam/#call-to-adventure
image

After:
https://jevakallio.github.io/foam/#call-to-adventure
image

The styling could be improved, if anyone has ideas. We can style the elements with background color, font weight, other pseudo-elements etc -- anything CSS can do.

Once this lands on Foam website, we can also apply these changes to the Foam template so other new projects can inherit them.

### DIFF
--- a/docs/_layouts/foam.html
+++ b/docs/_layouts/foam.html
@@ -19,10 +19,20 @@ window.addEventListener('DOMContentLoaded', (event) => {
   document
     .querySelectorAll(".markdown-body a[title]:not([href^=http])")
     .forEach((a) => {
-      // Hack: Replace page-link with "Page Title"...
-      a.innerText = a.title;
-      // ...and normalize the links to allow html pages navigation
-      a.href = normalizeMdLink(a.href);
+      // filter to only wiki-links
+      var prev = a.previousSibling;
+      var next = a.nextSibling;
+      if (
+        prev instanceof Text && prev.textContent.endsWith('[') && 
+        next instanceof Text && next.textContent.startsWith(']')
+      ) {
+
+        // replace page-link with "Page Title"...
+        a.innerText = a.title;
+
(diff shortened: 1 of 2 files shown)


ingalless:
A user in the discord asked how old projects can include this. I assume we will roll out a recipe/upgrade guide too?

riccardoferretti:
That's a really good question, we don't have a good way of pushing an update to the template, probably something worth addressing soon. @jevakallio do you have any thoughts around how to go about that?

martinlaws:
@riccardoferretti @jevakallio not sure if this is too unapproachable for non-developers but I'd be more than happy to add a recipe on how to rebase against a fork's upstream in order to update for the time being?

riccardoferretti:
@martinlaws I agree that a recipe to plug the functionality in the meantime is a good idea (I reckon the longer term solution needs to live somewhere in the CLI/vscode-extension). I would say go ahead and create a PR for this recipe and let's move any further conversation there. Thanks!

jevakallio:
@martinlaws, I agree with @riccardoferretti! I'd love a recipe for how to upgrade a template as a holdover until we implement a codemod type solution that can automate this. 

Ideally, a recipe should explain how to upgrade by hand by copying files over manually, and an git merge command for more programmer types.

PR #222, opened 2020-08-11
Fix issue with janitor crashing when file is only frontmatter

jevakallio:
In cases where a file contained frontmatter but no actual text, janitor would crash due to the contentStart line index being beyond the file's line count.

The main fix is to check whether line is whitespace by `===` comparison rather than `.length` check, but the diff is slightly larger due to readability improvements.


anku255:
@riccardoferretti I have added a test. Can you please take a look?

PR #223, opened 2020-08-11
Add support for basic link aliases

jevakallio:
Note: See [Problems](#problems) below.

## Summary

This PR adds basic support for using aliased link syntax `[[link|alias]]`. 

- The alias is used as the link title text in any tool that supports link reference definition titles.
- This work mirrors @thomaskoppelaar's work to add the same functionality to markdown notes in https://github.com/kortina/vscode-markdown-notes/pull/73.
- This work is happening out-of-band with the [Linking and Embedding proposal](https://github.com/foambubble/rfcs/pull/3), but it's fully compatible with the current line of thinking of piped aliases and should not introduce any new technical debt.

Input:
```md
- ✅ Aliases: [[inbox|link alias]]
- ✅ Different aliases to same document: [[inbox|different alias]]
- ✅ Aliases with spaces on either side of the pipe: [[inbox | yet another alias]]
- ✅ Links with no aliases [[inbox]]
```

Output:
```md
[//begin]: # "Autogenerated link references for markdown compatibility"
[inbox]: inbox "Inbox"
[inbox | yet another alias]: inbox "yet another alias"
[inbox|different alias]: inbox "different alias"
[inbox|link alias]: inbox "link alias"
[//end]: # "Autogenerated link references"
```

GitHub:
image

## Problems

On GitHub pages, piped wikilinks are getting incorrectly parsed to tables 🤦‍♂️

image

But only inside lists:

image

This appears to be an implementation issue in Jekyll and not with GFM. Here's a gist showing the same content rendering fine: https://gist.github.com/jevakallio/0085469e88c9f3e309bda204a1526c0f

### Possible solutions

We can add piped alias support, but it won't work on websites using the built-in Jekyll instance (unless we can customise it to not parse the links as tables, so it may not make sense to merge this work until we know how to fix it.

- Use an alternative aliasing character ([[remark-wiki-link](https://www.npmjs.com/package/remark-wiki-link) default is ":")
- Override syntax parsing in Jekyll somehow

### DIFF
--- a/packages/foam-core/src/markdown-provider.ts
+++ b/packages/foam-core/src/markdown-provider.ts
@@ -12,13 +12,15 @@ import { ID } from './types';
 
 let processor: unified.Processor | null = null;
 
+const ALIAS_DIVIDER_CHAR = '|';
+
 function parse(markdown: string): Node {
   processor =
     processor ||
     unified()
       .use(markdownParse, { gfm: true })
       .use(frontmatterPlugin, ['yaml'])
-      .use(wikiLinkPlugin);
+      .use(wikiLinkPlugin, { aliasDivider: ALIAS_DIVIDER_CHAR });
   return processor.parse(markdown);
 }
 
@@ -53,9 +55,20 @@ export function createNoteFromMarkdown(
     }
 
     if (node.type === 'wikiLink') {
+      // links can be either in format [[text]] or [[text|alias]].
+      const text = node.value as string;
+      const alias = node.data?.alias as string;
+      const literalContent = markdown.substring(
+        node.position!.start.offset! + 2,
+        node.position!.end.offset! - 2
+      );
+      const hasAlias =
+        literalContent !== text && literalContent.includes(ALIAS_DIVIDER_CHAR);
       links.push({
         type: 'wikilink',
-        slug: node.value as string,
+        slug: text.trim(),
+        alias: hasAlias ? alias.trim() : null,
+        literalContent,
         position: node.position!,
       });
     }
@@ -127,6 +140,7 @@ export function stringifyMarkdownLinkReferenceDefinition(
 
   return text;
 }
+
 export function createMarkdownReferences(
   graph: NoteGraph,
   noteId: ID,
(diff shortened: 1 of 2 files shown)


packages/foam-core/src/markdown-provider.ts
```diff
@@ -143,45 +157,78 @@ export function createMarkdownReferences(
...
+
+  // Try to generate a definition for each [[link]] in the note.
+  //
+  // A note may have multiple [[link]] expressions to the same target
+  // note with different aliases.s
+  //
+  // - note.links contains all the [[link]] expressions
+  // - forwardLinks contains all the edges (links) between documents
+  return (
+    note.links
```
riccardoferretti, line 176: shall we make these unique by `literalContent`? We should probably add a test just in case

e.g.
```
that was [[a-great-day]] and anytime I have [[a-great-day]] things are good.
```

(if this turns out to be a bug I guess it was already present in the existing code)

packages/foam-core/src/markdown-provider.ts
```diff
@@ -143,45 +157,78 @@ export function createMarkdownReferences(
...
+      .map(linkExpression => {
+        // find the link between this and other document
+        const link = forwardLinks.find(
+          note => note.link.slug === linkExpression.slug
```
riccardoferretti, line 180: I was finding this hard to understand, then I think this parameter is not really `note`, is it? isn't it the graph connection?

This might partially be a naming issue (as connection/link has a `link` attribute, which makes things harder to read), not sure what's the best naming pattern we should adopt (maybe `connection`?)

packages/foam-core/src/markdown-provider.ts
```diff
@@ -143,45 +157,78 @@ export function createMarkdownReferences(
...
+        );
+
+        // if other document is not found, bail
+        if (!link) {
```
riccardoferretti, line 184: looking at this I feel that variable should be called `connection`, as it would probably read  more easily and defines a separation between the link in the note and the connection in the graph.

```
if (!connection) {
```

(makes me wonder whether the incorrect naming is stemming from the `getForwardLinks` function, which we might instead wanna call `getForwardConnections`)

packages/foam-core/src/markdown-provider.ts
jevakallio: Yes, we should add it here, good catch! Right now the unique check is done on the consumer end after stringifying them, but it makes sense to move over here. Will do!

packages/foam-core/src/markdown-provider.ts
jevakallio: Yeah, there definitely is some terminology overloading happening for "link" that we should disambiguate. I like Connection, but also Edge?

In either case, I don't think this is the right PR to make that change.

packages/foam-core/src/markdown-provider.ts
jevakallio: Hmm... yeah, this is confusing to me too now. I'll look into it once I loop back to this PR -- for now I'm not sure if we should move ahead with this until we address the GitHub Pages rendering issues.

packages/foam-core/src/markdown-provider.ts
riccardoferretti: yeah it's possible that `edge` is a better name, I suggested `connection` because that's what we use in `NoteGraph`, but I am easy about it.
I agree 100% we shouldn't make that refactoring as part of this PR

PR #224, opened 2020-08-13
Update math-support.md

sashmit:
Explains how this mechanism works and adds some ideas on alternatives for people using foam for pretty math-heavy notes.


jevakallio:
@sashmit thank you! I restructured the document a little to make it easier to read, and rewrote bits about the alternative syntaxes to make it align better with Foam's direction. 

Thanks for the contribution!

@all-contributor add @sashmit for doc 🎉

PR #227, opened 2020-08-21
Added setting for toggling Link Reference generation.

jessems:
Adds a setting for toggling Link References. Defaults to "true" as per specifications in the roadmap.

Note, this doesn't affect the janitor script. When you run the janitor with the setting set to false, Link References will still be generated.

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -38,6 +38,11 @@
     "configuration": {
       "title": "Foam",
       "properties": {
+        "foam.edit.linkReferenceEnabled": {
+          "type": "boolean",
+          "default": true,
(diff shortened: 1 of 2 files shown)


jevakallio:
@jessems thanks for this suggestion! I'm not able to accept this PR, since removing link reference definitions affects many other places in the codebase (for example, the janitor, the CLI, documentation, etc.) and this PR does only a small fraction of the work.

We may add support for this feature down the line as we have a better understanding on how we're going to manage link targeting. There are some ongoing plans in:

- https://foambubble.github.io/foam/link-reference-definition-improvements
- https://github.com/foambubble/rfcs/pull/3

In the meantime, if you don't need the markdown compatibility links, you can disable the Foam extension and just use the Markdown Links and Markdown Notes extensions as you were.

Issue #229, opened 2020-08-23
First Daily note fails after changing `foam.openDailyNote.directory` 


tristansokol:
Hi, I'm new. Thank you for all your work so far, foam has been interesting to check out! 

I noticed that immediately after setting `foam.openDailyNote.directory` running the **Open Daily Note** command results in a little error pop-up: 
![image](https://user-images.githubusercontent.com/867661/90979921-badbca00-e51d-11ea-8477-4975185e1639.png)
```
Command 'Foam: Open Daily Note' resulted in an error (ENOENT: no such file or directory, open '.../foam/[new directory]/2020-08-23.md')
```

Running the command again works fine as the first invocation created the directory. Looking through the code it seems like the issue might be fixed by adding an `await` here:  
https://github.com/foambubble/foam/blob/d8512db7aa5c8f98e002d0744e54c3312bf50dc1/packages/foam-vscode/src/features/open-daily-note.ts#L66

I am happy to open a pull request if this fix is wanted and if it at least seems like I'm in the right spot to start debugging. 

Thank you!

jevakallio:
Fixed via https://github.com/foambubble/foam/pull/232. Thanks @sksmith! 🙌

PR #230, opened 2020-08-23
add inital template for bug type issues

tristansokol:
While reporting #229 I noticed that there was not any issue templates. I thought I had read about someone wanting one created but was unable to find any supporting documentation to that point. In any case I think they are generally helpful. Feel free to add any feedback and if you would like other use case templates made I'd be happy to oblige.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us be foamier
+labels: bug, awaiting triage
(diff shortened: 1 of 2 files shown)


riccardoferretti (commented):
really like this but will leave to @jevakallio to actually approve the PR, not sure whether he's got some ideas about how to structure the template

jsjoeio:
I agree with @riccardoferretti - thank you for the PR! 🎉 I added @jevakallio as a reviewer who can make the final call.

riccardoferretti (approved):

PR #232, opened 2020-08-24
Daily note could be created before the daily note directory

### LINKED ISSUE #229: First Daily note fails after changing `foam.openDailyNote.directory`

sksmith:
Fixes #229

### DIFF
--- a/packages/foam-vscode/src/features/open-daily-note.ts
+++ b/packages/foam-vscode/src/features/open-daily-note.ts
@@ -14,7 +14,7 @@ import { FoamFeature } from "../types";
 import { docConfig } from '../utils';
 
 const feature: FoamFeature = {
-  activate: async (context: ExtensionContext) => {
+  activate: (context: ExtensionContext) => {
     context.subscriptions.push(
       commands.registerCommand("foam-vscode.open-daily-note", openDailyNote)
     );
@@ -98,7 +98,7 @@ async function focusDailyNote(dailyNotePath: string, isNewNote: boolean) {
 }
 
 async function pathExists(path: string) {
-  return fs.promises
+  return await fs.promises
     .access(path, fs.constants.F_OK)
     .then(() => true)
     .catch(() => false);
--- a/packages/foam-vscode/src/features/open-daily-note.ts
+++ b/packages/foam-vscode/src/features/open-daily-note.ts
@@ -97,8 +97,8 @@ async function focusDailyNote(dailyNotePath: string, isNewNote: boolean) {
   }
 }
 
-async function pathExists(path: string) {
-  return await fs.promises
+function pathExists(path: string) {
+  return fs.promises
     .access(path, fs.constants.F_OK)
     .then(() => true)
     .catch(() => false);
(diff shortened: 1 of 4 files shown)


packages/foam-vscode/src/features/open-daily-note.ts
```diff
@@ -63,7 +63,7 @@ async function createDailyNoteIfNotExists(
     return false;
   }
 
-  createDailyNoteDirectoryIfNotExists(dailyNotePath);
+  await createDailyNoteDirectoryIfNotExists(dailyNotePath);
```
ctrlplusb, line 66: I think there was an eslint plugin/rule to help with these types of oversights. Such a killer to debug these. Well done for picking it up. 🙏

ctrlplusb: https://eslint.org/docs/rules/require-await

sksmith: I went ahead and added the require-await to this pull request. I then cleaned up the warnings for this file specifically. So ya know, require-await wouldn't have found this particular bug because there are already other awaits inside of that function, but it still seems like a handy warning!

packages/foam-vscode/src/features/open-daily-note.ts
```diff
@@ -98,7 +98,7 @@ async function focusDailyNote(dailyNotePath: string, isNewNote: boolean) {
 }
 
 async function pathExists(path: string) {
-  return fs.promises
+  return await fs.promises
```
anku255, line 101: Just curious about the addition of `await` here. Is it required?  https://eslint.org/docs/rules/no-return-await

anku255 (approved):

packages/foam-vscode/src/features/open-daily-note.ts
jevakallio: @anku255 thanks for noticing this, I fixed this in [`9a68fe9` (#232)](https://github.com/foambubble/foam/pull/232/commits/9a68fe9f52537fd86eb91ebb5e2dca3751870482)

jevakallio (approved):
Looks good @sksmith! This change broke link reference definition CodeLens was still using the old API, fixed it in [`52fe704` (#232)](https://github.com/foambubble/foam/pull/232/commits/52fe7041e9212b53758d105000c62dd89accd0ce)

jevakallio:
@all-contributors add @sksmith for code 🎉

sksmith:
Thanks @jevakallio !

PR #236, opened 2020-08-27
Fix link references spacing when there is no trailing newline

### LINKED ISSUE #226: Wiki-links MD compatibility generator spacing from end of content
**Current behavior:** When Markdown-compatible links are generated and added to the bottom of the document, they add only one additional newline. When VSCode is using application defaults (no newline on save), this results in only one newline between the content and the links, which Markdown doesn't interpret as a new paragraph/element, resulting in the links getting "attached" to the content.

**Expected behavior:** The links block should always include an additional newline to create at least one blank line between the end of the content and the links block so that it's always separate from the content.

**Note:** The current behavior is affected by VSCode's newline-on-save setting. VSCode defaults the "add newline to end of files on save" setting to `false`, which results in the above current behavior. However, when this setting is set to `true`, the expected behavior is observed. This feature should not be dependent on this VSCode setting to work properly and should have the expected behavior at all times to avoid getting tangled into the content.

anku255:
fixes #226 

## Summary

There was a typo in the code where the `hasEmptyTrailing` function was not being called and it always returned true. Thus, even when there was no trailing newline, the extension added only one newline instead of two. 

## Test Plan

1. Update VS Code setting `"files.insertFinalNewline": false` 
2. Open a note. Which contains unmaterialized link references. 
3. Delete any trailing newline.
4. Save the file to generate link references.
5. It should add two newlines between the content and the link references.

### DIFF
--- a/packages/foam-vscode/src/features/wikilink-reference-generation.ts
+++ b/packages/foam-vscode/src/features/wikilink-reference-generation.ts
@@ -87,7 +87,7 @@ async function createReferenceList(foam: NoteGraph) {
   if (refs && refs.length) {
     await editor.edit(function(editBuilder) {
       if (editor) {
-        const spacing = hasEmptyTrailing
+        const spacing = hasEmptyTrailing(editor.document)
           ? docConfig.eol
           : docConfig.eol + docConfig.eol;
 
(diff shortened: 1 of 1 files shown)


anku255:
@jevakallio If you end up merging this PR. Can you also add me as a "code contributor"? I am only listed in docs and tests 😛

riccardoferretti (approved):
LGTM

jevakallio (approved):
Ah @anku255 good catch! What a classic JavaScript bug! 💩🧇

jevakallio:
@all-contributors add @anku255 for code 😉

PR #238, opened 2020-08-30
added recipe for local testing

IvoLeist:
Hello Foamers,

first thank you for creating that awesome note taking tool set this helps a lot in organising my PhD Notes 
and for the friendly welcome on your discord !

Since I am using it already for a quite a while and it is "soft" lockdown in Barcelona anyway...
I wanted to give something back to the community.

So as promised to Ingalles on discord here is my PR for adding a recipe for local testing.

Please test it and give me feedback if it works for you as well.
*Feel free to correct typos, wording and such a like ;D

Best,
Ivo


docs/local-deploy.md
```diff
@@ -0,0 +1,58 @@
...
+    
+    To install Bundler:
+
+    - Install Ruby (https://www.ruby-lang.org/en/documentation/installation/)
+    - Install Bundler: `gem install bundler`
+
+# Instructions
+- Open the command line
+- Navigate into your foam folder
+- Switch to the github pages branch (if you do not have one create it):
```
ingalless, line 21: I think we can skip this step. As it is for tech-savvy users, if they want to create a new branch to do the changes in then that's up to them.

docs/local-deploy.md
```diff
@@ -0,0 +1,58 @@
+
+# How to deploy your foam locally (using jekyll)
+
+Recipe for building your GitHub Pages site locally to preview and test changes to your site.
+
+! **Disclaimer**: This Recipe is more suited for "tech savy" foam users since it requires the usage of the command line.
```
ingalless, line 6: Really being pedantic here so I do apologise!! 😆 Correct spelling is "savvy"

ingalless (changes requested):
Looks really good! I've just made a few comments, which I'm happy to discuss 😄 
Thank you for this contribution!

docs/local-deploy.md
IvoLeist: no worries ;D, thank you for spotting!

docs/local-deploy.md
IvoLeist: Good point! - removed

docs/local-deploy.md
```diff
@@ -0,0 +1,55 @@
...
+- Get a jekyll config file [here](https://jekyllrb.com/docs/configuration/default/)
+- In the jekyll config set `baseurl` to your foam folder name
+- FINALLY build and serve your pages: `bundle exec jekyll serve`
+- Enjoy your local foam deploy on: http://localhost:4000/your-folder-name/ (do not forget the trailing '/')
+
+# TL;DR
+```
+git clone https://github.com/IvoLeist/foam-gatsby-template.git
+cd foam-gatsby-template
+git checkout jekyll
```
ingalless, line 38: Just need this removed too!

ingalless (changes requested):
Almost there! Just one little bit to remove and we will be good. Thank you 🎉

riccardoferretti (commented):
I really like having a recipe to easily allow people to test locally their knowledge base.

I have a couple of questions/comments:
- I like the approach of having a `jekyll` (or whatever name) branch that would quickly set up the repo to support the local environment. A long term approach IMO would be to support this in the CLI but it's another order of magnitude in terms of effort, and this workaround is clean and works
- I feel the branch should always be ahead of master (content shouldn't change) and only differ in the setup files. Actually I even wonder if it would be better achieved as a patch/commit to cherry pick to make this clearer - curious to hear thoughts
- I am not sure about not using the official foam template (https://github.com/foambubble/foam-template/), is there a reason to not do so?

docs/local-deploy.md
IvoLeist: Done

IvoLeist:
@riccardoferretti thank you for your input :) 
I was not sure if official foam template is up to date I thought foam-gatsby is the suggested one ? 
@ingalless can you elaborate on that?

> I feel the branch should always be ahead of master (content shouldn't change) and only differ in the setup files. Actually I even wonder if it would be better achieved as a patch/commit to cherry pick to make this clearer - curious to hear thoughts.

I would suggest the branch version like I did in my fork because addressing by commits only might be hassle if you want to upgrade a old foam template or how would you do that if every feature is in a commit?

Another idea (my preferred one) would be to have every foam template element (graph, search etc) in one big repo and 
you can cherry pick in a build file. But I am not sure if that is feasible (how much work that would be?)

I have not looked into the cli repo yet. Why do you think this is so difficult to set up for the cli?

ingalless:
@IvoLeist To my knowledge the official one is up to date and the gatsby one is **not** the official one at this moment in time.
@riccardoferretti has been around this project for longer than me, so his review holds more weight than my own! Although, to the point of a branch or cherry picking commits, I agree that it seems quite difficult.

@jevakallio made a comment [here](https://github.com/foambubble/foam-template/issues/3#issuecomment-650827065) where he suggests a recipe be created. He doesn't (didn't?) want to maintain deployment-specific code in the foam repository.

@riccardoferretti was not suggesting the work on the CLI would be so difficult, only that it is more effort than authoring a markdown file. I'm sure it is perfectly doable, but after reading Jani's comments I feel that it may not be the preferred solution.

For me the recipes allow those wanting more from foam to find guides that can help them achieve that. Interested to hear both your thoughts on this 😄

riccardoferretti:
@IvoLeist @ingalless some thoughts

- I really like that by having the setup on a branch, people that are not interested won't see the extra files/complexity, while people that are interested will easily be able to switch to the branch and get access to a local jekyll environment 
- the problem with the branch is that if I do 5 commits, then want to enable the local environment and switch to the branch, the content I will see won't include my commits, unless I rebase the branch
-- thinking about it, this can easily be "fixed" by adding the rebase step in the recipe

The CLI approach will probably be the preferred approach in the future, but that part is in "draft stage" (as it is part of a wider publishing effort). 

The recipe looks really good to me, my comment about the repo was referring to the fact that this approach wouldn't work on the official template. Maybe the solution there is to add such branch on the template as well, so the recipe can say "this works on the official foam repo (url) and the semi-official gatsby foam repo (url)". Possibly a nitpick, keen to hear @jevakallio 's take on that.

IvoLeist:
@ingalless ah ok good to know but apparently there is also the plan to move to gatsby at one point:
https://foambubble.github.io/foam/improved-static-site-generation

> He doesn't (didn't?) want to maintain deployment-specific code in the foam repository
Does this hold true for all repos inside foambubble or only for this repo.

I guess it would be just a convenience thing to have it in the cli otherwise a bash script like I have it TL;DR should be sufficient

@riccardoferretti adding rebase to the recipe is indeed a good idea!
Where would you put it?

After:
`Navigate into your foam folder`

Insert:
```
Switch to the jekyll branch: switch: `git checkout jekyll`*
git fetch upstream
git rebase upstream/master
```
?

*So before merging that PR we first have to create a new PR for foambubble/foam-template.
But for this someone with write privileges has to create that Branch first, since you cannot create a new remote branch via a PR,right?

riccardoferretti:
@IvoLeist for me the goal of the rebase is that your (local) commits on master are available on the `jekyll` branch, so I think we don't need to involve upstream in the flow.

```
Switch to the jekyll branch : git checkout jekyll
Insert the newest commits in master in front of your jekyll branch: git rebase master
```

If someone wants to always have access to the local environment, an alternative would be to just merge the `jekyll` branch into master (might be worth adding a note to the recipe about this).

wdyt?

jevakallio (changes requested):
@IvoLeist, thanks for this recipe! I made a few improvements to the language and pushed it up to the branch. However, when testing this locally with my own Foam I ran into a couple of issues:

- [ ] Jekyll was not able to read the default `styles.scss` file due to the frontmatter `---` section at the top of the file. Are we missing some Jekyll plugin that's required?
- [ ] After fixing above issue manually, the bundler server started, but instead of serving a rendered HTML site, it serves a directory listing of the repository. Additionally, the generated `_site` directory simply contains `.md` files, not generated HTML files. Are we missing some configuration of build step?

To address @riccardoferretti's and @ingalless's points, it would be great for Foam template to be locally runnable by default, but we do want to keep the template as low-code as possible.

In the meantime I'd be happy to accept this recipe, assuming the instructions are easy to follow and work out of the box.

IvoLeist:
Hello @jevakallio 

finally found some time to look into foam again:

Cool that you like it and thank you for improving the language.

The errors you describe are arising for me as well when I am using https://github.com/foambubble/foam-template
that is one of the reasons why I used the [foam-gatsby-template](https://github.com/mathieudutour/foam-gatsby-template).
Despite the missing _config.yml that template worked right out of the box for me.

Thanks for the tip with the plugins I am going to try if installing one of these solves the issues:

> it would be great for Foam template to be locally runnable by default, but we do want to keep the template as low-code as possible

So thats why you have not added that 
```
Switch to the jekyll branch : git checkout jekyll
Insert the newest commits in master in front of your jekyll branch: git rebase master
```
suggestion from @riccardoferretti ?

riccardoferretti:
@IvoLeist sorry for the belated response, I thought I already had

> The errors you describe are arising for me as well when I am using https://github.com/foambubble/foam-template
> that is one of the reasons why I used the [foam-gatsby-template](https://github.com/mathieudutour/foam-gatsby-template).
> Despite the missing _config.yml that template worked right out of the box for me.

I am guessing when you say _that_ you are referring to `foam-gatsby-template`.
In this case I would _at least_ add that information in the recipe (I just re-read it and I couldn't find a reference to the template used), and even better ( ;) ) address both cases - after all `foam-template` is the default template for foam, and we should include support for it in our recipes (where reasonably possible).
 
> Thanks for the tip with the plugins I am going to try if installing one of these solves the issues:

Did you have any luck with this?

> > it would be great for Foam template to be locally runnable by default, but we do want to keep the template as low-code as possible
> 
> So thats why you have not added that

yes - the medium term solution will be having a `.foam` folder where the "coding" bits will live, so to no pollute the note taking space of the user, while still supporting various use cases. That's still a WIP and taking shape as we talk

> 
> ```
> Switch to the jekyll branch : git checkout jekyll
> Insert the newest commits in master in front of your jekyll branch: git rebase master
> ```
> 
> suggestion from @riccardoferretti ?

Not sure I understand this, is it a question for me?

Thanks for looking into this!

riccardoferretti:
closed because of inactivity

PR #261, opened 2020-09-12
added support for graph middleware via local plugin

riccardoferretti:
This is a proposal for supporting plugins that live within the local foam repo.
This allows users to hook into foam and interact with it.

If this direction shows potential, it can be used to inject custom code also in the CLI (e.g. add new commands), and the vscode extension.

Edit:
Here are some example on how such a system could be used:
- currently we support YAML properties, with this system I could integrate roam style attributes (e.g. `stage:: seedling`) directly in my repo, even if not available officially in Foam
- I could automatically add tags to my notes based on the location in the repo (e.g. notes in `/area/finance` will automatically get the `#finance` tag)
- I could add a new CLI command to support some internal use case (e.g. what @scott-joe has done for his workspace)
- I could add a new CLI command to import/export to an unsupported source/target 
- Extend the VSCode experience to support one's own workflow, e.g. weekly note, templates, extra panels, foam model derived TOC, ... all without having to write/deploy a VSCode extension

That this API is currently unstable and subject to changes, this needs to be properly communicated, suggestions appreciated :)
There is a caveat - at the moment this makes little difference because we don't really use the Foam metadata anywhere, but as we start leveraging the model for things like backlink generation, TOC, graph exploration, LSP integration, ... augmenting the model will provide a lot of flexibility into our single source of truth


packages/foam-core/src/markdown-provider.ts
```diff
@@ -6,88 +6,106 @@ import { parse as parseYAML } from 'yaml';
...
-  title = frontmatter.title ?? title;
+  },
+};
+
+export function createMarkdownParser(extraPlugins: ParserPlugin[]): NoteParser {
+  const parser = unified()
+    .use(markdownParse, { gfm: true })
+    .use(frontmatterPlugin, ['yaml'])
+    .use(wikiLinkPlugin);
+  // plugin.onDidInitializeParser(parser)
```
jsjoeio, line 70: nit: Remove?

packages/foam-core/src/markdown-provider.ts
```diff
@@ -6,88 +6,106 @@ import { parse as parseYAML } from 'yaml';
...
+  const plugins = [
+    yamlPlugin,
+    titlePlugin,
+    wikilinkPlugin,
+    definitionsPlugin,
+    ...extraPlugins,
+  ];
+  return {
+    parse: (uri: string, markdown: string, eol: string): Note => {
+      // markdown = plugin.onWillParseMarkdown(markdown)
```
jsjoeio, line 81: nit: remove?

packages/foam-core/src/markdown-provider.ts
```diff
@@ -6,88 +6,106 @@ import { parse as parseYAML } from 'yaml';
...
+          text: markdown,
+          contentStart: tree.position!.start,
+          end: tree.position!.end,
+          eol: eol,
+        },
+      };
 
-  const end = tree.position!.end;
-  const definitions = getFoamDefinitions(linkDefinitions, end);
+      // plugin.onWillVisitTree(tree, note)
```
jsjoeio, line 99: nit: remove?

packages/foam-core/src/markdown-provider.ts
```diff
@@ -6,88 +6,106 @@ import { parse as parseYAML } from 'yaml';
...
+      visit(tree, node => {
+        for (let i = 0, len = plugins.length; i < len; i++) {
+          plugins[i].visit?.(node, note);
+        }
+      });
+      // plugin.onDidVisitTree(tree, note) (the next line would go here)
```
jsjoeio, line 105: nit: remove?

packages/foam-core/src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  url: string;
+  title?: string;
+  position?: Position;
+}
+
+export interface Note {
+  title: string | null;
+  slug: string; // note: this slug is not necessarily unique
+  properties: any;
+  // sections: NoteSection[]
```
jsjoeio, line 37: nit: remove?

packages/foam-core/src/utils.ts
```diff
@@ -2,7 +2,11 @@ import path from 'path';
 import crypto from 'crypto';
 import { titleCase } from 'title-case';
 import GithubSlugger from 'github-slugger';
-import { URI, ID } from 'types';
+import { URI, ID } from './types';
+
+export function isNotNull<T>(p: T | null): p is T {
+  return p != null;
```
jsjoeio, line 8: What does p stand for? 

nit: consider adding a comment or renaming to be more clear?

jsjoeio (commented):
Generally speaking, I feel like all looks good 🎉

I left a few nit comments - mostly to remove commented out code or add comments to explain uncertainties.

jsjoeio:
I feel like I don't understand what the purpose is or the context enough to approve. Sorry if I sound dense - I've been at hovering at a higher level in terms of understanding Foam's technical aspects.

> This is a proposal for supporting plugins that live within the local foam repo.
> This allows users to hook into foam and interact with it.

Could you elaborate or give an example of how this might benefit a user? I like the sound of plugins and allowing users to hook in but I would love to hear more! 😄

packages/foam-core/src/markdown-provider.ts
riccardoferretti: yeah, probably. The reason why it's there is because before implementing everything I wanted to highlight the possible extension points. So I am actually leaning more towards implementing that rather than removing it.
But I agree that this half baked code doesn't help (also, I realize I haven't documented this enough, will do in a bit)
Thoughts?

packages/foam-core/src/utils.ts
riccardoferretti: good point, should just be `value` - no idea what I was thinking about when I went for `p` (I am sure it was very clever at the time ;) )

riccardoferretti:
thx @jsjoeio I updated the description to provide some examples and replied to your comments, let me know what you think!

jsjoeio:
That makes it a lot more clear! 
> I updated the description to provide some examples and replied to your comments, let me know what you think!

jsjoeio (approved):
LGTM!

jsjoeio:
You probably want another review or so though before we merge since this is a big change, right?

ingalless (changes requested):
Looks good! Just a few comments

docs/foam-local-plugins.md
```diff
@@ -0,0 +1,25 @@
...
+- add a new CLI command to support some internal use case or automate import/export
+- extend the VSCode experience to support one's own workflow, e.g. weekly note, templates, extra panels, foam model derived TOC, ... all without having to write/deploy a VSCode extension
+
+## Technical approach
+
+When Foam is loaded it will:
+- check `.foam/plugins` directory.
+	- each directory in there is considered a plugin
+	- the layout of each directory is
+		- `index.js` contains the main info about the plugin, specifically it exports:
```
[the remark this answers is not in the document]
riccardoferretti: unfortunately this is the case (atm). 
We _could_ depend on typescript and compile them on the fly, but that's a heavy dependency, I didn't think it is worth it at this point (if someone wants to take it on, that'd be great!). 
Not sure if there is another approach that would work to avoid the dependency

jojanaho:
That's pretty clever design! It certainly would give users more flexibility to take Foam into directions that they would like. I tried to think this through on a high level and not concentrate on the code. The following things come to mind:

### Security

The first thing that comes to mind is what happens if user opens a folder with a malicious .foam-directory? Currently none of the extension that I use execute code automatically from the folders that I open, so this might come as a surprise.

### API stability

How stable is the API that we would define for the plugins? It would suck to get foam-extension auto-updated and see that some of the functionality that I developed stops working.

### Workspace specifity

What if I'd like to use some functionality across all my foam workspaces - should I somehow replicate the code to each workspace, or would e.g. a common lookup from my home folder be supported?

How about if I'd like to publish my plugin functionality as own extension into vscode marketplace? This mechanism wouldn't directly support getting it hooked into the same Foam graph / parser (?)

### Supported languages

What about if I'm a python developer, and would like to extend Foam with python? Current mechanism supports .js only, but with LSP and IPC interaction I potentially could develop with any language. This could be something similar as done with Jupyter (IPython) kernels. However, it certainly would be more complex than this, so a trade-off to consider...

In short I'm neutral towards this change, since I'm lacking the big picture where Foam is going in general.

riccardoferretti:
Hey @jojanaho 
really good thoughts there, let me address them below:

## Comments
### Security
This is a veeeeery good point. Although the idea is that this is your repo and in theory you only have access to it (or people you trust), but given it can execute _any_ code on your system I agree we need to at least **put some checks in place**.

### API stability
The API is not meant to be stable at this point, use at your own risk. **This needs to be properly communicated**.

### Workspace specificity
Yes, these are meant to be "local" plugins, if you want to use a plugin in more than one workspace, it's up to you to either link the two, have a repo that is checked out in both repos, and so on.
This PR is not addressing creating VSCode extensions that leverage the Foam API

### Supported languages
Good considerations, and once we have an LSP or stand-alone server more languages would be supported. At the moment this is out of scope, part of the goal here is to see how people use this functionality (will they?), so we want to give enough to make it possible, without capturing all use cases.

## Proposal
With regards to **Security** and **API stability** I was thinking of the following:
- put this feature behind something like a [Proposed API](https://code.visualstudio.com/api/advanced-topics/using-proposed-api) feature flag (e.g. `foam.experimental.localPlugins.enable`). This would provide:
  - additional security, as you can decide to not enable local plugins
  - communicate that the API is not stable
- as a second security check, we could require the repo path to be whitelisted before execution is allowed. In case the repo is not whitelisted, prompt the user. The whitelist includes the full local path to the repo, as a generic setting could just be set in the `.vscode/settings.json` file. e.g. `foam.experimental.localPlugins.whitelist`

Missing anything? Thoughts?

riccardoferretti:
@jsjoeio @ingalless I have pushed some code to account for the conversation with @jojanaho - please let me know what you think and whether the PR is good to go!

prettier.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
ingalless, line 1: Not overly bothered by this, but has this snuck in? 😮

ingalless (approved):

ingalless (approved):
Looks good! I did leave a comment, but regardless of the outcome I like it. Nice spots @jojanaho

prettier.config.js
riccardoferretti: yes, for some reason my prettier started to use different settings, so I realized the repo doesn't have explicit prettier rules, so I added them :)

PR #274, opened 2020-10-03
feat(vscode): add command "Copy To Clipboard Without Brackets 

jsjoeio:
This adds a new command for `foam-vscode`  called "Copy To Clipboard Without Brackets". I discussed this in the Foam Discord a [while back](https://discordapp.com/channels/729975036148056075/730056943997288540/739321725967138847), asking if this should be an outside thing (separate extension) or added as a feature to the core extension. It was suggested to do the later. 

## Use Case

Sometimes when drafting messages to people in Foam, I use references. Instead of sending them the messages with the brackets or removing them, I thought it would be nice to have this command which copies to clipboard and removes them for you.

### DIFF
--- a/packages/foam-vscode/src/utils.ts
+++ b/packages/foam-vscode/src/utils.ts
@@ -103,23 +103,36 @@ export function removeBrackets(s: string): string {
   // loop through words
   const modifiedWords = stringSplitBySpace.map(currentWord => {
     if (currentWord.includes("[[")) {
+
+      // all of these transformations will turn this "[[you-are-awesome]]"
+      // to this "you are awesome"
       let word = currentWord.replace(/(\[\[)/g, "");
       word = word.replace(/(\]\])/g, "");
       word = word.replace(/(.mdx|.md|.markdown)/g, "");
       word = word.replace(/[-]/g, " ");
 
-      // now capitalize every word
-      const modifiedWord = word
-        .split(" ")
-        .map(word => word[0].toUpperCase() + word.substring(1))
-        .join(" ");
+      // then we titlecase the word so "you are awesome"
+      // becomes "You Are Awesome"
+      const titleCasedWord = toTitleCase(word);
 
-      return modifiedWord;
+      return titleCasedWord;
     }
 
     return currentWord;
   });
 
   return modifiedWords.join(" ");
+}
+
+/**
+ * Takes in a string and returns it titlecased
+ *
+ * @example toTitleCase("hello world") -> "Hello World"
+ */
+export function toTitleCase(word: string): string {
+  return word
+        .split(" ")
+        .map(word => word[0].toUpperCase() + word.substring(1))
+        .join(" ");
 }
 
--- a/packages/foam-vscode/src/utils.ts
+++ b/packages/foam-vscode/src/utils.ts
@@ -91,3 +91,35 @@ export function dropExtension(path: string): string {
 export const astPositionToVsCodePosition = (point: Point): Position => {
   return new Position(point.line - 1, point.column - 1);
 };
+
+/**
+ * Used for the "Copy to Clipboard Without Brackets" command
+ *
+ */
+export function removeBrackets(s: string): string {
+  // take in the string, split on space
+  const stringSplitBySpace = s.split(" ");
+
+  // loop through words
+  const modifiedWords = stringSplitBySpace.map(currentWord => {
+    if (currentWord.includes("[[")) {
+      let word = currentWord.replace(/(\[\[)/g, "");
+      word = word.replace(/(\]\])/g, "");
+      word = word.replace(/(.mdx|.md|.markdown)/g, "");
+      word = word.replace(/[-]/g, " ");
+
+      // now capitalize every word
+      const modifiedWord = word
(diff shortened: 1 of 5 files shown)


packages/foam-vscode/src/utils.ts
```diff
@@ -91,3 +91,35 @@ export function dropExtension(path: string): string {
...
+ * Used for the "Copy to Clipboard Without Brackets" command
+ *
+ */
+export function removeBrackets(s: string): string {
+  // take in the string, split on space
+  const stringSplitBySpace = s.split(" ");
+
+  // loop through words
+  const modifiedWords = stringSplitBySpace.map(currentWord => {
+    if (currentWord.includes("[[")) {
```
riccardoferretti, line 105: Are we happy with not supporting spaces in the file name/path?
(Even not taking into consideration aliases which we don't support yet)

packages/foam-vscode/src/utils.ts
```diff
@@ -91,3 +91,35 @@ export function dropExtension(path: string): string {
...
+      let word = currentWord.replace(/(\[\[)/g, "");
+      word = word.replace(/(\]\])/g, "");
+      word = word.replace(/(.mdx|.md|.markdown)/g, "");
+      word = word.replace(/[-]/g, " ");
+
+      // now capitalize every word
+      const modifiedWord = word
```
riccardoferretti, line 112: Basically are we "deslugifying" here?

packages/foam-vscode/src/utils.ts
```diff
@@ -91,3 +91,35 @@ export function dropExtension(path: string): string {
...
+export function removeBrackets(s: string): string {
+  // take in the string, split on space
+  const stringSplitBySpace = s.split(" ");
+
+  // loop through words
+  const modifiedWords = stringSplitBySpace.map(currentWord => {
+    if (currentWord.includes("[[")) {
+      let word = currentWord.replace(/(\[\[)/g, "");
+      word = word.replace(/(\]\])/g, "");
+      word = word.replace(/(.mdx|.md|.markdown)/g, "");
```
riccardoferretti, line 108: Do we have this list somewhere already? Should we consolidate it or do you think it's better to keep things more decoupled at the moment?

riccardoferretti (approved):
This looks really good to me. 
I have only left a couple of comments around some edge cases or minor points.

packages/foam-vscode/test/utils.test.ts
```diff
@@ -1,10 +1,49 @@
...
   test("returns file name without extension", () => {
     expect(dropExtension('file.md')).toEqual('file');
   });
 });
+
+describe("removeBrackets", () => {
+  it("removes the brackets", () => {
+    const input = "hello world [[this-is-it]]";
+    const actual = removeBrackets(input);
+    const expected = "hello world This Is It";
```
ingalless, line 16: Perhaps it isn't for this PR, but I sometimes name files differently to my heading, so I would want to grab the title. Is that an easy change? No worries if not but just floating it...
Like the tests though!

ingalless (approved):
This looks great! Very useful command

packages/foam-vscode/test/utils.test.ts
riccardoferretti: That's actually a very good point, especially if we already have the definitions. And will probably leverage more the foam metadata as opposed to simply the markdown itself

packages/foam-vscode/src/utils.ts
jsjoeio: Good call! I didn't even think to check. I'll look and if it is, I'll use that

packages/foam-vscode/src/utils.ts
jsjoeio: Yes! Exactly! I can update the comments to make that more obvious. Happy to even break it into a small function called `deslugify`

packages/foam-vscode/test/utils.test.ts
jsjoeio: Good point! I didn't even think about. 

>  Is that an easy change?

I don't know 😅 More than happy to accept change if one of y'all want to do it!

packages/foam-vscode/src/utils.ts
jsjoeio: Okay I found it inside `docs/_layouts/foam.html`, but can't export it because it's just in a script tag in an HTML file 🤔

I'm not seeing it elsewhere...I'm going to leave for now, but if you see a better option, let me know and I'll happily change it!

packages/foam-vscode/src/utils.ts
jsjoeio: Actually...it title cases the word, so I broke it out into a function, added more comments and a few tests. Good callout!

jsjoeio (commented):
Thanks for the specific feedback @ingalless and @riccardoferretti ! 🙌🏼

I responded and made a few changes. Feel free to take another look! If there's changes to be made, hit the "request changes". If it's ready to go by chance, feel free to merge!

jsjoeio:
Chatted over Discord - good to merge!

PR #282, opened 2020-10-07
Added option to disable wikilink definitions generation

riccardoferretti:
This PR adds a new option to the wikilink definitions generation that disables the feature.

The default behavior is still the same and the change is backwards compatible.

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -48,11 +48,13 @@
           "default": "withoutExtensions",
           "enum": [
             "withExtensions",
-            "withoutExtensions"
+            "withoutExtensions",
+            "off"
           ],
           "enumDescriptions": [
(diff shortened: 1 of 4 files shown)


ingalless:
Code looks good! Can you explain the purpose behind this feature?

riccardoferretti:
Wikilink definition is necessary for a specific use case (e.g. support within github). 
For people that use their notes mostly locally, and use tools link Markdown Links (or Obsidian) to do the navigation locally, generating links can be pretty noisy and disruptive.
Especially as we develop more the core editing feature and "separate" the editing from the publishing (for those who want such separation), it's good to keep some features on an opt-in basis (well, in this case opt-out for backwards compatibility).

ingalless (approved):
Nice work, thanks for the clear explanation too!

Issue #275, opened 2020-10-04
Automatic git sync


chrisands:
Hi, just today started to use foam. Since there is no discussion about automatic git sync...

I think it is possible to use vscode "editor.codeActionsOnSave" to git commit and git push (& pull?) every time file is saved. I'm not familiar with this option, but, briefly looking at it, I think foam needs to be markdown provider to able use it.

What you think about it? I could do this, but I don't think I'll have time in next two or three weeks.

scott-joe:
I have this in my personal list of desired features. I had some ideas on how to cut down on the clutter as well. Perhaps some sort of automated time based gitflow? I'm interested in finding a balance between the granularity of backups and overloading the commit log. So, some sort of local branching each day with a squash commit to the week/month/etc and only daily squash commits to remote? The usefulness the idea is down to the timing of these events in relation to the needs of the user. If you write a lot often, you may want something better than that? But...perhaps at that point, you'd be better off with something other than Git for providing change logs to your files...

riccardoferretti:
What's the "problem" you are trying to solve with this approach @scott-joe ?
Asking because I have been thinking about it myself, and I wonder whether a EOD commit is best (which would give me timeframes), or editing session commit (which would give some sort of clustering of ideas/notes), or even on save (per @chrisands 's comment) just to ensure things are backed up..

SanketDG:
I think given the large spectrum of audience that Foam could be useful too, among all other possible options, manually comitting should be _default_, because otherwise, you would be defining for the user.

One thing I'd like to see it commit message summarization based on the changes.

scott-joe:
@riccardoferretti I'd like to emulate some of the features I rely on in Bear/Notion, such as syncing. To that end, I'd like my workspace to automatically commit or at least stage changes at some regular interval and push those changes to my remote repos.

To keep those repos tidy, I'd like to have some sort of housekeeping done there, but that's just a bonus at this point.

The simplest in-workflow solution would be to wire up a Command Palette entry to something like `git commit -am "{timestamp}" && git push origin main` so I could easily push my local changes to a remote repo before closing my work machine for the day and do the opposite when opening up Foam on my personal machine later on.

chrisands:
So there should be two separate things. First is when and how foam commits changes. And the second thing is how often we sync remote repo with local.

[...]

PR #287, opened 2020-10-12
[WIP] auto sync with repo

chrisands:
Try to implement #275

I want feedback about PR.

I think PR still needs to complete:
- [ ] config keys
- [ ] extract auto sync function
- [ ] optimize auto sync not to pulling every time
- [ ] better function names?

### DIFF
--- a/packages/foam-vscode/src/features/commit-on-save.ts
+++ b/packages/foam-vscode/src/features/commit-on-save.ts
@@ -1,0 +1,85 @@
+import { commands, ExtensionContext, TextDocument, workspace, extensions } from "vscode";
+import { GitExtension, Commit } from "../git";
+import { FoamFeature } from "../types";
+import { exec } from 'child_process';
+import { promisify } from 'util';
+
+const execP = promisify(exec);
+let latestCommitTime;
+
+const feature: FoamFeature = {
+  activate: (context: ExtensionContext) => {
+    context.subscriptions.push(
+      commands.registerCommand("foam-vscode.run-hook-on-save", commitOnSave)
+    );
+    workspace.onDidSaveTextDocument((document: TextDocument) => {
+      commands.executeCommand("foam-vscode.run-hook-on-save", document);
+    });
+  }
+};
+
+async function commitOnSave (document: TextDocument) {
+  let amend = false;
+
+  const gitExtension = extensions.getExtension<GitExtension>('vscode.git').exports;
+  const git = gitExtension.getAPI(1);
+  // get current repo instance
+  const repo = git.repositories[0];
+  // stage current document
+  await execP(`git add ${document.fileName}`, { cwd: repo.rootUri.path });
+  // get current status (should be one file)
+  const statusCmd = await execP(`git status -s`, { cwd: repo.rootUri.path });
+  const status = statusCmd.stdout.trim().split('\n');
+  // get array of filenames of staged files
+  const filesInStage = status.reduce((acc, cur) => {
+    // first argument is file status more details: https://git-scm.com/docs/git-status#_short_format
+    const [, filename] = cur.split('  ');
+    if (filename) {
+      acc.push(filename);
+    }
+
+    return acc;
+  }, []);
+  // get latest commit edited file names
+  const showCmd = await execP(`git show --numstat --oneline`, { cwd: repo.rootUri.path });
+  const show = showCmd.stdout.trim().split('\n');
+  const latestCommitHash = show[0].split(' ')[0];
+  // get array of filenames of latest commit
+  const filesInLatestCommit = show.reduce((acc, cur) => {
+    const [,, filename] = cur.split('	');
+    if (filename) {
+      acc.push(filename);
+    } return acc;
+  }, []);
+
+  // before checking for ammend commit we should check if latest commit if already pushed
+  const branchCmd = await execP(`git branch -r --contains ${latestCommitHash}`, { cwd: repo.rootUri.path });
+  const branch = branchCmd.stdout.trim().split('\n');
+
+  // todo: also add check for new commit interval using latestCommitTime
+  if (
+    branch.length === 0 // latest commit isn't pushed
+    && filesInStage.length === 1 // only one file in stage
+    && filesInLatestCommit.length === 1 // only one file in latest commit
+    && filesInStage[0] === filesInLatestCommit[0] // files are same
+  ) {
+    amend = true;
+  }
+
+
+  await repo.commit(`${(new Date).toLocaleTimeString()}`, { amend });
+  // write commit time
+  latestCommitTime = new Date();
+}
+
+/**
+ * @param d1
+ * @param d2
+ * @param diffence in seconds
+ */
+function isDateIsLargerBy(d1: Date, d2: Date, diffence: number): boolean {
+  const getSeconds = (date: Date): number => +((date.getTime() / 1000).toFixed(0));
+  return (getSeconds(d1) - getSeconds(d2)) > diffence;
+}
+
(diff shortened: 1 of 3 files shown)


SanketDG (commented):
This looks good!

I am just not sure about two things:

- `commit` on every save may be controversial, considering this may be based on one's habit how frequently they save (I save pretty often)

- Why commit by `timestamp`, when the timestamp is there in the git log anyway.

chrisands:
Thank you.

I think `commit` on every save is the best option to keep repository updated. It use `commit --amend`, so git log shouldn't be bloated. But I'm open for suggestions to implement it. Maybe timer which will call commit every N seconds and if document is closed?

Commit message `timestamp` was easiest to implement. I can replace it with commit message like "Add project todo". 
First word is file status from status
> **from [git docs](https://git-scm.com/docs/git-status#_short_format)**
> M = Update
> A = Add
> D = Remove
> R = Rename
> C = Copy
> U = Update

Other words is parsed file name from kebab-case.

SanketDG:
> Maybe timer which will call commit every N seconds and if document is closed

I guess this looks good, I would also be interested in just a command like `Foam: Sync` (for example)

> It use `commit --amend`, so git log shouldn't be bloated.

So there would be one commit always?

> Commit message timestamp was easiest to implement. I can replace it with commit message like "Add project todo".
First word is file status from status

I never knew this is so comprehensive from git, this would be really pleasant.

packages/foam-vscode/src/features/commit-on-save.ts
```diff
@@ -0,0 +1,100 @@
...
+const execP = promisify(exec);
+let latestCommitTime;
+let latestSyncTime;
+let commitInterval = 10; // replace with configuration
+let syncInterval = 60 * 5;
+
+const feature: FoamFeature = {
+  activate: (context: ExtensionContext) => {
+    context.subscriptions.push(
+      commands.registerCommand("foam-vscode.run-hook-on-save", commitOnSave)
```
riccardoferretti, line 16: what does this command do?

packages/foam-vscode/src/features/commit-on-save.ts
```diff
@@ -0,0 +1,100 @@
...
+    );
+    workspace.onDidSaveTextDocument((document: TextDocument) => {
```
riccardoferretti, line 18: I saw that in your list of todos, I imagine this would live behind a feature flag to enable/disable it?

packages/foam-vscode/src/features/commit-on-save.ts
```diff
@@ -0,0 +1,100 @@
...
+      commands.executeCommand("foam-vscode.run-hook-on-save", document);
+    });
+  }
+};
+
+async function commitOnSave (document: TextDocument) {
+  let amend = false;
+
+  const gitExtension = extensions.getExtension<GitExtension>('vscode.git').exports;
```
riccardoferretti, line 27: is this the git built-in extension in vscode? what I mean is, is it possible it would not be defined?

packages/foam-vscode/src/features/commit-on-save.ts
```diff
@@ -0,0 +1,100 @@
...
+  const git = gitExtension.getAPI(1);
+  // get current repo instance
+  const repo = git.repositories[0];
```
riccardoferretti, line 30: a few questions about some edge cases:
- if we are not in a git repo, I imagine this would fail?
- when would this array have more than one element? 
- what happens if someone has multiple git repos as subdirectories (something similar to what @scott-joe suggested somewhere)? Scott, can you share here your use case for reference?
- as we currently support workspaces, it's possible/likely they will live in separate directories/repos

Sorry about all the nitpicking, my fear is that if we add this feature we could end up with a bunch of edge cases, and we need to be clear about what we support/bless and what we don't (and of course we shouldn't break in any case :) )

packages/foam-vscode/src/features/commit-on-save.ts
```diff
@@ -0,0 +1,100 @@
...
+  const branchCmd = await execP(`git branch -r --contains ${latestCommitHash}`, { cwd: repo.rootUri.path });
+  const branch = branchCmd.stdout.trim().split('\n');
+
+  if (
+    branch.length === 0 // latest commit isn't pushed
+    && filesInStage.length === 1 // only one file in stage
+    && filesInLatestCommit.length === 1 // only one file in latest commit
+    && filesInStage[0] === filesInLatestCommit[0] // files are same
+    && isDatesInSameInterval(new Date(), latestCommitTime, commitInterval) // check if latest commit included in time interval
+  ) {
```
chrisands, line 68: > So there would be one commit always?

@SanketDG it uses amend if conditions and met:
- latest commit isn't pushed. This way don't have to force push changes.
- only one file in latest commit. By this we ensure that commit message won't be indefinite.
- only one file in stage. For the same reason as previous condition.
- files in stage and latest commit are the same files.
- latest commit is in time interval. This way we we will have history of editing file.

packages/foam-vscode/src/features/commit-on-save.ts
chrisands: Yes

packages/foam-vscode/src/features/commit-on-save.ts
chrisands: It registers a command so we can use it later.

packages/foam-vscode/src/features/commit-on-save.ts
chrisands: Yes, it is built-in, but it possible that user disabled it manually. We should throw an error if it is undefined.

packages/foam-vscode/src/features/commit-on-save.ts
chrisands: > if we are not in a git repo, I imagine this would fail?

Probably we should throw an error and suggest to create repository.

> as we currently support workspaces, it's possible/likely they will live in separate directories/repos

Would be great to have a reference, to test it.

> when would this array have more than one element?

It will pick first repo, that is detected. [Here's source code](https://github.com/microsoft/vscode/blob/2b7bf327b9c59ef56ce7608acab637800714a039/extensions/git/src/api/api1.ts#L248).

packages/foam-vscode/src/features/commit-on-save.ts
ingalless: > latest commit isn't pushed. This way don't have to force push changes.

Will the foam workspace setting interfere with this? From what I understand it automatically runs `git sync` following a commit (when using the version control panel)

jasonwilliams:
I sometimes work behind a proxy, will it be possible to commit automatically but switch off push?

chrisands:
> I sometimes work behind a proxy, will it be possible to commit automatically but switch off push?

Yes, it will be possible turn off everything.

I will update PR only on next weekend.

packages/foam-vscode/src/features/commit-on-save.ts
SanketDG: IIRC you need to have

```
  "git.postCommitCommand": "sync",
```

riccardoferretti:
@chrisands @SanketDG @ingalless @jasonwilliams 
[GitDoc](https://marketplace.visualstudio.com/items?itemName=vsls-contrib.gitdoc) looks like an extension that does exactly this, with a few extra features - is there anything there that you feel is missing in Foam?

ingalless:
@riccardoferretti good spot! Looks like exactly what most are looking for

riccardoferretti:
closing this for lack of activity and having found a viable alternative

PR #290, opened 2020-10-15
Graph based on foam metadata

riccardoferretti:
This PR shows a graph that is driven by the foam metadata.

This is currently a spike, and there are 3 graph appraoches being tested:

- [D3](https://d3js.org/)
- [G6](https://g6.antv.vision/)
- [Force Graph (D3 on WebGL)](https://github.com/vasturiano/force-graph) 

Currently it's pretty easy to switch between the three, just uncomment the corresponding line in `graph.js`

Here are  my thoughts so far:
- G6 has some really cool graphing capabilities, but it is the slowest of the three, has a small community/knowledge base (mostly in Chinese) -> so will probably pass
- D3 is super flexible, has a big community and most problems have already been encountered/solved. Unfortunately it gets pretty slow fairly quickly
- Force Graph is a point library, which does really well what we need here, and nothing else. It's very fast, and the configuration options match our needs so far

foam-graph


packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -0,0 +1,111 @@
...
+        graph.nodes.push({
+          id: n.id,
+          type: "note",
+          uri: n.source.uri,
+          title: path.basename(n.source.uri),
+          nOutLinks: foam.notes.getForwardLinks(n.id).length,
+          nInLinks: foam.notes.getBacklinks(n.id).length
+        });
+        foam.notes.getForwardLinks(n.id).forEach(link => {
+          if (foam.notes.getNote(link.to) == null) {
```
ingalless, line 36: Is there a reason for the == over ===?

packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -0,0 +1,111 @@
...
+        nodes: [],
+        edges: []
+      };
+
+      foam.notes.getNotes().forEach(n => {
+        graph.nodes.push({
+          id: n.id,
+          type: "note",
+          uri: n.source.uri,
+          title: path.basename(n.source.uri),
```
ingalless, line 31: Markdown links grabs the title of the file. Should we be maintaining that functionality? Or is this more a proof of concept PR that we can refine later?

ingalless (commented):
Looks good! I tried the default graph on my foam workspace and works flawlessly. Very speedy too.

One further improvement I'd add is clicking a node could navigate to the file.

packages/foam-vscode/src/features/dataviz.ts
riccardoferretti: `==` will compare both against `null` and `undefined`, whereas `===` would only do `null` (and we'd need another check for the `undefined` case).
I don't love it (it's the only place where I don't use the triple equal), but I like even less having both checks.

As I was writing I realized we have a `isNotNull` fn, we could create something along the same lines. 
To avoid loaded names, maybe we can use `isSome` / `isNone` (inspired by FP)

packages/foam-vscode/src/features/dataviz.ts
riccardoferretti: that's a great point, and a wider one in that regard.

In the context of this line of code, it should really be `title: n.title`.

I had a conversation with @jevakallio some time ago about how to extract the `title` from filenames/headings/yaml.
Long story short (happy to elaborate offline), the idea is to go back to simplicity, where filenames are what we can rely on, and use those. We could make this more configurable, the wider point is that this needs to be reconsidered.

Let me know if this makes sense :)

riccardoferretti:
yup, once we are happy with a direction, I need to do the following before releasing:
- [ ] click a file maintains the highlighted state (useful if you want to navigate the connections of a specified node)
- [ ] cmd+click navigates to the file
- [ ] remove the other implementations

stretch: 
- [ ] slight clean up of the code to prepare it for extensability
- [ ] provide extension points (e.g. to style the graph to one's liking)

Is there anything I am missing?

Other ideas for the future:
- [ ] highlight by text search
- [ ] show/hide tags

packages/foam-vscode/src/features/dataviz.ts
riccardoferretti: more context on this: https://github.com/foambubble/foam/blob/master/docs/foam-file-format.md

jmg-duarte:
I am unable to setup the graph. I've cloned the repo, changed branches and put the folder on the `.vscode/extensions`.
When trying to show the graph for my workspace I get the following
image

jmg-duarte:
Meanwhile I got up and running with some help from @ingalless 
However my graph looks like this
Screenshot from 2020-10-20 19-13-19
And well, there should be links. Any clue? 

Also, how do I change between graph engines?

riccardoferretti:
@jmg-duarte what problem was causing the issue with the command?

Also, could you share a sample repo to reproduce the problem with the graph? I'd be happy to debug it

jmg-duarte:
@riccardoferretti I've made my repo public you can find it here https://github.com/rustype/notes
In hindsight, something that might be causing this behavior may be because the links are not done in the "Foam" way but rather manually inserted to make sure they work in the GitHub repo

jmg-duarte:
As predicted the link shows when created the "Foam" way.

Screenshot from 2020-10-21 21-39-33

I have some suggestions about how to address this issue, as in my opinion it should still show the links.
Checking for foam-style links, with double braces should consider that some links may be hyperlinks (in the form of `[[link-title](link)]` rather than foam ones.
From here we can:
- Render every link destination as a node
- Render only the file system ones, ignoring the ones which start with `http://`

--- 

I have a few suggestions for the graph and I am interested in helping implement them, I hope this is the right place as this is the PR for the graph.

The graph does not have hot-reload nor a easy way of refreshing it.

About the extension points, I think it would be helpful to enumerate possible ones:
- Theme (as you can see in my screenshots, they don't match)
- Text Alignment (centered, left or right aligned)
- Quick Stats (Such as number of nodes etc)
- Filters (I've seen people working on tags so it would go hand in hand)

riccardoferretti:
@jmg-duarte I totally agree on everything you say.

Currently the way links are defined needs to evolve (it's on my plate).
One of the main reasons for this graph PR was to surface limitations in the foam model itself so that we can drive more and more from there.
Your comment sounds like a great starting point 👍 

This conversation is unrelated to the graph itself, as it is about the link parsing, let's take it offline (feel free to create an issue to track it)

jevakallio (approved):
Rubber stamping for release 🦆

PR #297, opened 2020-10-24
Update daily-notes.md

Craskermasker:
explicitly state that using a subdirectory for daily notes will prevent them from appearing in the graph visualisation.

### DIFF
--- a/docs/daily-notes.md
+++ b/docs/daily-notes.md
@@ -23,6 +23,8 @@ These settings can be overridden in your workspace or global `.vscode/settings.j
 
 The above configuration would create a file `journal/note-2020-07-25.mdx`, with the heading `Journal Entry, Sunday, July 25`.
 
+Specifying a sub-directory for daily notes in this way will prevent them from appearing in the graph visualisation.
+
 ## Daily Note Templates
 
 In the future, Foam may provide a functionality for specifying a template for new Daily Notes and other types of documents.


ingalless:
@Craskermasker I think we'd need some more testing on this...
I store my daily notes in the subdirectory `/journal`, and they appear on my graph. See

image

Do you think you can chuck together a repo that demonstrates the issue? Then we can get some debugging on it.

Craskermasker:
Sorry, I think this is a misunderstanding. 

I *don't* want my daily notes appearing in the graph. as they clutter it up. I've put them in a different subdirectory `/dailies`. I thought this meant they were no longer indexed by the graph, but actually they are still on the graph, but not linked to any other nodes. 

Is it possible to have an option to exclude subdirectories from the graph while still allowing them to link to other nodes?

Craskermasker:
Actually, on looking more closely at the settings, I think I've achieved this by adding my subdirectory `**/dailies": true` to the `"files.exclude"` and `"files.watcherExclude"` settings.

ingalless:
@Craskermasker The only thing to consider with those two settings is that this will also remove them from your explorer too, which may be undesirable.

The graph for Foam is currently handled by markdown links. There's a PR [here](https://github.com/tchayen/markdown-links/pull/69) that may be of interest

Craskermasker:
yeah it's not optimal, but it works all right for now. Thank you. I'll have a look at that.

PR #79, opened 2020-07-08
Add open daily note base feature

francishamel:
Open Daily Note is a core feature of roam. That's why this feature is considered as a core feature in foam's roadmap. This feature is quite simple. With the `Foam: Open Daily Note` command, vscode will create today's note (`yyyy-mm-dd.md`) if it's not already created and focus the editor on today's note. It is also possible to trigger the command via the `Alt+d` keyboard shortcut.

Some things related to the daily note are configurable:

- Filename format
- File extension
- Title format of the daily note
- Daily notes directory

This PR implements the [Must Have](https://foambubble.github.io/foam/daily-notes#must-have) and the [Should Have](https://foambubble.github.io/foam/daily-notes#should-have).

Related issue: https://github.com/foambubble/foam/issues/54

PR #298, opened 2020-10-24
Add date snippets for daily notes

ingalless:
Hey

This adds support for daily note date-driven snippets. They all expand out to be the wikilink to the day, and then open the target date's note.

The following snippets are available:

- `/today` or `/day`: Today
- `/tomorrow`: Tomorrow
- `/{dayOfWeek}`: Get the day coming up in the week (e.g. `/monday`)
- `/+nd`: Add n days, where n is a number
- `/+nw`: Add n weeks, where n is a number
- `/+nm`: Add n months, where n is a number
- `/+ny`: Add n years, where n is a number

In future, a weekly/monthly/yearly note should open the configured "weekly note" template (once templates are implemented).

## Today, Tomorrow, +n days/weeks/months/years
snippets

## Days of week
snippets

### DIFF
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -10,7 +10,7 @@ import dateFormat from "dateformat";
 import * as fs from "fs";
 import { docConfig } from "./utils";
 
-async function openDatedNote(date?: Date) {
+async function openDailyNoteFor(date?: Date) {
   const foamConfiguration = workspace.getConfiguration("foam");
   const currentDate = date !== undefined ? date : new Date();
 
@@ -98,4 +98,4 @@ function pathExists(path: string) {
     .catch(() => false);
 }
 
-export { openDatedNote, getDailyNoteFileName };
+export { openDailyNoteFor, getDailyNoteFileName };
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -21,7 +21,7 @@ async function openDailyNoteFor(date?: Date) {
     dailyNotePath,
     currentDate
   );
-  await focusDailyNote(dailyNotePath, isNew);
+  await focusNote(dailyNotePath, isNew);
 }
 function getDailyNotePath(configuration: WorkspaceConfiguration, date: Date) {
   const rootDirectory = workspace.workspaceFolders[0].uri.fsPath;
@@ -79,8 +79,8 @@ async function createDailyNoteDirectoryIfNotExists(dailyNotePath: string) {
   }
 }
 
-async function focusDailyNote(dailyNotePath: string, isNewNote: boolean) {
-  const document = await workspace.openTextDocument(Uri.file(dailyNotePath));
+async function focusNote(notePath: string, isNewNote: boolean) {
+  const document = await workspace.openTextDocument(Uri.file(notePath));
   const editor = await window.showTextDocument(document);
 
   // Move the cursor to end of the file
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -8,7 +8,7 @@ import {
 import { dirname, join } from "path";
 import dateFormat from "dateformat";
 import * as fs from "fs";
-import { docConfig } from "./utils";
+import { docConfig, pathExists } from "./utils";
 
 async function openDailyNoteFor(date?: Date) {
   const foamConfiguration = workspace.getConfiguration("foam");
@@ -89,13 +89,6 @@ async function focusNote(notePath: string, isNewNote: boolean) {
     const { range } = editor.document.lineAt(lineCount - 1);
     editor.selection = new Selection(range.end, range.end);
   }
-}
-
-function pathExists(path: string) {
-  return fs.promises
-    .access(path, fs.constants.F_OK)
-    .then(() => true)
-    .catch(() => false);
 }
 
 export { openDailyNoteFor, getDailyNoteFileName };
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -1,0 +1,101 @@
+import {
+  Selection,
+  Uri,
+  window,
+  workspace,
+  WorkspaceConfiguration
+} from "vscode";
+import { dirname, join } from "path";
+import dateFormat from "dateformat";
+import * as fs from "fs";
+import { docConfig } from "./utils";
+
+async function openDatedNote(date?: Date) {
+  const foamConfiguration = workspace.getConfiguration("foam");
+  const currentDate = date !== undefined ? date : new Date();
+
+  const dailyNotePath = getDailyNotePath(foamConfiguration, currentDate);
+
+  const isNew = await createDailyNoteIfNotExists(
+    foamConfiguration,
+    dailyNotePath,
+    currentDate
+  );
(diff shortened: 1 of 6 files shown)


packages/foam-vscode/src/dated-notes.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  return `${dateFormat(date, filenameFormat, false)}.${fileExtension}`;
+}
+
+async function createDailyNoteIfNotExists(
+  configuration: WorkspaceConfiguration,
+  dailyNotePath: string,
+  currentDate: Date
+) {
+  if (await pathExists(dailyNotePath)) {
```
jsjoeio, line 54: TIL you can await within the condition of an `if` statement 🤯

jsjoeio (approved):
Wow! Amazing work @ingalless and what a great feature to add 🎉

packages/foam-vscode/src/dated-notes.ts
ingalless: I don't think this was actually me...this looks like some moved code (daily notes were refactored to share code) So TIL too!!! 😅

ingalless:
Following a recommendation from discord I added a `/yesterday` command

packages/foam-vscode/src/dated-notes.ts
anku255: It was @francishamel in the PR - https://github.com/foambubble/foam/pull/79/files#diff-c620ef40775aeffc01b5a8468974b524e8f1ce4dfcc842f73ffc0dbf34e53227R60

anku255 (approved):
Great work @ingalless. Love this feature. Thank you!

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -0,0 +1,101 @@
...
+
+async function createDailyNoteDirectoryIfNotExists(dailyNotePath: string) {
+  const dailyNoteDirectory = dirname(dailyNotePath);
+
+  if (!(await pathExists(dailyNoteDirectory))) {
+    await fs.promises.mkdir(dailyNoteDirectory, { recursive: true });
+  }
+}
+
+async function focusDailyNote(dailyNotePath: string, isNewNote: boolean) {
```
riccardoferretti, line 82: minor: this code would work for any file, not just daily notes - might be worth adjusting the name to describe that?

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -0,0 +1,101 @@
...
+  const document = await workspace.openTextDocument(Uri.file(dailyNotePath));
+  const editor = await window.showTextDocument(document);
+
+  // Move the cursor to end of the file
+  if (isNewNote) {
```
riccardoferretti, line 87: iiuc this will move the cursor to the end of the of file if it is a new note:
- shouldn't this happen when the note is _not_ new?
- or, maybe this should just always happen? (if the note is new the beginning and end of file will match, so nothing material changes, but the code is cleaner)

Let me know if I am misunderstanding something

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -0,0 +1,101 @@
...
+    const { lineCount } = editor.document;
+    const { range } = editor.document.lineAt(lineCount - 1);
+    editor.selection = new Selection(range.end, range.end);
+  }
+}
+
+function pathExists(path: string) {
```
riccardoferretti, line 94: this is very handy! I actually needed something like this before and my solution was not as clean.
Can you move it to a utils module so that other features can also use it?

packages/foam-vscode/src/features/open-dated-note.ts
```diff
@@ -0,0 +1,205 @@
...
+    const today = new Date();
+    return {
+      detail: "Insert a link to yesterday's daily note",
+      snippet: "/yesterday",
+      date: new Date(today.getFullYear(), today.getMonth(), today.getDate() - 1)
+    };
+  }
+];
+
+const computedSnippets: ((number: number) => DateSnippet)[] = [
```
riccardoferretti, line 111: I really like this approach!

packages/foam-vscode/src/features/open-dated-note.ts
```diff
@@ -0,0 +1,205 @@
...
+};
+
+const createCompletionItem = ({ snippet, date, detail }: DateSnippet) => {
+  const completionItem = new CompletionItem(
+    snippet,
+    CompletionItemKind.Snippet
+  );
+  completionItem.insertText = getDailyNoteLink(date);
+  completionItem.detail = `${completionItem.insertText} - ${detail}`;
+  completionItem.command = {
```
riccardoferretti, line 63: do I understand correctly that this will also automatically open the note?
will it be possible to add the link without navigating to the note?

riccardoferretti (commented):
This is looking really good @ingalless, I left a few comments, let me know what you think!

packages/foam-vscode/src/dated-notes.ts
ingalless: I wonder if the reasoning is that, if you've already opened the file before, then you'll have moved the cursor. An example:

I'm editing a specific heading halfway down the page in today's daily note. Now I jump out to have a look at a related thought. When I jump back in (using the Focus Daily Note command) I don't want to jump to the end of the file, I'm expecting the cursor to be where it was when I left?

I only moved this code, so I don't know if that was the intent, but that strikes me as plausible

packages/foam-vscode/src/features/open-dated-note.ts
ingalless: Yes it will.

It makes a good candidate for a config option, but perhaps that's best saved for a follow up PR? Up to you, I'm happy to address it as it shouldn't be too long to add in.

packages/foam-vscode/src/features/open-dated-note.ts
ingalless: Thanks!!

packages/foam-vscode/src/dated-notes.ts
riccardoferretti: yeah, that makes a lot of sense!

packages/foam-vscode/src/features/open-dated-note.ts
riccardoferretti: Intuitively I would feel that the snippet shouldn't navigate away automatically for a couple of reasons (both very common e.g. in my usage):
- I mainly use the snippet to not worry about the daily note format and have shortcuts for text expansion
- I don't always/necessarily want to create an actual note, an "empty" placeholder is usually enough (so that in the graph things that happened in the same day are connected)
- I can still easily navigate through the link by cmd+click (or F12)
- I could go back to the original file, but it feels it interferes with the writing flow


My suggestion would be:
- by default only expand the text without navigating
- have a configuration option to add the navigation
- (not sure if possible, and probably wouldn't worry about this in this PR unless trivial) have a key modifier on the snippet (e.g. cmd+enter) to expand + navigate

wdyt?

ingalless: Yeah, you're right. I think because I wanted to validate the file destination was correct I ended up with this behaviour 😆

I've pushed a new commit which addresses this now. If you'd like the settings name changed lmk and I'll get to it!
The key modifier doesn't _seem_ possible, but defo one to investigate for a separate PR

riccardoferretti:
Looking great!

riccardoferretti (approved):

Issue #300, opened 2020-10-28
The link generator should take into account ignored folders


jmg-duarte:
I'm currently trying to build a site for https://github.com/rustype/notes using https://github.com/mathieudutour/foam-gatsby-template.
However, when building locally (for debug purposes) the wiki-link generator starts using the `.md` files from the `_layouts` folder.

This should be an easy fix, add an `ignore` setting as follows:
```json
"foam.edit.linkReferenceDefinitions.ignore": {
  "type": [
    "array"
  ],
  "default": []
}
```

And filter out files whose path is prefixed with any of the strings present in the array.
From my brief research, I think maybe the following function would be a nice place to filter the files out:

https://github.com/foambubble/foam/blob/532348633b22002088973b4fd608867b50670f9e/packages/foam-vscode/src/features/wikilink-reference-generation.ts#L131-L171

Probably after checking with `uniq`.
However, the number of operations may be too high in some cases, making the filtering slow, the paths could be split and represented as a tree, then filtering is easy as any branch with a given path can be removed.

riccardoferretti:
hey @jmg-duarte, thanks for reporting this bug, can you tell me which file and definition is being incorrectly generated? I will try reproduce it to better understand the issue.
A configuration option to exclude some links might be a good solution, but let me first fully understand the problem at hand, thanks!

jmg-duarte:
I was trying to reproduce the problem to upload it as a branch but I was unable to.
Either way, what happened was:
- I built the site under `_layouts` using https://github.com/mathieudutour/foam-gatsby-template
- When the link generator kicked in, instead of using my actual markdown files, it would get the rendered `.md` files present in the `_layout`

I'd end up with a final link like:
```
[//begin]: # "Autogenerated link references for markdown compatibility"
[inbox]: _layouts/public/some_really_long_hash/inbox "Inbox"
[//end]: # "Autogenerated link references"
```

riccardoferretti:
it sounds like foam was parsing the files inside your `_layout` directory (which would make sense) and that would cause a conflict with the regular files.

The solution here is telling foam to ignore the files in the `_layout` directory altogether.
More in general, the solution is to support glob include/exclude patterns in foam.

jmg-duarte:
Exactly! I am available to help implement the feature

ingalless:
@jmg-duarte at least to get you started for now, you may want to specify the `files.exclude`.

For reference, here's our companies `files.exclude` value in the `.vscode/settings.json`

```json
[...]

PR #304, opened 2020-10-28
Add file ignore setting (issue #300)

jmg-duarte:
This PR is related to issue #300, the issue will be closed and discussion moved here.

This serves as a basic implementation for an ignore mechanism for Foam related files, this is useful to avoid files appearing in the graph (see below).

Before:
image

After:
image

Standing problems:
- [X] Maybe this should be glob based, to be discussed.

- [ ] Settings do not refresh, this also happens for other Foam things, such as the graph.

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -48,6 +48,13 @@
     "configuration": {
       "title": "Foam",
       "properties": {
+        "foam.files.ignore": {
+          "type": [
+            "array"
+          ],
+          "default": [],
+          "description": "List of files/folders Foam should ignore. Paths are relative to the workspace root."
+        },
         "foam.edit.linkReferenceDefinitions": {
           "type": "string",
           "default": "withoutExtensions",
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -52,7 +52,10 @@
           "type": [
             "array"
           ],
-          "default": [],
+          "default": [
+            ".vscode/*",
+            "_layouts/*"
+          ],
           "description": "List of files/folders Foam should ignore. Paths are relative to the workspace root."
         },
         "foam.edit.linkReferenceDefinitions": {
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -56,7 +56,7 @@
             ".vscode/*",
             "_layouts/*"
           ],
-          "description": "List of files/folders Foam should ignore. Paths are relative to the workspace root."
+          "description": "Specifies the list of globs that will be ignored by Foam (e.g. they will not be considered when creating the graph)."
         },
         "foam.edit.linkReferenceDefinitions": {
           "type": "string",
--- a/packages/foam-vscode/src/extension.ts
+++ b/packages/foam-vscode/src/extension.ts
@@ -23,6 +23,7 @@ import {
 } from "foam-core";
 
 import { features } from "./features";
+import { getIgnoredFilesSetting } from "./settings";
 
 let workspaceWatcher: FileSystemWatcher | null = null;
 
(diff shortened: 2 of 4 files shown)


riccardoferretti:
thanks @jmg-duarte, this is looking good. The only question I have is why not using globs?
It sounds like a more solid/flexible/standard solution, what's your thinking around this?

jmg-duarte:
The glob library goes hand in hand with the file system and is not usable for matching along lists.

riccardoferretti:
I am missing something, why is that a problem?
I also noticed the `workspace.findFiles` function also supports globs with explicit `exclude` support, would that be a viable option?

jmg-duarte:
We could use the globs if we put both the list of files found and the ignored ones in sets and intersected them.
Otherwise we would could be comparing a lot of files vs a lot of more files.

> I also noticed the workspace.findFiles function also supports globs with explicit exclude support, would that be a viable option?

Probably, I'll do some research on it

jmg-duarte:
It seems to me that the `workplace.findFiles` function does not work for this use case due to the limitation of supporting a single exclude.
However I'm curious about the set intersection approach and plan to try and profile both.

riccardoferretti:
hi @jmg-duarte, did you get a chance to look into this? No rush, was thinking of shipping it with 0.5.0 (in one week), but wanted to check your status, thanks!

jmg-duarte:
I got sidetracked due to the lack of big repositories to profile with.
However I believe the set intersection approach is better (at least in big repositories) since comparison is hash-based and not string-prefix.
I can implement the set based approach to be merged and if necessary in the future we switch.

jmg-duarte:
I tried to use a set but the `Uri` objects do not "equal" reliably *Javascript 🤷*, to cope with it I replaced the Set with a Map, comparing on keys, which in this case is the `fsPath`.
I didn't test performance wise, but I'd say this approach is probably faster.
And even if it is not, it is MUCH cleaner and easier to read/understand.

riccardoferretti (approved):
LGTM!

packages/foam-vscode/package.json
```diff
@@ -48,6 +48,13 @@
     "configuration": {
       "title": "Foam",
       "properties": {
+        "foam.files.ignore": {
+          "type": [
+            "array"
+          ],
+          "default": [],
+          "description": "List of files/folders Foam should ignore. Paths are relative to the workspace root."
```
riccardoferretti, line 56: this needs to be updated right? Might also be worth adding an example in parenthesis for clarity

riccardoferretti (approved):
Still looks good to me :) but let's update the description of the property so that it's clearer for users

packages/foam-vscode/package.json
jmg-duarte: Instead of an example we can set a default `.vscode/*` and `_layouts/*`for example.
Description wise I'll try to make it clearer since the description is out of date

riccardoferretti (approved):

patchworquill:
Hi all -- having trouble with this. My `foam.files.ignore` entry doesn't seem to be blocking the globbed folders from appearing in my Foam Graph.

jmg-duarte:
> Hi all -- having trouble with this. My `foam.files.ignore` entry doesn't seem to be blocking the globbed folders from appearing in my Foam Graph.

Can you show your entry?

patchworquill:
Fixed. Didn't realize I had to `git commit` for the settings to take effect. 

I did the following:

- removed all the extras within _layout (probably not necessary but it felt like clutter since I'm not yet publishing online)
- added "/.vscode" and "/_layouts" to the settings.json file _within_ the /.vscode folder in my repo (not the global settings file, although likely that would work fine as well)
```
  "foam.files.ignore": [
    "/.vscode",
    "/_layouts",
  ]
```
- added and commited these changes in git

Now my graph is once again clean and useable. :)

PR #310, opened 2020-10-30
Graph v0.2 - fixing bugs and adding labels

riccardoferretti:
fix graph related bugs and lint code


riccardoferretti:
> I wonder if we should do a format PR over the whole codebase. A lot of this code was formatting, so I found it quite hard to spot the changes. What do you think?

Yeah my bad, we should run prettier as part of the workflow (I have it set up in vscode, and the rules are in the root of the repo, so they are consistent), it was my bad for not running the formatter when I first wrote the file.
We have checks in github (warnings mostly), somehow I missed them on my previous PR.

Issue #302, opened 2020-10-28
Add basic tag support


riccardoferretti:
Add support for tags in the foam workspace.

Each note would be able to define tags in 2 ways:
- by having a `tags` property in YAML
  - syntax for this property will be lose, so all of the following will yield the same result:
    - `tags: tag1, tag2`
    - `tags: tag1 tag2`
    - `tags: #tag1 tag2`
    - `tags: [tag1, #tag2]`
- by having `#tag` anywhere in the file

---

- [ ] Support defining tags via YAML
- [ ] Support defining tags via `#tag`
- [ ] Add a new tag panel which lists all the tags in the workspace in alphabetical order
- [ ] Clicking on a tag in the panel will trigger a search for the tag in the workspace
- [ ] Cmd+Clicking on a tag in the text will trigger a search for the tag in the workspace
- [ ] (extra) style tags in text (check interaction with Markdown Notes)

jmg-duarte:
I was led here by the necessity to move from categories to tags in my work (inspired by [Zettelkasten](https://zettelkasten.de/posts/no-categories/)).
I think that the tag should be denoted in the note file itself. *Example:*
```md
# Super Awesome Title

---
@tags: t1, t2, t3
---

Note description
```
We can prepend the `tags` line with a special marker (e.g. `@`) to declare it as metadata.
This metadata could then be extracted to the `foam.json` file to create a list (or map to allow descriptions) of tags (furthermore, the janitor could then generate a special file containing all tags).

The metadata symbol could also be configured in `foam.json` or in the `settings.json`, in this case is a matter of "what really should be the purpose of `foam.json`".

Personally, I hate YAML, it's whitespace rules are awkward and hard to debug. I also believe that since Foam already makes plenty use of JSON, we should keep everything in the same format, in the name of simplicity.

canpolat:
I think YAML front-matter is the best place to keep all metadata (including tags) about a note. But using `#` in front-matter may render the resulting YAML invalid. `#` is used for comments in YAML and the following line will definitely be invalid YAML (which may cause problems if the user wants to view the file in another editor). 

```
tags: [tag1, #tag2]
```

Other than that, it would also be great if the tags in YAML front-matter had auto-complete functionality.

riccardoferretti:
thx for the info @canpolat! I didn't know about that, will review the YAML docs and give an update about the specs for this PR

riccardoferretti:
this is addressed by #311

[...]

PR #311, opened 2020-10-30
Adding support for tags

riccardoferretti:
Based on #302 

Add support for tags in the foam workspace.

Each note would be able to define tags in 2 ways:
- by having a `tags` property in YAML
  - syntax for this property will be lose, so all of the following will yield the same result:
    - `tags: tag1, tag2`
    - `tags: tag1 tag2`
    - `tags: [tag1, tag2]`
- by having `#tag` anywhere in the file

---

- [x] Support defining tags via YAML
- [x] Support defining tags via `#tag`
- [x] Add a new tag panel which lists all the tags in the workspace in alphabetical order
- [x] Clicking on a tag in the panel will trigger a search for the tag in the workspace
- [ ] (extra) Cmd+Clicking on a tag in the text will trigger a search for the tag in the workspace
- [ ] (extra) style tags in text (check interaction with Markdown Notes)


packages/foam-core/src/markdown-provider.ts
```diff
@@ -8,26 +8,22 @@ import { Parent, Point } from 'unist';
...
-        column: 1,
-        offset: node.position!.end.offset! + 1,
-      };
-    }
+const tagsPlugin: ParserPlugin = {
+  onWillVisitTree: (tree, note) => {
+    note.tags = extractHashtags(note.source.text);
+  },
+  onDidFindProperties: (props, note) => {
+    const yamlTags = extractTagsFromProp(props.tags);
```
ingalless, line 25: We can probably skip the intermediary setting of this variable, and instead chain from the function?

packages/foam-core/src/utils/hashtags.ts
```diff
@@ -0,0 +1,22 @@
+import { isSome, isNumeric } from './core';
+import matchAll from 'string.prototype.matchall';
```
ingalless, line 2: Woah, I didn't know you could do this!

packages/foam-core/test/markdown-provider.test.ts
```diff
@@ -196,6 +196,54 @@ describe('wikilinks definitions', () => {
...
+    );
+    expect(noteA.tags).toEqual(new Set(['text', 'tags', 'care-about']));
+  });
+
+  it('can find tags as text in yaml', () => {
+    const noteA = createNoteFromMarkdown(
+      '/dir1/page-a.md',
+      `
+---
+tags: hello, world  this_is_good
```
ingalless, line 217: Does the double space here result in "world" and "this_is_good" separate tags? This is just curiousity on my part! Seems weird to me

ingalless (approved):
Good work @riccardoferretti! Made a couple comments, but they aren't anything that stop me approving 👍

packages/foam-core/src/utils/hashtags.ts
riccardoferretti: unfortunately atm we use node 10, which doesn't have `string.matchAll`, so I added this shim as a dependency (this is an actual dependency, the package name can be misleading though!). 
I don't love it, but updating to node 12 just for this seemed a bit overkill.
Nonetheless I see no problem in doing the update in the future, just wanted to keep it isolated and understand if there are possible side effects :)

packages/foam-core/test/markdown-provider.test.ts
riccardoferretti: I actually did it on purpose :) 
I am mixing various separators in this test and wanted to validate they can be mixed (basically I don't want to be strict about the syntax of the `tags` property, people might have different conventions)

PR #312, opened 2020-11-02
Fix error in cli command janitor & migrate

hikerpig:
I got `workspaceFolders.reduce is not a function` error during calling command 'migrate' from 'foam-cli'.

Turns out it was caused by calling 'createConfigFromFolders' with `string` while the function signature requires `string[]`.


riccardoferretti (approved):
LGTM!

Issue #317, opened 2020-11-04
["BUG"] Foam: Show graph' fails with "All collection items must start at the same column"


fmagin:
**Describe the bug**

The new graph does not work or it is unclear what is required to make it work.

**Affected package**
unclear

**To Reproduce**
Upgrade to v0.4

Run command `Foam: Show Graph`

`Command 'Foam: Show graph' resulted in an error (All collection items must start at the same column)`

**Expected behavior**
Whatever that command is actually supposed to do.

**Additional context**
https://github.com/foambubble/foam/issues/301 brings up yaml (i.e. frontmatter) as a possible cause for a different error. I also have that in some files and might investigate this. Still I do not think that this feature should break simply because I am using YAML, without ever being told how YAML interacts with it.

fmagin:
Running the command from the other issue and restarting does seem to fix it. Still I consider this a bug, because this forces me to conform to some unknown frontmatter spec (which I'd like to use for other purposes or extensions).

riccardoferretti:
@fmagin sounds like there is a syntax problem in your YAML metadata. I agree that shouldn't cause the whole thing to break, we need to push some error handling in the document parsing.

riccardoferretti:
this has now been fixed by #320

Issue #316, opened 2020-11-04
["BUG"] Wiklinks not being generated with v0.4 "command 'foam-vscode.update-wikilinks' not found"


fmagin:
**Describe the bug**

No wikilinks are being generated on file save.

Specifically, when manually running the command `Foam: Update Markdown Reference List` I get the error message:

`Command 'Foam: Update Markdown Reference List' resulted in an error (command 'foam-vscode.update-wikilinks' not found)
`

**Affected package**
Probably mainly `foam-vscode` due to the above error

**To Reproduce**

Not fully clear, but I can reproduce this on two different installs and was most likely caused by the update to the v0.4 version

**Expected behavior**
Generate the links like it used to happen

fmagin:
The error message seems to disappear if I strip all frontmatter metadata like in #301 but the links still aren't generated

riccardoferretti:
hi @fmagin, regarding the frontmatter, we are adding better error handling around it.
Regarding the fact that they are still not generated, can you check in your settings the value of `Link Reference Definitions`? if it's `off` the references won't be generated

fmagin:
That is set to `withExtensions`.

fmagin:
I think this specific bug is fixed with v0.5, and I found out more about second bug:

The name of the link, i.e. the text between `[[` and `]]` has to match the target file name (without the `.md`). So I guess the links are not normalized by applying the appropriate slugger before checking if the appropriate file exists?

riccardoferretti:
at the moment we rely on markdown-notes to connect the wikilinks to the notes. on our immediate roadmap we have to support links based on the foam model

fmagin:
But this used to work before, and seems like a major regression to me. Are you sure we are talking about the same issue? The behavior I am encountering breaks the links in the preview pane, and the ones links detected by the Markdown Links Addon. Those links would then also not appear on the Foam graph either.

riccardoferretti:
> So I guess the links are not normalized by applying the appropriate slugger before checking if the appropriate file exists?

Maybe it is something related to this, although I don't believe this has been changed (I might be wrong).

Do you have an example to reproduce the issue?

riccardoferretti:
does it sound like #323 ?

fmagin:
Yes, that is the same issue I am encountering. Closing this issue then, because it wasn't primarily about this problem.

riccardoferretti:
sounds good, let's track it there

PR #320, opened 2020-11-06
improved error handling when parsing markdown files

riccardoferretti:
This fix makes the parsing more friendly towards badly formatted md files (error occurring especially in YAML parsing).
will address #317 (and part of #316)

The PR does the following:
- add a `name` property to plugins, to better identify the source of problems
- wrap all the calls to plugin functions in `try/catch` blocks, so to report any error occurred in there (and continue)
- wraps the yaml parsing in a `try/catch` block, to report any error, for now in the console, and continue with the parsing


ingalless (approved):
Looks great! Good work @riccardoferretti 

Maybe in future we can show some of these errors (that the user can fix, for example) as vscode notifications?

riccardoferretti:
will merge in the meantime to fix the crashes, but that's a very good point, we should do that soon

Issue #325, opened 2020-11-09
Long titles should be cut or wrapped


jmg-duarte:
This issue is a "duplicate" of https://github.com/tchayen/markdown-links/issues/66
The problem is the same, and the solution (at least for cutting, is as well)

![image](https://user-images.githubusercontent.com/15343819/98575269-e1a1d500-22b0-11eb-8b66-2dd83955f75f.png)

From the picture I think the problem is obvious

jmg-duarte:
Closing the issue as it is addressed in #327

PR #327, opened 2020-11-09
Add title cutting after given length

jmg-duarte:
This PR is related with https://github.com/foambubble/foam/issues/325, the approach is the same as https://github.com/tchayen/markdown-links/pull/67/
The graph now looks like this (with length of 24)

image

I tried to wrap lines, however `\n` gets translated to a space.


riccardoferretti (approved):
This looks really good!

re new line, the graph uses (in `graph.js`) `Canvas.fillText` function, which doesn't support multiline. Your approach is perfect

PR #344, opened 2020-11-10
Refactor the bootstrap function to be simpler

jmg-duarte:
Currently the bootstrap function has a lot of bloat.
It makes the files go around several times instead of applying the filters in one go.

This PR aims to reduce the bloat and with it add some documentation (a starting effort!)

I'd also like to add that file registering would make more sense as: `foam.registerFile(file)` instead of `registerFile(foam, file)`.


riccardoferretti (approved):
I am also looking at this part of the code in the context of my branch, so this might change again soon - but in the interim it looks good to me!

jmg-duarte:
So, how should we proceed?

riccardoferretti:
I am happy for this to be merged in the meantime, I don't know the extent of my possible changes, but this is better than what we have in any case :)

PR #346, opened 2020-11-11
Feature/templates

ingalless:
Todo:

- [ ] Conflicts!
- [ ] Exclude `.foam/` directory from NoteGraph
- [ ] Tests? (Although this is very "scripty". Only the `getTemplates` could probably be tested unless we have experience in integration testing vs code extensions)

**Future PR?**
- [ ] Setting a daily note template to be the ID of one of the templates
- [ ] A "create new template" command that can also be presented when no templates exist!

@riccardoferretti you're assigned as a reviewer, but don't worry about properly reviewing yet!


jmg-duarte:
> Exclude .foam/ directory from NoteGraph
With this, do you mean, not consider the `.md` files under `.foam/`?

If so https://github.com/foambubble/foam/pull/304 introduces support for file ignore, you can just add `.foam/**/*` or `.foam/**/*.md`

ingalless:
> > Exclude .foam/ directory from NoteGraph
> 
> With this, do you mean, not consider the `.md` files under `.foam/`?
> 
> If so https://github.com/foambubble/foam/pull/304 introduces support for file ignore, you can just add `.foam/**/*` or `.foam/**/*.md`

I found where to make the change, but a few tests failed as a result. I'd want some extra tests in place to prove it works too. Thank you for the work you've done in #304! I was pleasantly surprised to see how easy a change it was.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -0,0 +1,66 @@
...
+          });
+          const title = await window.showInputBox({
+            prompt: `Enter the Title Case name for the new note`,
+            value: ``,
+            validateInput: value =>
+              value.length ? undefined : "Please enter a value!"
+          });
+          const targetFile = path.join(
+            folder,
+            `${new GithubSlugger().slug(title)}.md`
```
riccardoferretti, line 48: what's your thinking around this? I feel we are forcing a (our) convention over users by slugifying the filename.
I wonder if it should be a setting (e.g. in my notes, the title is the filename, without slugification - same in Obsidian), or (and personally this is my preference) whether we just ask for the filename, and then it's up to the template to use the filename as it pleases (e.g. to deslugify it, or simply use it as title) - what do you think?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -0,0 +1,66 @@
...
+          const templateText = await workspace.fs.readFile(
+            Uri.file(`${templatesDir}/${selectedTemplate}`)
+          );
+          const snippet = new SnippetString(templateText.toString());
+          await workspace.fs.writeFile(
+            Uri.file(targetFile),
+            new TextEncoder().encode("")
+          );
+          await focusNote(targetFile, true);
+          await window.activeTextEditor.insertSnippet(snippet);
```
riccardoferretti, line 59: amazing! this way we can leverage all the variable substitutions! love it!

packages/foam-vscode/src/utils.ts
```diff
@@ -162,6 +165,20 @@ export function isSome<T>(value: T | null | undefined | void): value is T {
...
  * @param value The object to verify
  */
-export function isNone<T>(value: T | null | undefined | void): value is null | undefined | void {
+export function isNone<T>(
+  value: T | null | undefined | void
+): value is null | undefined | void {
   return value == null;
 }
+
+export async function focusNote(notePath: string, isNewNote: boolean) {
```
riccardoferretti, line 174: yeah, this makes sense as a utility function.
I know it's unrelated to this PR, but as we are officially making this a utility fn what do you think of making the semantics to be "utility-like"? 
We could rename `isNewNote` with `moveCursorToEnd` or something like that, so it's easier for clients of the fn to understand the result of the parameter value

riccardoferretti (commented):
definitely looking good! I have a comment around some semantics and a nitpick around the utility fn - let me know what you think!

packages/foam-vscode/src/utils.ts
ingalless: Resolved!

packages/foam-vscode/src/features/create-from-template.ts
ingalless: The latter is the easiest to implement for now as you'll notice there's no "special" substitution support (e.g. title). I agree, probably not a good idea to force our convention.

ingalless: Should be resolved now!

ingalless:
@riccardoferretti Implemented your feedback, thanks for the swift response.

This PR isn't feature-complete, but it provides enough to prove the concept. Hopefully we can iterate on this!

riccardoferretti (approved):
LGTM!

jmg-duarte:
Great job @ingalless! I'd just suggest adding some documentation along with the new functions.
If we start adding documentation to the functions we work on not only it becomes a less tedious job as well as an easier one in the future!

ingalless:
@jmg-duarte that's a great idea! Mind raising an issue for it?

jmg-duarte:
Done. #348

jmg-duarte (approved):

PR #349, opened 2020-11-17
Keep foam in sync with file system

riccardoferretti:
This PR consolidates the file management aspect of foam inside foam-core.
- simplifies bootstrap
- fixes various incongruences between bootstrap settings in foam-core and foam-vscode
- watches for file changes to reload the docs in foam

### Concepts
- adds a few utility modules from vscode, especially around lifecycle and event management
  - `IDisposable` manages lifecycle for objects that hang on to resources (e.g. listeners, watchers, ...). it's responsibility of the object that created a disposable to clean up after itself.
  - `Event` class will be the base for event management in foam (it leverages a lot of good design and optimizations built for vscode)
- introduces the idea of service in foam (for now data store, and a dummy logger)
  - this will allow e.g. to inject different loggers based on whether we are running foam in CLI or vscode
- introduces DataStore abstraction
  - the data store abstracts where files are fetched from
  - the first implementation uses the file system as a store backend 
- introduces watcher for changes in data store
  - the watcher monitors a data store and fires events when changes are detected

### Architectural note
Foam model is what all clients and features interact with. Foam is the only interface with the underlying data store. 
This means e.g. that if the graph wants to update when changes are detected, it listens for changes in Foam, not in the FS or in the data store directly.
In the future Foam will be even more specific about events, e.g. not just notifying that a note has changed, but also what has changed in it. This will allow clients to listen for changes in links, tags, sections (when implemented) and other foam metadata.
It will still be possible to listen to the generic `noteChanged` event for not yet supported use cases.

### Other work scheduled for this PR
- [x] the tags feature is now hooked up to the foam events to validate the work above.
- [x] foam graph events need to use the new `Event` model
- [ ] ~graph dataviz needs to listen for updates~ I have decided against implementing that, because it's easy to add the updates, but currently they scramble the graph (pretty annoying), so this task has been moved to #358


packages/foam-vscode/src/services/config.ts
```diff
@@ -0,0 +1,17 @@
+import { workspace } from "vscode";
+import { FoamConfig, createConfigFromFolders } from "foam-core";
+import { getIgnoredFilesSetting } from "../settings";
+
+// TODO this is still to be improved - foam config should
+// not be dependent on vscode but at the moment it's convenient
+// to leverage it
+export const getConfigFromVscode = (): FoamConfig => {
```
jmg-duarte, line 8: This goes back to the discussion of how coupled Foam and VSC should be. Maybe this could be escalated as an issue striving for clearer goals?

packages/foam-vscode/src/services/vscode-watcher.ts
```diff
@@ -0,0 +1,29 @@
+import { IDisposable, Emitter, URI, IDataStoreWatcher } from "foam-core";
+import { FileSystemWatcher } from "vscode";
+
+export class VscodeBaseWatcher implements IDataStoreWatcher, IDisposable {
```
jmg-duarte, line 4: Nitpick but shouldn't it be `VSCodeBaseWatcher`?

riccardoferretti: close, was double checking the vscode codebase to see what convention they use, and it's `VsCode`, so we could call this `VsCodeBaseWatcher`

ingalless (commented):
Great work Riccardo. I really like this! I've left a few comments

packages/foam-cli/src/commands/janitor.ts
```diff
@@ -38,8 +40,12 @@ export default class Janitor extends Command {
...
 
     if (isValidDirectory(workspacePath)) {
-      const graph = (await bootstrap(createConfigFromFolders([workspacePath])))
-        .notes;
+      const config = createConfigFromFolders([workspacePath]);
+      const services: Services = {
+        logger: console,
+        dataStore: new FileDataStore(config),
+      };
+      const graph = (await bootstrap(config, services)).notes;
```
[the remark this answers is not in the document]
riccardoferretti: the `bootstrap` could fail and would throw an exception, I think it's ok because the command couldn't work in any case (we could add a nicer error message, if that's what you are referring to, but in the end I wonder how much different it would be from the standard stack trace).
What was your concern specifically?

packages/foam-vscode/src/services/config.ts
riccardoferretti: Sounds good to me, can you please create one and start the conversation there?

packages/foam-cli/src/commands/janitor.ts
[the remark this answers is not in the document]
riccardoferretti: nevermind, I actually just read your second comment :) 
I feel for now it's ok this way (this PR doesn't change anything from the pre-existing behavior).
At some point there will be also a more comprehensive approach to the CLI (probably on our next command, or based on the conversation around foam core/vscode/cli), and I would wait until then to make further changes here

ingalless (approved):
Great work! Agree with your comments about errors

jmg-duarte:
I found a bug. The `FETCH_HEAD` appears when using your branch
image
I've also noticed `settings.json` getting added to the graph but the `FETCH_HEAD` is the most weird since VSCode does not even show `.git`

riccardoferretti:
@jmg-duarte that's really odd, both because `.` directories should automatically be excluded, and because it should only index markdown files, will look into it..

ingalless:
@jmg-duarte could you try the master branch out and see if the problem is there too?

I find it unlikely that my branch will have introduced the issue, but master was merged in recently.

jmg-duarte:
I found the problem with the `settings.json`, whenever I change the `.vscode/settings.json` the change is being propagated and lands on the graph.
I was not able to reproduce this bug on master (due probably to the lack of file sync).

The `FETCH_HEAD` bug is not present in master.

The repository I'm using is https://github.com/rustype/notes

riccardoferretti:
@jmg-duarte oh yeah, great find, thanks!
The problem is that the event listeners are not filtering the files based on the globs, will fix that later.

riccardoferretti:
@jmg-duarte this should be fixed now (I tested manually locally, I have an action item to start building some integration/e2e tests for the extension)

jmg-duarte:
I just tested, it is fix as far as I can tell!
Great job!

PR #350, opened 2020-11-17
Fix icon not found error while vs-code packaging

litanlitudan:
Ran into the following error while doing `yarn package-extension` under folder foam-vscode
image
Turns out that vsce will somehow drop a default .vscodeignore, therefore seems the icon folder outside of foam-vscode is undetectable to vsce.
https://stackoverflow.com/questions/44423212/error-detecting-icon-when-publishing-vscode-extension

This error can be fixed by moving the icon folder into the foam-vscode folder.
image

Not sure whether this is a good practice though, since this icon might be shared with other non-vscode components in the codebase.


jmg-duarte:
> Not sure whether this is a good practice though, since this icon might be shared with other non-vscode components in the codebase.

Originally I didn't put the icon in the extension folder because of this, while it currently isn't, it is useful for people to quickly see the icon in the top level.

litanlitudan:
> > Not sure whether this is a good practice though, since this icon might be shared with other non-vscode components in the codebase.
> 
> Originally I didn't put the icon in the extension folder because of this, while it currently isn't, it is useful for people to quickly see the icon in the top level.
> 
> We could symlink it but it would probably break on Windows (?)

Do you prefer putting the link within the foam-vscode folder or the top of the project folder? Currently, I don't have a windows machine to try out, but I guess this might be a good idea.

riccardoferretti (approved):
I am happy with the icon in the `foam-vscode` folder. this is the only place where we need it now, we can always revisit in the future if necessary.

riccardoferretti:
@all-contributors add @litanlitudan for code

PR #352, opened 2020-11-18
Center graph on active note, plus other tweaks

riccardoferretti:
Selecting a note in the editor will now highlight it in the graph, and center on it.

Also added some UX/style improvements:
- show label for highlighted node regardless of zoom
- show labels earlier (in terms of zoom)
- thinner edges for lighter visual impact of graph
- reduced relative size of nodes

### DIFF
--- a/packages/foam-vscode/src/features/dataviz.ts
+++ b/packages/foam-vscode/src/features/dataviz.ts
@@ -4,6 +4,7 @@ import { FoamFeature } from "../types";
 import { Foam } from "foam-core";
 import { TextDecoder } from "util";
 import { getTitleMaxLength } from "../settings";
+import { isSome } from "../utils";
 
 const feature: FoamFeature = {
   activate: (context: vscode.ExtensionContext, foamPromise: Promise<Foam>) => {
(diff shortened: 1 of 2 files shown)


packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -20,6 +21,17 @@ const feature: FoamFeature = {
         foam.notes.unstable_removeEventListener(onNoteAdded);
       });
 
+      vscode.window.onDidChangeActiveTextEditor(e => {
+        if (e.document.uri.scheme === "file") {
+          const note = foam.notes.getNoteByURI(e.document.uri.fsPath);
+          if (isSome(note)) {
```
ingalless, line 27: Not for this PR, but I'm noticing this isSome is starting to get used heavily. Perhaps we should try make the returns from functions that currently result in an "isSome" a little more explicit (null for nothing, for example). What do you think?

ingalless (approved):
I made a comment, but not relevant to this PR. A broader discussion

packages/foam-vscode/src/features/dataviz.ts
riccardoferretti: IIUC that's what we use it for. e.g. `getNoteByURI` will return `null` if no note was found with that URI.
`isSome` is the way we know that something was returned.
Basically anywhere we would tend to use `!== null` or `!== undefined` we can (and should) use `isSome` instead. The same, in the opposite direction, applies to `isNone`.

Am I understanding your point correctly?

ingalless: Ah, yes you do understand correctly. I misunderstood isSome's usage. Thanks for the explanation!

PR #359, opened 2020-11-20
Templates improvements

### LINKED ISSUE #356: Create new note from template improvements
- [ ] When there are no templates, we should show an appropriate message in (or in place of) the quick pick
- [ ] Simplify the steps by merging directory and file name questions into one ("Enter note path"), include a pre-filled value with the current directory (as it is now) and the filename "new-note.md" and pre-highlight "new-note"
- [ ] pressing ESC at any step should stop the flow of input boxes

from comments in #346 :
- [ ] Setting a daily note template to be the ID of one of the templates
- [ ] A "create new template" command that can also be presented when no templates exist!

ingalless:
Will resolve #356

- [x] When there are no templates, we should show an appropriate message in (or in place of) the quick pick
- [x] Simplify the steps by merging directory and file name questions into one ("Enter note path"), include a pre-filled value with the current directory (as it is now) and the filename "new-note.md" and pre-highlight "new-note"
- [x] pressing ESC at any step should stop the flow of input boxes
- [ ] insert template at cursor position - **Handle in a future PR**

from comments in #346 :
- [ ] Setting a daily note template to be the ID of one of the templates - **Handle in a future PR**
- [x] A "create new template" command that can also be presented when no templates exist!


riccardoferretti (commented):
This looks really good @ingalless!
I have no comments on the code, looks great! I need two more things before merging the PR:
- [ ] rebase off the latest master, which includes a framework for e2e testing in vscode
- [ ] write some tests for the feature (basically based on the specs of the PR). You will need to have rebased to be able to write e2e tests 
- [ ] update and improve the `docs/features/note-templates.md` page to help people create and use templates

This is important both because now that we have the means to test in vscode I want to be more rigorous about testing, and because you will be the first one beside me writing e2e tests in Foam, and I'd love to hear your comments afterwards :) 

Thanks!

edit: I added a third item to the list, updating the documentation :)

.vscode/settings.json
```diff
@@ -20,5 +20,8 @@
     ],
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true,
-    "prettier.requireConfig": true
+    "prettier.requireConfig": true,
+    "[typescript]": {
+        "editor.defaultFormatter": "vscode.typescript-language-features"
```
riccardoferretti, line 25: why this?

packages/foam-vscode/__mocks__/vscode.ts
```diff
@@ -11,6 +12,9 @@ https://github.com/microsoft/vscode-test/issues/37#issuecomment-584744386
 
 const vscode = {
     // Add values and methods as needed for tests
+
+    // Keep env the same, no need to mock this.
+    env: env
```
riccardoferretti, line 17: can you add an explanation for this?

packages/foam-vscode/src/features/create-from-template.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        return Promise.resolve(template)
+      });
+
+      await commands.executeCommand('foam-vscode.create-new-template');
+
+      const file = await workspace.fs.readFile(
+        Uri.file(template)
+      )
+      expect(window.showInputBox).toHaveBeenCalled();
+      expect(file).toBeDefined();
```
riccardoferretti, line 41: what is this checking?

packages/foam-vscode/src/features/create-from-template.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      await workspace.fs.delete(Uri.file(template));
```
riccardoferretti, line 42: can we put the various clean ups in support methods? (see https://jestjs.io/docs/en/setup-teardown)

packages/foam-vscode/src/features/create-from-template.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      window.showInputBox = jest.fn(() => {
+        return Promise.resolve(undefined);
+      })
+
+      await commands.executeCommand('foam-vscode.create-new-template');
+
+      expect(window.showInputBox).toHaveBeenCalled();
+      await expect(workspace.fs.readFile(
+        Uri.file(template)
+      )).rejects.toThrow();
```
riccardoferretti, line 62: why is this rejecting?

packages/foam-vscode/src/test/support/extended-vscode-environment.js
```diff
@@ -8,7 +9,13 @@ class ExtendedVscodeEnvironment extends VscodeEnvironment {
     // Implementation of getWordRangeAtPosition uses "instanceof RegExp" which returns false
     // due to Jest running tests in the different vm context.
     // See https://github.com/nodejs/node-v0.x-archive/issues/1277.
+    // And also https://github.com/microsoft/vscode-test/issues/37#issuecomment-700167820
     this.global.RegExp = RegExp;
+    this.global.vscode = vscode;
+  }
+  async teardown() {
+    this.global.vscode = {};
```
riccardoferretti, line 17: probably it doesn't matter, but should we try to put back whatever was overridden at setup time, if we are going to assign it in any case?

riccardoferretti (commented):
this is really good!
I have left a few comments, to highlight things that are not clear to me on first sight and therefore I would imagine would be hard to understand to someone first looking at the tests, but the direction is really good!

packages/foam-vscode/__mocks__/vscode.ts
ingalless: This has been removed, seems it wasn't necessary after all!

.vscode/settings.json
ingalless: Whoops! That's vscode's doing, will remove

packages/foam-vscode/src/features/create-from-template.spec.ts
ingalless: fs.readFile throws if the template isn't created. I could check the contents of the file if you prefer?

packages/foam-vscode/src/test/support/extended-vscode-environment.js
ingalless: Happy to do it if you like, but tests seem fine with it. Let me know your preference

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -11,54 +11,122 @@ import { FoamFeature } from '../types';
 import { TextEncoder } from 'util';
 import { focusNote } from '../utils';
 
-const templatesDir = `${workspace.workspaceFolders[0].uri.path}/.foam/templates`;
+const templatesDir = `${workspace.workspaceFolders[0].uri.fsPath}/.foam/templates`;
```
riccardoferretti, line 14: why `fsPath` here instead of `path`? Generally speaking we should be very wary of using `fsPath` unless absolutely necessary

edit: Looking at this more closely I would actually have this as a `URI` (see my comment below about `URI` vs `path` vs `fsPath`)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -11,54 +11,122 @@ import { FoamFeature } from '../types';
...
+- $3
+
+To create a note from this template, run the 'Foam: Create new note from template' command.
+`;
 
 async function getTemplates(): Promise<string[]> {
   const templates = await workspace.findFiles('.foam/templates/**.md');
-  // parse title, not whole file!
-  return templates.map(template => path.basename(template.path));
+  return templates.map(template => path.basename(template.fsPath));
```
riccardoferretti, line 30: same as above, can we use `.path` instead?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -11,54 +11,122 @@ import { FoamFeature } from '../types';
...
+async function createNoteFromTemplate(): Promise<void> {
+  const templates = await getTemplates();
+  if (templates.length === 0) {
+    return offerToCreateTemplate();
+  }
+  const activeFile = window.activeTextEditor?.document?.fileName;
+  const currentDir =
+    activeFile !== undefined
+      ? path.dirname(activeFile)
+      : workspace.workspaceFolders[0].uri.fsPath;
```
riccardoferretti, line 53: same here, ideally we would:
- always use `URI`
- if we can't use the `URI`, we use `path`
- if we really can't use `path`, use `fsPath`

if we use `URI` than it will be the `DataStore` to take care of the platform specific behavior.
if we use `fsPath` the rest of that code needs to be platform-safe.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -11,54 +11,122 @@ import { FoamFeature } from '../types';
...
+  );
+  await focusNote(filename, true);
+  await window.activeTextEditor.insertSnippet(snippet);
+}
+
+async function createNewTemplate(): Promise<void> {
+  const isFirstTemplate = (await getTemplates()).length === 0;
+  const defaultFileName = 'new-template.md';
+  const templatesDir = path.join(
+    workspace.workspaceFolders[0].uri.fsPath,
```
riccardoferretti, line 91: this makes sense to me, as we are showing the path to the user, so we want to present it in the format they are expecting it based on their platform 👍  - I would just do it directly in the `showInputBox`, so that it's super clear you only need/access this property there

packages/foam-vscode/src/test/support/extended-vscode-environment.js
riccardoferretti: just to keep things consistent and avoid random unexpected behaviors, it feels to me that we should restore whatever was here before

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -11,54 +11,122 @@ import { FoamFeature } from '../types';
...
+    new TextEncoder().encode('')
+  );
+  await focusNote(filename, true);
+  await window.activeTextEditor.insertSnippet(snippet);
+}
+
+async function createNewTemplate(): Promise<void> {
+  const isFirstTemplate = (await getTemplates()).length === 0;
+  const defaultFileName = 'new-template.md';
+  const templatesDir = path.join(
```
riccardoferretti, line 90: you also defined this `templatesDir` at the top of the file, this feels more like it's the template itself?
Maybe properly name it if it's the file, or reuse `templatesDir` if not.

Also, I would make it a `URI` and get the `fsPath` only in the place where you actually need it (it makes it easier to understand what's going on)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -11,54 +11,122 @@ import { FoamFeature } from '../types';
...
+      templatesDir.length - 3,
+    ],
+    validateInput: value =>
+      value.length ? undefined : 'Please enter a value!',
+  });
+  if (filename === undefined) {
+    return;
+  }
+
+  const template = isFirstTemplate ? firstTemplateContent : '';
```
riccardoferretti, line 110: Have you considered having a more generic template and using it all the times instead of just on first creation? Worst case they would just delete it, and best case it refreshes people's memory when they are creating new templates

riccardoferretti (commented):
thanks @ingalless ! I have left a few comments, let me know what you think.
They are mostly around `URI`, as our "understanding" of when/where to use them (and all the path related issues we have had) have taught us a few lessons :)

packages/foam-vscode/src/features/create-from-template.ts
ingalless: Good idea!

packages/foam-vscode/src/features/create-from-template.ts
ingalless: Just so we're talking about the same `URI`, do you mean a Foam `URI` and not a `vscode.Uri`?

riccardoferretti: Either one, they are fully compatible. but generally speaking let's use the Foam one so that we keep data in our abstractions as much as possible

ingalless:
@riccardoferretti Implemented your PR feedback.
Let me know if there's anything else 👍🏻

riccardoferretti (approved):
looks great!

riccardoferretti:
2 final questions now that the code and tests are in good shape:
- [ ] do we need to update the docs?
- [ ] do we need to update foam-template?

ingalless:
> Do we need to update the docs?

Good idea, just to highlight the extra commands.

> Do we need to update the foam-template

We could ship a template in `.foam/templates` which documents the feature further in-situ?

riccardoferretti:
sounds good:
- let's add the documentation changes in this PR.
- let's make a PR in foam-template for the updates in that repo

ingalless:
@riccardoferretti Documented here. PR landing in `foam-template` shortly!

Issue #358, opened 2020-11-20
Make dataviz graph updates smooth


riccardoferretti:
Currently each time the dataviz graph is refreshed (e.g. because a new note was added to foam), it will restart the simulation.
That can be annoying, especially if it happens frequently (e.g. every time a file is saved).

Because of this the dataviz is currently only refreshed when new notes are added to foam.

- [x] make the graph refresh smooth
- [x] make the graph refresh on every foam update, including file changes (and removals)

Issue #322, opened 2020-11-07
Add theme support in graph


riccardoferretti:
Currently the graph colors are static and hardcoded.
Make them configurable, and ideally seeded by the vscode theme.

riccardoferretti:
the webview example in vscode was updated to include theme info, see here: https://github.com/microsoft/vscode-extension-samples/commit/100b1b20459e2026fe686ef1da0348abf744f4c8#diff-fd51a8de166b021fee2cec8351064a4fdd3ea6c42ebf59da981bd94cb6106e28

PR #360, opened 2020-11-21
Dataviz: graph smooth update and theme

riccardoferretti:
Addresses #358 and #322 

- Refactored `graph.js` to have some sort of design (still lightweight and with a few corners cut). 
- Simplified and consolidated styling, while picking properties from vscode theme
- Completely changed the way the graph data is updated, to support the fact that the simulation requires objects to stay constant. The approach was to make the simulation object as small as possible (it only contains the ID of the nodes), and the rest of the data is looked up in a separate object. On refresh we perform a diff of just the IDs in the graph.
- Added support for multiple node selection in graph by pressing SHIFT key while clicking on nodes

### DIFF
--- a/packages/foam-vscode/src/features/dataviz.ts
+++ b/packages/foam-vscode/src/features/dataviz.ts
@@ -34,7 +34,6 @@ const feature: FoamFeature = {
           }
         }
       });
-      updateGraph(panel, foam);
     });
   }
 };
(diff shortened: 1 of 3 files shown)


ingalless (approved):
I don't fully understand the changes, so a more graph-savvy contributor should probably look over this too!

But all looks good!

PR #377, opened 2020-11-25
Adding service that can direct logging to vscode console

riccardoferretti:
This should make debugging reported issues much easier

image


packages/foam-core/src/services/logger.ts
```diff
@@ -4,4 +4,45 @@ export interface ILogger {
...
+    getLevel: () => currentLogLevel,
+    setLevel: level => {
+      currentLogLevel = level;
+    },
+  };
+};
+
+export const consoleLogger: ILogger = createLogger(
+  (level, message, ...params) => {
+    if (level != 'off') {
```
ingalless, line 44: can we do `level.toLower() !== 'off'` here to clear the warnings?

ingalless (commented):
Just one small suggestion! Great work Riccardo, this will really improve our bug reports

packages/foam-core/src/services/logger.ts
riccardoferretti: thx for spotting the `!=`, I fixed it now!

riccardoferretti:
@ingalless @jmg-duarte major changes:
- I went the OO way to model the `Logger`, I think it fits better with the use case (the code is in the end more readable)
- I am now using a global `Logger` object, to avoid passing that around (consider we might want to monitor low level functions, and passing the logger around like that gets out of control)

riccardoferretti:
@ingalless @jmg-duarte I have merged this to unblock the release, but feel free to still look at the code and leave comments, they can be addressed in a separate PR

packages/foam-core/src/utils/log.ts
```diff
@@ -0,0 +1,89 @@
+export interface ILogger {
+  debug(message?: any, ...params: any[]): void;
+  info(message?: any, ...params: any[]): void;
+  warn(message?: any, ...params: any[]): void;
+  error(message?: any, ...params: any[]): void;
+  getLevel(): LogLevelThreshold;
+  setLevel(level: LogLevelThreshold): void;
+}
+
+export type LogLevel = 'debug' | 'info' | 'warn' | 'error';
```
jmg-duarte, line 10: I think this makes more sense as an `enum`, as well as `LogLevelThreshold` and `severity`

packages/foam-core/src/utils/log.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+  getLevel(): LogLevelThreshold {
+    return this.level;
+  }
+  setLevel(level: LogLevelThreshold): void {
+    this.level = level;
+  }
+}
+
+export class ConsoleLogger extends BaseLogger {
```
jmg-duarte, line 54: This and the remaining classes could be isolated into their own file

packages/foam-core/src/utils/log.ts
```diff
@@ -0,0 +1,89 @@
...
+  log(level: LogLevel, msg?: string, ...params: any[]): void {
+    console[level](`[${level}] ${msg}`, ...params);
+  }
+}
+
+export class NoOpLogger extends BaseLogger {
+  log(_l: LogLevel, _m?: string, ..._p: any[]): void {}
+}
+
+export class Logger {
```
jmg-duarte, line 64: I feel that the setter in this class should only work once and before any logging is done, but I'm not familiar enough with TS to know how to do it

packages/foam-core/src/utils/log.ts
ingalless: They could be, but personally I don't mind them being in the same file. All have the same responsibility and relation, so the colocation is nice here 👍🏻

jmg-duarte: The motivation comes from file system navigation, it becomes easier than grepping the whole project

packages/foam-core/src/utils/log.ts
riccardoferretti: I haven't found a reliable way to execute the logging init code before everything else, doing this in JS is really hard.
Happy to go into detail if you are interested as to why.

For this reason, I went for relaxing that constraint (I do agree with you that it would be the better way to enforce consistency).
The other implication was that I couldn't create named loggers, which I liked the idea of but were could still create a stale state (I actually lost a good hour the other day because the default export got me there!).

Interestingly enough I haven't even seen established logging libraries for JS address this 🤷‍♂️

packages/foam-core/src/utils/log.ts
riccardoferretti: I think it's also a matter of convention. In Java-land what @jmg-duarte says is more the norm, in JS-land the norm is what @ingalless says.

What I like about JS is that you can have `utils/log.ts`, and as the file becomes too big turn it into a module `utils/log/index.ts`, which then allows to have more files in the directory

packages/foam-core/src/utils/log.ts
riccardoferretti: it started as an enum, the reason why I turned it into a union type is because as I was using them, having enums was making the code clunkier.
Per https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant for the usage we make of log levels/thresholds, the two seemed equivalent to me, but happy to reconsider.

What's your thinking?

packages/foam-core/src/utils/log.ts
ingalless: You could have a boolean property on the class, `hasLogged` or something to that effect that is set to true on each of the method calls

Now you can guard the `setDefaultLogger` method with this and throw. Does that work?

riccardoferretti: the problem is not detecting the stale state, it is being able to do something about it :)

jmg-duarte: The boolean idea could kinda work but I guess it's "too much" work and we can just expect the developer to not go around and constantly change it

packages/foam-core/src/utils/log.ts
jmg-duarte: The strings as types approach seems error prone while `Enum.Something` seems better.
If the types however are just compile-time this approach is better

packages/foam-core/src/utils/log.ts
riccardoferretti: oh yes, we wouldn't pass around to plugins the global `Logger` object, but an instance of `ILogger` interface.
only core foam code can call this function, and that is really meant to only be called at bootstrap time, I can't see any other circumstance, so it should be easy to spot a mistake there..

PR #382, opened 2020-11-26
Fixed parsing of tags

riccardoferretti:
fixes problem referred to in #302 (see when reopened)

### DIFF
--- a/packages/foam-core/src/utils/hashtags.ts
+++ b/packages/foam-core/src/utils/hashtags.ts
@@ -6,9 +6,9 @@ const WORD_REGEX = /(^|[ ])([\w_-]+\b)/gm;
 export const extractHashtags = (text: string): Set<string> => {
   return isSome(text)
     ? new Set(
-        Array.from([...text.matchAll(HASHTAG_REGEX)])
-          .map(m => m[2].trim())
-          .filter(tag => !isNumeric(tag))
+        Array.from(text.matchAll(HASHTAG_REGEX), m => m[2].trim()).filter(
+          tag => !isNumeric(tag)
+        )
       )
     : new Set();
 };
--- a/packages/foam-core/src/utils/hashtags.ts
+++ b/packages/foam-core/src/utils/hashtags.ts
@@ -1,15 +1,10 @@
 import { isSome, isNumeric } from './core';
-
-const HASHTAG_REGEX = /(^|[ ])#([\w_-]+\b)/gm;
-const WORD_REGEX = /(^|[ ])([\w_-]+\b)/gm;
+const HASHTAG_REGEX = /(^|[ ])#([\w_-]*[a-zA-Z][\w_-]*\b)/gm;
+const WORD_REGEX = /(^|[ ])([\w_-]*[a-zA-Z][\w_-]*\b)/gm;
 
 export const extractHashtags = (text: string): Set<string> => {
   return isSome(text)
-    ? new Set(
-        Array.from(text.matchAll(HASHTAG_REGEX), m => m[2].trim()).filter(
-          tag => !isNumeric(tag)
-        )
-      )
+    ? new Set(Array.from(text.matchAll(HASHTAG_REGEX), m => m[2].trim()))
     : new Set();
 };
 


packages/foam-core/src/utils/hashtags.ts
```diff
@@ -6,9 +6,9 @@ const WORD_REGEX = /(^|[ ])([\w_-]+\b)/gm;
 export const extractHashtags = (text: string): Set<string> => {
   return isSome(text)
     ? new Set(
-        Array.from([...text.matchAll(HASHTAG_REGEX)])
-          .map(m => m[2].trim())
-          .filter(tag => !isNumeric(tag))
+        Array.from(text.matchAll(HASHTAG_REGEX), m => m[2].trim()).filter(
+          tag => !isNumeric(tag)
```
jmg-duarte, line 10: Is this necessary? The \w should not match numbers

jmg-duarte, line 10: I think the regex could be more restrictive as to avoid this check

jmg-duarte (changes requested):

packages/foam-core/src/utils/hashtags.ts
jmg-duarte: This should work `const HASHTAG_REGEX = /(^|[ ])#([a-zA-Z][\w_-]+\b)/gm;`

riccardoferretti: I agree that more regex magic could do the trick, but after spending so much time on it I got pragmatic (tired) and settled here.

This is not the best tag parsing code out there, I have seen something by twitter https://github.com/twitter/twitter-text/ but it's much more (I have also pulled out their tag code, it just feels a bit heavier and doesn't quite match the tags cases we want).
I have not been able to find a simple tag extraction library/gist/file out there, would love pointers if people are aware of some.

jmg-duarte: The addition I made is simple enough to work in "every" case, you just force the first character to be a letter

riccardoferretti: thing is, it can start with a number, it just can't be all numbers:

```
      62 |     expect(
      63 |       extractHashtags('this #123 tag should be ignore, but not #123four')
    > 64 |     ).toEqual(new Set(['123four']));
```

jmg-duarte: Hm then this should work: `(^|[ ])#([\w_-]*[a-zA-Z][\w_-]*\b)`
It ensures there is at least one alphabetic char before or after the 0 or more `\w` matches.
https://regex101.com/r/JEqxC9/1

riccardoferretti: that works, great!

bserrao: Why a tag can't be all numbers? For instance if i want a tag to represent the year... like #2020

Cheers

jmg-duarte: @riccardoferretti what is your opinion? IMO it makes sense, maybe we should yank some code out of the Twitter extension, or if necessary I can come up with a regex

riccardoferretti: yeah, I think in the end it does make sense, and creates a more consistent behavior.
Thanks for volunteering to take care of this @jmg-duarte, happy to go for a regex

bserrao: Any update on this? Did you find a way for having numeric tags?

Thanks

jmg-duarte: This was merged, have you tried with the latest version?

bserrao: Yes, using v0.77. The code is a bit different know.. But i don't know if i'm doing something wrong...

```
export const extractHashtags = (text: string): Set<string> => {
  return isSome(text)
    ? new Set(Array.from(text.matchAll(HASHTAG_REGEX), m => m[2].trim()))
    : new Set();
};
```
 Vscode highlight it as a tag but it's not recognized as one in tag explorer. Maybe the problem is with tag explorer itself?

jmg-duarte: @riccardoferretti do you have any idea what this might be?

jmg-duarte: In the meantime, please submit an issue @bserrao

riccardoferretti: interesting.. I reproduced it as well, that's odd. yes let's start by filing an issue

PR #386, opened 2020-11-27
WIP: Fix wikilinks slug querying

SanketDG:
This attempts to fix the wikilink generation issues.

### DIFF
--- a/packages/foam-core/src/note-graph.ts
+++ b/packages/foam-core/src/note-graph.ts
@@ -1,6 +1,6 @@
 import { Graph } from 'graphlib';
 import { URI, ID, Note, NoteLink } from './types';
-import { computeRelativeURI } from './utils';
+import { computeRelativeURI, nameToSlug } from './utils';
 import { Event, Emitter } from './common/event';
 
 export type GraphNote = Note & {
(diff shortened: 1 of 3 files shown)


packages/foam-core/src/note-graph.ts
```diff
@@ -91,7 +91,7 @@ export class NoteGraph implements NoteGraphAPI {
     // prettier-ignore
     const filterFn =
       query == null ? (note: Note | null) => note != null
-        : 'slug' in query ? (note: Note | null) => note?.slug === query.slug
+        : 'slug' in query ? (note: Note | null) => [nameToSlug(query.slug), query.slug].includes(note?.slug as string)
```
SanketDG, line 94: I am checking for both [[Slug Name]] and [[slug-name]] here, however they are defined in the note, so both can be used interchangably.

SanketDG: @riccardoferretti

riccardoferretti: @SanketDG I have a feeling this wouldn't work with the dataviz graph, have you tested it?

SanketDG: Yes it definitely works with my notes, which all have filenames like "javascript-execution.md" and referenced as [[Javascript Execution]], this is uniform throughout.

Any other cases you want me to check?

riccardoferretti (approved):
LGTM!

PR #394, opened 2020-12-02
fix(dataviz): Don't delete node for note on update - #393

### LINKED ISSUE #393: Saving a file removes the graph edge in foam graph.
- Foam version: `0.7.2`
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes | No

**Summary**
Saving a file removes incoming graph edges

**Steps to reproduce**
1. create a new `[[link]]` in a file `origin.md`
2. cmd-click to go to new file `link.md`
3. save new file
Result: the link between `origin` and `link` in the foam graph disappears.

AndrewNatoli:
Will resolve #393 

Looks like as part of `0.7.2` a `deleteNote` call was added to `setNote`. ([View changes](https://github.com/foambubble/foam/commit/26ab27e06fb7747bb58b39c311ff54c674b70f52#diff-139497fe1094e55b5ad85d381b423b985ba4040dd5adbdf75ac89adf934c99a9R62))

Doing so causes graphlib to delete the inbound links from other nodes so when this node is re-created it will not have its inbound links. ([View relevant graphlib source](https://github.com/dagrejs/graphlib/blob/master/lib/graph.js#L456))

I switched the `doDelete` back to `getNote` so it still satisfies finding if the note already existed for the sake of triggering the onUpdate / onAdd events-- it seems to be behaving as expected.

That said, I noticed the _original_ approach was to delete the outbound links for the note so they can be re-created. That looked like this:
```
if (noteExists) {	
      (this.graph.outEdges(id) || []).forEach(edge => {	
        this.graph.removeEdge(edge);	
      });	
    }
```

In that previous code, the actual node for the note was not being deleted nor its inbound references. 

I can update this PR to restore that code if it seems necessary. Once I have feedback on that I'll check on the tests.

- [x] Awaiting initial feedback
- [x] Tests passing in `foam-core`
- [x] Tests passing in `foam-vscode`

### DIFF
--- a/packages/foam-core/src/note-graph.ts
+++ b/packages/foam-core/src/note-graph.ts
@@ -59,7 +59,7 @@ export class NoteGraph implements NoteGraphAPI {
 
   public setNote(note: Note): GraphNote {
     const id = this.createIdFromURI(note.source.uri);
-    const oldNote = this.doDelete(id, false);
+    const oldNote = this.getNote(id);
     const graphNote: GraphNote = {
       ...note,
       id: id,
--- a/packages/foam-core/src/note-graph.ts
+++ b/packages/foam-core/src/note-graph.ts
@@ -60,6 +60,9 @@ export class NoteGraph implements NoteGraphAPI {
   public setNote(note: Note): GraphNote {
     const id = this.createIdFromURI(note.source.uri);
     const oldNote = this.getNote(id);
+    if (isSome(oldNote)) {
+      this.removeForwardLinks(id);
+    }
     const graphNote: GraphNote = {
       ...note,
       id: id,
@@ -91,9 +94,7 @@ export class NoteGraph implements NoteGraphAPI {
     const note = this.getNote(noteId);
     if (isSome(note)) {
       this.graph.removeNode(noteId);
-      (this.graph.outEdges(noteId) || []).forEach(edge => {
-        this.graph.removeEdge(edge);
-      });
+      this.removeForwardLinks(noteId);
       fireEvent && this.onDidDeleteEmitter.fire(note);
     }
     return note;
@@ -131,6 +132,12 @@ export class NoteGraph implements NoteGraphAPI {
     return (this.graph.outEdges(noteId) || []).map(edge =>
       this.graph.edge(edge.v, edge.w)
     );
+  }
+
+  public removeForwardLinks(noteId: ID) {
+    (this.graph.outEdges(noteId) || []).forEach(edge => {
+      this.graph.removeEdge(edge);
+    });
   }
 
   public getBacklinks(noteId: ID): GraphConnection[] {
--- a/packages/foam-core/src/note-graph.ts
+++ b/packages/foam-core/src/note-graph.ts
@@ -93,8 +93,11 @@ export class NoteGraph implements NoteGraphAPI {
   private doDelete(noteId: ID, fireEvent: boolean): GraphNote | null {
     const note = this.getNote(noteId);
     if (isSome(note)) {
-      this.graph.removeNode(noteId);
-      this.removeForwardLinks(noteId);
+      if (this.getBacklinks(noteId).length >= 1) {
+        this.graph.setNode(noteId, null); // Changes node to the "no file" style
+      } else {
+        this.graph.removeNode(noteId);
+      }
       fireEvent && this.onDidDeleteEmitter.fire(note);
     }
     return note;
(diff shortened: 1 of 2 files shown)


riccardoferretti:
hi @AndrewNatoli, thanks for looking into this and spotting the issue with the inbound links.
The reason why the tests are failing because if we don't remove the outbound links, they will linger even after the note has been updated.
From this it will probably make sense to, in `doDelete`, set the note to `null` or `undefined` instead of removing it - this way graphlib will not eagerly remove the inbound links. 
Also, we need to add a test case for the bug this PR is meant to fix.

AndrewNatoli:
> The reason why the tests are failing because if we don't remove the outbound links, they will linger even after the note has been updated.

Aha, I see that now. I'll try and get back to this a bit later today, thanks for confirming how that process works!

AndrewNatoli:
How would you feel about lifting this:
```
(this.graph.outEdges(noteId) || []).forEach(edge => {
        this.graph.removeEdge(edge);
      });
```
out of `doDelete` and into `removeForwardLinks`? Figure the name lines up with `getForwardLinks`.

In `setNote`, `removeForwardLinks` would be called if the note exists and we wouldn't be mixing different actions in `doDelete`.

riccardoferretti:
happy to go with that 👍

packages/foam-core/src/note-graph.ts
```diff
@@ -91,9 +94,7 @@ export class NoteGraph implements NoteGraphAPI {
     const note = this.getNote(noteId);
     if (isSome(note)) {
       this.graph.removeNode(noteId);
```
riccardoferretti, line 96: this would cause the inbound/outbound links to be removed, no? 
I reckon we would need to set it to `null` to keep the existing inbound connections

AndrewNatoli: It looks like `doDelete` is now only being called when a note is deleted which I think means this shouldn't be a problem anymore.
image

If there's a concern about a deletion rippling to connections between neighboring files, those seem fine.

Screen Shot 2020-12-02 at 11 51 57 AM

Here if I delete "types of food" the original connections in "soup" remain
Screen Shot 2020-12-02 at 11 52 26 AM

riccardoferretti: if `pasta` has a connection to `types-of-food`, I would expect to see that connection even afterwards, but instead of being associated to a real note, it would be associated to a placeholder note

AndrewNatoli: Ah I see. 

So... On delete, if a note has inbound links, do not entirely remove the node just set it back to a placeholder. 

This seems to be an existing problem with the current release. 

Mind if I add that to the issue tracker and try to address that separately? It doesn't look like it'll overlap with this change to update events.

I'm also having some recurring issues running the extension in dev mode that are making it a little tricky, I'll ask about those in Discord later and see if it's anything I can update the docs for as well.

riccardoferretti: I understand that your specific issue was with the removal of the inbound links, which you have now solved.
I see this issue as fixing a problem with the delete function, which has not been solved yet. 
Is there a reason why you'd rather not make the change here?

Also, it's important to add the test cases for:
- [ ] the bug that was fixed
- [ ] the current incorrect behavior (the fact that tests are not failing now means we are not capturing this case)

I am also happy to contribute to the branch, for me what matters is merging a complete fix to the deletion bug.

riccardoferretti:
hi @AndrewNatoli, thanks again for taking the initiative to fix this bug.

Let me know if my message on Discord helped you with the dev issues. The problem is caused by a mixture of commonjs and non-commonjs modules, so the fixes in `tsconfig.json` should do it. Having done various tests I am pretty comfortable that's the issue, but it was also sporadic before on my machine, so having a second report would be helpful.
Let me know, and if it works for you we can add it to this PR.

Also, let me know if you are happy to implement the changes I have requested, or you'd rather me doing it - whether on this branch or not (I will not publish a new release until all those issues are fixed though, so functionally it's the same).

I don't want to step on your work, but I understand you might be busy and I am keen to release a new version with the fix - let me know how you would like to proceed. Thanks!

AndrewNatoli:
Hi @riccardoferretti  - I was able to add a test for the new code which catches the regression with the previous code. 

I'll try and revise the `doDelete` method later today (early afternoon here right now) -- I might have some time to hit it. If you need to guarantee it getting done I'm certainly open to you taking it, but I don't want to throw that on you unnecessarily.

AndrewNatoli:
This should be ready for a review now :D

riccardoferretti (approved):
Looks great!

PR #407, opened 2020-12-17
Fix daily note path by using normalized path

### LINKED ISSUE #405: ENOENT: mkdir on Windows
The "Open Daily Note" command seems to be trying to run mkdir on Windows.

MCluck90:
`Uri.path` returns an OS-dependent path. On Windows, this was coming back with a `/` at the beginning which, when resolved, would converts paths like:

```
/c:/path/to/file.md
```

to

```
/C:/c:/path/to/file.md
```

`Uri.fsPath` returns an OS-independent path which does not add an additional slash and works correctly with the `path` module.

Fixes #405


riccardoferretti (approved):
LGTM thanks for fixing this!

riccardoferretti:
@allcontributors add @MCluck90 for code

PR #416, opened 2020-12-28
Remove extension in dated snippets

MCluck90:
Fixes a bug where changing `foam.edit.linkReferenceDefinitions` would not change the way that dated snippets expand.

## Repro

1. Assume `"foam.edit.linkReferenceDefinitions": "withExtensions"`
2. `/today` expands to `[[2020-12-28.md]]`
3. Change to `"foam.edit.linkReferenceDefinitions": "withoutExtensions"`

## Expected

`/today` expands to `[[2020-12-28]]`

## Actual

`/today` still expands to `[[2020-12-28.md]]`

This will be true until the extension is restarted.

## Changes from PR

Fixes this problem by pulling from the config every time a daily note link is generated

### DIFF
--- a/packages/foam-vscode/src/features/open-dated-note.ts
+++ b/packages/foam-vscode/src/features/open-dated-note.ts
@@ -14,7 +14,6 @@ import {
   openDailyNoteFor,
   getDailyNotePath
 } from "../dated-notes";
-import { LinkReferenceDefinitionsSetting } from "../settings";
 import { FoamFeature } from "../types";
 
 interface DateSnippet {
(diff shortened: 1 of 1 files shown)


riccardoferretti:
thanks for the fix @MCluck90 - I actually wonder whether the date wikilink should ever include the extension.
That is just a wikilink, even if it represents a "special note". When wikilink definitions are generated, any daily note link will be included, just like any other link.

@ingalless - what do you think?

ingalless:
@riccardoferretti I don't feel strongly either way! I see why the settings may seem "linked", and I'm guessing that's the interpretation I made?

Perhaps this is a good candidate for a separate config, although then we are adding a lot of configurability (but that's not necessarily bad depending on Foam's goals)

riccardoferretti:
ok, in this case I would just leave them as simple wikilinks.
The setting property will only be used during the generation of the wikilink definition. Basically these wikilinks will be treated as any other wikilink.

@MCluck90 does that make sense?

MCluck90:
@riccardoferretti Yup, sounds good. I'll update this PR to just remove the extension from these snippets altogether

riccardoferretti (approved):

riccardoferretti:
@allcontributors add @MCluck90 for code

PR #417, opened 2020-12-28
Fallback to word-based suggestions when not triggered by trigger character

### LINKED ISSUE #415: Foam disables word based suggestions in markdown files
- Foam version: 0.7.6
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**

It seems having the Foam vscode extension installed disables or breaks word based suggestions in markdown files even with `"editor.wordBasedSuggestions": true`.

I verified with a fresh install of vscode with only the foam extension installed and when the extension is enabled I don't get any word based suggestions but as soon as I disabled the extension and reload vscode then the suggestions are working again. 

I also noticed that suggestions work if the extension is installed but I'm not working in a directory with a `foam.json`

Is this intentional? If not, I'd be willing to investigate if someone could point me in the direction of where to start looking in the codebase.

bpugh:
Fixes #415 

So this is somewhat of an open issue with vscode discussed here: https://github.com/Microsoft/vscode/issues/21611

The TLDR is that currently you have to return undefined or an empty array for suggestions in order for vscode to fallback to word-based suggestions, which is what other markdown extensions I looked at do.

Since all of our snippets start with `/` I assume our intent is to primarily show them as suggestions when the user starts typing a `/` (or `+` for the computed snippets). Since this is handled with the trigger characters we specify, a simple solution for us is to only return suggestions if triggered by trigger character instead of during normal typing of words.

This change means that if a user starts typing `/yes` they will see the suggestion for the `/yesterday` snippet but not if they just `yes` or `yesterday`.

Let me know your thoughts.

### DIFF
--- a/packages/foam-vscode/src/features/open-dated-note.ts
+++ b/packages/foam-vscode/src/features/open-dated-note.ts
@@ -6,7 +6,8 @@ import {
   CompletionItemProvider,
   CompletionItem,
   CompletionItemKind,
-  CompletionList
+  CompletionList,
+  CompletionTriggerKind
 } from "vscode";
 import {
   createDailyNoteIfNotExists,
@@ -171,6 +172,12 @@ const computedSnippets: ((number: number) => DateSnippet)[] = [
 
 const completions: CompletionItemProvider = {
   provideCompletionItems: (_document, _position, _token, _context) => {
+    if (_context.triggerKind === CompletionTriggerKind.Invoke) {
+      // if completion was triggered without trigger character then we return [] to fallback
+      // to vscode word-based suggestions (see https://github.com/foambubble/foam/pull/417)
+      return [];
+    }
+
     const completionItems = [
       ...snippets.map(item => createCompletionItem(item())),
       ...generateDayOfWeekSnippets().map(item => createCompletionItem(item))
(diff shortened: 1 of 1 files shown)


riccardoferretti:
hi @bpugh, I like your approach and reasoning, and yes, foam suggestions should only occur after the trigger characters.

What is the test `_context.triggerKind === CompletionTriggerKind.Invoke` checking?
I want to make it easy for the next person who sees that code to understand what it does :)

bpugh:
Great point! I added comments to code to clarify what's going on. Let me know if that's what you had in mind.

Also somewhat related, I couldn't find any mention of the snippets in the documentation. Is there a ticket to add it?

packages/foam-vscode/src/features/open-dated-note.ts
```diff
@@ -171,6 +172,12 @@ const computedSnippets: ((number: number) => DateSnippet)[] = [
 
 const completions: CompletionItemProvider = {
   provideCompletionItems: (_document, _position, _token, _context) => {
+    if (_context.triggerKind === CompletionTriggerKind.Invoke) {
+      // if completion was triggered without trigger character then we return [] to fallback
+      // to vscode word-based suggestions
```
riccardoferretti, line 177:
```suggestion
      // to vscode word-based suggestions (see https://github.com/foambubble/foam/pull/417)
```

packages/foam-vscode/src/features/open-dated-note.ts
```diff
@@ -181,6 +188,12 @@ const completions: CompletionItemProvider = {
 
 const computedCompletions: CompletionItemProvider = {
   provideCompletionItems: (document, position, _token, _context) => {
+    if (_context.triggerKind === CompletionTriggerKind.Invoke) {
+      // if completion was triggered without trigger character then we return [] to fallback
+      // to vscode word-based suggestions
```
riccardoferretti, line 193:
```suggestion
      // to vscode word-based suggestions (see https://github.com/foambubble/foam/pull/417)
```

riccardoferretti (approved):
Looks good to me. Just added a reference to this PR in the code to quickly give people extra context

bpugh:
Sounds good, I made the changes.

riccardoferretti:
@allcontributors add @bpugh for code

Issue #436, opened 2021-01-07
Customizable graph style


jing-bi:
It would be great that we can add a css file to modify the graph style.
The changeable style might be:
- color of the node
- thickness of the link
- highlight style when hovering the node
- force between nodes

jmg-duarte:
This has been in the works, but the graph and its place in the project has been subject to discussion

riccardoferretti:
To clarify @jmg-duarte, although we are debating how to organize the graph code, that doesn't stop us from implementing new features to the graph in the meantime.
I am actually very keen to add custom styling ASAP, there are a couple of considerations (where does the style sheet live, which properties to expose, ..) but nothing that is impacted by where the graph code comes from.

At the moment these are the style properties available:
- background-color
- font-size
- highlight-color
- node-color by node type
  - note
  - placeholder
  - unknown (I don't think this one is used)

I would expect in the future:
- color for tags (currently we don't show tags)
- color for other node types (based on e.g. a `type` property on the node)
- color for attachments (currently we don't show attachments)

jmg-duarte:
I'm interested in addressing this as I think some aspects of the graph are currently _not optimal_.
I think it makes sense we take this on a step by step basis instead of trying to fit every possible feature into a single PR.

# Style File

## Format
### JavaScript
@riccardoferretti proposed to use `.js`, however I dislike the idea since it opens the possibility for code execution within what was supposed to be a style file

### CSS
CSS is usually the default for this kind of things, however, the styling we require is not more complex than some colors, and sizes.
It ends up being unnecessarily verbose and complex for new users.

### JSON
I my opinion, this is probably the most practical solution, we can simply load the file as an object, filter the required fields and fill in the missing ones.
It is also structured, exposing the hierarchy in styles, and being somewhat reflective of WYSIWYG/more intuitive since the hierarchy is laid out in front of the user.

## Location
Regarding the file location, I am not opposed to adding a `.foam` folder moving forward and deprecating the `.vscode/foam.json` file, it seems to be an overall better solution.

However, I suggest that the file placement be customizable, enabling the user to set the path to the file (or maybe even the folder). This would enable a global Foam config moving forward. (AFAIK `foam-core` is already moving in this direction)

## Naming
If the file path is customizable, this is a non issue. Regardless I propose `graph.json` as the default name.

## Structure 
The current structure is as follows:
```js
const style = {
  background: "",
  fontSize: 10,
  highlightedForeground: "",
  node: {
    note: "",
[...]

PR #438, opened 2021-01-11
Add custom graph styling support 

jmg-duarte:
This PR addresses #436 

The implementation makes use of the webview communication mechanism to
switch messages between the webview and the graph.
It works as follows:

- When the webview is loaded, it now sends a single request to VSCode,
the request asks VSCode for the graph style
- When VSCode answers with the style, the graph style is updated and
the webview loading process continues as normal.

The style change does not modify the API, in fact it makes use of the
shadiest programming tatic ever, *global variables* to remain
compatible.

The style object *currently* supports four base fields:
- background: string
- fontSize: int
- highlightedForeground: string
- node: object
  - note: string
  - nonExistingNote: string
  - unknown: string

For an *ugly* example, see below:
image

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -156,6 +156,20 @@
           "type": "number",
           "default": 24,
           "description": "The maximum title length before being abbreviated. Set to 0 or less to disable."
+        },
+        "foam.graph.customStyle": {
+          "type": "object",
+          "description": "Custom graph styling settings.",
+          "default": {
+            "background": "#202020",
+            "fontSize": 12,
+            "highlightedForeground": "#f9c74f",
+            "node": {
+              "note": "#277da1",
+              "nonExistingNote": "#545454",
+              "unknown": "#f94144"
+            }
+          }
         }
       }
     },
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -160,16 +160,7 @@
         "foam.graph.customStyle": {
           "type": "object",
           "description": "Custom graph styling settings.",
-          "default": {
-            "background": "#202020",
-            "fontSize": 12,
-            "highlightedForeground": "#f9c74f",
-            "node": {
-              "note": "#277da1",
-              "nonExistingNote": "#545454",
-              "unknown": "#f94144"
-            }
-          }
+          "default": {}
         }
       }
     },
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -159,7 +159,7 @@
         },
         "foam.graph.style": {
           "type": "object",
-          "description": "Graph styling settings.",
+          "description": "Custom graph styling settings. An example is present in the documentation.",
           "default": {}
         }
       }
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -157,9 +157,9 @@
           "default": 24,
           "description": "The maximum title length before being abbreviated. Set to 0 or less to disable."
         },
-        "foam.graph.customStyle": {
+        "foam.graph.style": {
           "type": "object",
-          "description": "Custom graph styling settings.",
+          "description": "Graph styling settings.",
           "default": {}
         }
       }
--- a/packages/foam-vscode/src/features/dataviz.ts
+++ b/packages/foam-vscode/src/features/dataviz.ts
@@ -7,11 +7,11 @@ import { getGraphStyle, getTitleMaxLength } from "../settings";
 import { isSome } from "../utils";
 
 const feature: FoamFeature = {
-  activate: async (context: vscode.ExtensionContext, foamPromise: Promise<Foam>) => {
-    const foam = await foamPromise;
-    const panel = await createGraphPanel(foam, context);
+  activate: (context: vscode.ExtensionContext, foamPromise: Promise<Foam>) => {
 
     vscode.workspace.onDidChangeConfiguration(async event => {
+      const foam = await foamPromise;
+      const panel = await createGraphPanel(foam, context);
       if (event.affectsConfiguration('foam.graph.style')) {
         const style = getGraphStyle();
         panel.webview.postMessage({
@@ -22,6 +22,8 @@ const feature: FoamFeature = {
     });
 
     vscode.commands.registerCommand("foam-vscode.show-graph", async () => {
+      const foam = await foamPromise;
+      const panel = await createGraphPanel(foam, context);
       const onFoamChanged = _ => {
         updateGraph(panel, foam);
       };
--- a/packages/foam-vscode/src/features/dataviz.ts
+++ b/packages/foam-vscode/src/features/dataviz.ts
@@ -112,7 +112,6 @@ async function createGraphPanel(foam: Foam, context: vscode.ExtensionContext) {
       switch (message.type) {
         case "webviewStyleRequest":
           const style = getGraphStyle();
-          console.log(style)
           panel.webview.postMessage({
             type: "didUpdateStyle",
             payload: style,
--- a/packages/foam-vscode/src/features/dataviz.ts
+++ b/packages/foam-vscode/src/features/dataviz.ts
@@ -77,7 +77,7 @@ function generateGraphData(foam: Foam) {
       if (!(link.to.path in graph.nodes)) {
         graph.nodes[link.to.path] = {
           id: link.to,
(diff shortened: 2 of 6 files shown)


packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -264,12 +300,12 @@ try {
   const vscode = acquireVsCodeApi();
 
   window.onload = () => {
-    initDataviz(vscode);
-    console.log("ready");
+    console.log("post webviewStyleRequest");
     vscode.postMessage({
-      type: "webviewDidLoad"
```
riccardoferretti, line 270: mmm... here I would actually keep the `webviewDidLoad` event, which has a specific semantic (webview is ready, give me what I need, whether it's data, style, ..)

packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -110,6 +110,15 @@ async function createGraphPanel(foam: Foam, context: vscode.ExtensionContext) {
   panel.webview.onDidReceiveMessage(
     async message => {
       switch (message.type) {
+        case "webviewStyleRequest":
+          const style = getGraphStyle();
+          console.log(style)
+          panel.webview.postMessage({
+            type: "graphStyle",
```
riccardoferretti, line 117: it's not documented, but the naming convention for these events follows the VsCode events naming pattern.
In this case it would be something like `didUpdateStyle`

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -264,12 +300,12 @@ try {
   const vscode = acquireVsCodeApi();
 
   window.onload = () => {
-    initDataviz(vscode);
-    console.log("ready");
+    console.log("post webviewStyleRequest");
```
riccardoferretti, line 303: I would keep it `ready` here? I am not sure how this is `post webviewStyleRequest`

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -300,6 +336,15 @@ try {
           Actions.selectNode(noteId);
         }
         break;
+      case "graphStyle":
+        const style = message.payload;
+        updateStyle(style);
```
riccardoferretti, line 341: mmm... I think here you need a slightly different approach, to be in line with the other actions (and I think it will also simplify the code).

Because style is now part of the model (that is, it can be changed by the app), I would:
- add it to the model itself (and the style code in `initDataviz` needs to reference the model, not the `style` global)
- make `updateStyle` an action, and in its implementation make the changes to the style in the model 

If the design doesn't make sense to you let's chat.

packages/foam-vscode/package.json
```diff
@@ -156,6 +156,11 @@
           "type": "number",
           "default": 24,
           "description": "The maximum title length before being abbreviated. Set to 0 or less to disable."
+        },
+        "foam.graph.customStyle": {
```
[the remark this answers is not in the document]
jmg-duarte: Agreed

packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -110,6 +110,15 @@ async function createGraphPanel(foam: Foam, context: vscode.ExtensionContext) {
   panel.webview.onDidReceiveMessage(
     async message => {
       switch (message.type) {
+        case "webviewStyleRequest":
+          const style = getGraphStyle();
+          console.log(style)
```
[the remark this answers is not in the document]
jmg-duarte: What specifically? The log?

packages/foam-vscode/src/features/dataviz.ts
jmg-duarte: Will update

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -1,29 +1,65 @@
...
+function getStyle(name) {
+  return getComputedStyle(document.documentElement).getPropertyValue(name);
+}
+
+function updateStyle(newStyle) {
+  if (!newStyle) {
+    return;
+  }
+  let node = style.node;
+  if (newStyle.node !== undefined
```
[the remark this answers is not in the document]
jmg-duarte: Indeed, I wasn't aware of that just the `??`

packages/foam-vscode/static/graphs/default/graph.js
jmg-duarte: This was debug code that I forgot to remove, it got switched around and ended like this. My bad

packages/foam-vscode/static/graphs/default/graph.js
jmg-duarte: Hm I see, I thought of it in another way. Load everything else first, the webview last

packages/foam-vscode/static/graphs/default/graph.js
jmg-duarte: By model, do you mean this?
https://github.com/foambubble/foam/blob/32f40fa0decd20014a9e22699e39004a58bf01df/packages/foam-vscode/static/graphs/default/graph.js#L39-L49

riccardoferretti: yes.
actions are used to update the model.
the model is rendered by the view (which was initialized once in `initGraph`, the `force-graph` library doesn't have a clear separation between model and view, but it's still good enough for us)

packages/foam-vscode/static/graphs/default/graph.js
jmg-duarte: So I tried to write 
```js
window.onload = () => {
  vscode.postMessage({
    type: "webviewStyleRequest",
  });
  initDataviz(vscode);
  console.log("ready");
  vscode.postMessage({
    type: "webviewDidLoad"
  });
};
```
And on the `didUpdateStyle`:
```js
case "didUpdateStyle":
        const style = message.payload;
        Actions.updateStyle(style);
```
And the graph does not update afterwards. I'm guessing the webview would need to be refreshed

packages/foam-vscode/static/graphs/default/graph.js
jmg-duarte: New commits fix this

jmg-duarte:
That CI fail is weird...

packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -3,15 +3,25 @@ import * as path from "path";
...
 import { Foam, Logger } from "foam-core";
 import { TextDecoder } from "util";
-import { getTitleMaxLength } from "../settings";
+import { getGraphStyle, getTitleMaxLength } from "../settings";
 import { isSome } from "../utils";
 
 const feature: FoamFeature = {
   activate: (context: vscode.ExtensionContext, foamPromise: Promise<Foam>) => {
+    let panel: vscode.WebviewPanel | undefined = undefined;
+    vscode.workspace.onDidChangeConfiguration(async event => {
```
riccardoferretti, line 12:
```suggestion
    vscode.workspace.onDidChangeConfiguration(event => {
```

packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -23,6 +33,7 @@ const feature: FoamFeature = {
         noteAddedListener.dispose();
         noteUpdatedListener.dispose();
         noteDeletedListener.dispose();
+        panel = undefined;
```
riccardoferretti, line 36:
```suggestion
        panel.dispose();
```

packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -110,7 +121,20 @@ async function createGraphPanel(foam: Foam, context: vscode.ExtensionContext) {
   panel.webview.onDidReceiveMessage(
     async message => {
       switch (message.type) {
+        case "webviewStyleRequest":
```
riccardoferretti, line 124: I don't believe you need this anymore

riccardoferretti (approved):
looks good to me. I added a couple of suggestions, and you can now remove the `webviewStyleRequest` message.

Once those changes are in place feel free to merge!

packages/foam-vscode/src/features/dataviz.ts
jmg-duarte: This is per-VSCode documentation https://code.visualstudio.com/api/extension-guides/webview

PR #443, opened 2021-01-13
Allow for the reuse of the graph panel

jmg-duarte:
This PR allows for the reuse of the graph panel.
In another words, `Foam: Show graph` no longer starts a new graph instead reusing the existing one.
## Before
before
## After
after

### DIFF
--- a/packages/foam-vscode/src/features/dataviz.ts
+++ b/packages/foam-vscode/src/features/dataviz.ts
@@ -20,33 +20,40 @@ const feature: FoamFeature = {
     });
 
     vscode.commands.registerCommand("foam-vscode.show-graph", async () => {
-      const foam = await foamPromise;
-      panel = await createGraphPanel(foam, context);
-      const onFoamChanged = _ => {
-        updateGraph(panel, foam);
-      };
+      if (panel) {
+        const columnToShowIn = vscode.window.activeTextEditor
+        ? vscode.window.activeTextEditor.viewColumn
+        : undefined;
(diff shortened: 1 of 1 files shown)


packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -20,33 +20,40 @@ const feature: FoamFeature = {
...
-      panel.onDidDispose(() => {
-        noteAddedListener.dispose();
-        noteUpdatedListener.dispose();
-        noteDeletedListener.dispose();
-        panel = undefined;
-      });
+      if (panel) {
+        const columnToShowIn = vscode.window.activeTextEditor
+        ? vscode.window.activeTextEditor.viewColumn
+        : undefined;
```
riccardoferretti, line 26: what does `undefined` do here? Could it make sense to use the `vscode.ViewColumn.Two` value, which is what we use when we create the panel the first time?

riccardoferretti (approved):
I have one more comment/question. Resolve it as you see fit, then feel free to merge

packages/foam-vscode/src/features/dataviz.ts
jmg-duarte: I'll be honest, I took it from here https://code.visualstudio.com/api/extension-guides/webview#webviews-api-basics
However, I don't think we should default to a value since the column might have changed, as I do in the example.

riccardoferretti: sounds good

PR #448, opened 2021-01-14
Prettier format and chores

riccardoferretti:


ingalless (approved):
Looks good to me! Nice changes
LGTM!

jmg-duarte:
There are some unrelated formats (such as author related) but regardless, I think that an husky pre-commit hook to enforce format should be added

riccardoferretti:
yes I have put a bunch of changes for which a PR wasn't "worth it", but they are all unrelated. see commits and files outside fo `foam-vscode` for anything else.

looking at husky now

riccardoferretti:
done. I have (re-)enabled  husky and fixed another configuration issue, which uncovered some more broken rules, which have also been fixed now

jmg-duarte:
I don't understand why you're disabling the linter in some places

riccardoferretti:
because the rule per se is not bad, but in that specific case the warning/error is uncalled for.
(e.g. you can't use `{` in strings unless they are template strings, but there we do so on purpose)
 
So I don't want to disable the rule globally, but I don't want to change the code to make the linter happy.

PR #449, opened 2021-01-14
Style graph nodes by type

riccardoferretti:
This allows nodes in the graph to be styled by using the `type` property of the note (which can be set via YAML frontmatter, or via a plugin).

image


jmg-duarte (approved):
Seems good to me

PR #457, opened 2021-01-23
Orphaned notes

### LINKED ISSUE #367: Feature Request: List all docs and images which are unlinked by others
Sometimes there are so many files that I can't find out if there are some docs or images in my repository that aren't referenced by other files.

So I was wondering if it could create a file from command palette in a specified path which listed all the unlinked files and images by wiki-link or markdown-link to let me know all the files that are not linked, so that I can  link them or delete them.

leonhfr:
Closes #367 

### Summary

This PR adds an `Orphans` panel that references all notes that have neither backlinks nor forward links.

There are two new settings:
- `foam.orphans.exclude`: exclude some repositories from the orphans report (for example, excluding the `/journal` for daily notes)
- `foam.orphans.groupBy`: either group orphans by folder, or not at all, defaults to folder. The orphans are then sorted alphabetically.

I had a first iteration that allowed creating an `orphans.md` note with a command. However, it feels redundant with the panel, which also has the advantage of self updating, so I think we could leave that part of the issue out.

### Todo

- [ ] Update documentation/recipe
- [ ] Testing

### Screenshot

orphans-1

Updating the groupBy setting on the fly (does not overwrite settings.json):

orphans-4

Toggle using the tree list / tree flat icons:

orphans-5

### DIFF
--- a/packages/foam-vscode/src/features/orphans.ts
+++ b/packages/foam-vscode/src/features/orphans.ts
@@ -1,17 +1,8 @@
-import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { Foam } from 'foam-core';
+import { Foam, Note } from 'foam-core';
 import { getOrphansConfig, OrphansConfig } from '../settings';
-import { focusNote } from '../utils';
 import { FoamFeature } from '../types';
-
-const ORPHANS_FILENAME = 'orphans';
-const ORPHANS_EXTENSION = '.md';
-const ORPHANS_TITLE = 'Orphans';
-
-// TODO: path option?
-// TODO: other than markdown for dialog?
 
 const feature: FoamFeature = {
   activate: async (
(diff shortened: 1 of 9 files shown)


packages/foam-vscode/src/features/orphans.ts
```diff
@@ -0,0 +1,130 @@
...
+    this.computeOrphans();
+    this._onDidChangeTreeData.fire();
+  }
+
+  private computeOrphans() {
+    this.orphans = this.foam.notes
+      .getNotes()
+      .filter(note => !this.foam.notes.getAllLinks(note.uri).length)
+      .filter(note => {
+        const p = note.uri.path.replace(this.root, '');
```
riccardoferretti, line 51: I noticed that `root` uses `fsPath`, whereas here you use `path`, might be better to use `path` across the board?

packages/foam-vscode/src/features/orphans.ts
```diff
@@ -0,0 +1,130 @@
...
+        const { dir } = path.parse(p);
+        return !this.exclude.includes(dir);
```
riccardoferretti, line 53: have you considered using globs for excludes?

packages/foam-vscode/src/features/orphans.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    return orphans;
+  }
+
+  getTreeItem(item: OrphanTreeItem): vscode.TreeItem {
+    return item;
+  }
+
+  getChildren(directory?: Directory): Thenable<OrphanTreeItem[]> {
+    if (!directory && this.config.groupBy === 'folder') {
```
riccardoferretti, line 83: IIRC unfortunately in VsCode it's not possible for extensions to have buttons in their treeview.
It might be worth double checking because it would be amazing to be able to flip the setting here

image

riccardoferretti (commented):
This looks amazing! I left a few questions, in general I prefer it much more to the command approach

packages/foam-vscode/src/features/orphans.ts
leonhfr: Is it an option to add micromatch https://github.com/micromatch/micromatch as a dependency?

riccardoferretti: micromatch is already a dependency of `foam-core`, so I don't see any issue in adding it here as well

packages/foam-vscode/src/features/orphans.ts
leonhfr: @riccardoferretti actually I managed to add see, see the incoming coming and the gif in the PR description

packages/foam-vscode/src/features/orphans.ts
leonhfr: sorted!

packages/foam-vscode/src/features/orphans.ts
riccardoferretti: That's great!

A few notes:
- what's your take on the best way to manage the setting vs the switch?
- can we use the same icons that VsCode already uses in the Source Control panel? https://github.com/microsoft/vscode-extension-samples/tree/master/webview-codicons-sample (this example is for webview, but might be a good starting point to find the needed info)
- is there a way to not show the switch command in the command palette? I can't imagine someone needing it there (but not sure whether VsCode allows for that)

image
image

leonhfr: @riccardoferretti 

> what's your take on the best way to manage the setting vs the switch?

I meant to ask your take on that too. I see two solutions:
- either keep both switch and setting; the setting is the default for the next time the repo is opened
- discard the setting completely and rely uniquely on the switch

Will look at the other points later, but I guess it's possible maybe

riccardoferretti: I agree, the one thing I want to avoid is the switch affecting the settings, which would be too noisy. 

I am AFK now, but I am curious to see if the source control panel remembers the last state, and if so, how.

I am comfortable with either approach, as we already support the setting let's keep it, and make it the default state for the panel.
For people who don't care, they won't have the setting defined and this would just work. For people who care, they will be able to adjust it.

leonhfr: @riccardoferretti coming back on the two other points:

- it is possible to hide the groupby toggle command from the palette, see latest commit
- I went through quite a few docs but I don't think it's possible to have a toggle between two icons in an extension-contributed menu, if you know of any extension that has this let me know and I'll check how they did it. What we can do now is use either the flat list or tree list icon, but no toggle. Below is a more detailed explanation of my findings

---

First, it seems the toggle from the SCM view is baked into vscode, so extensions cannot reuse the same mechanism:

https://github.com/microsoft/vscode/blob/d4be66da200acc87e01d3517bfddf799fd8d110b/src/vs/workbench/contrib/scm/browser/scmViewPane.ts#L1208-L1217

Then, what I though would do the trick would be to register the button for the toggle command two times, with a different set of icons and a different command, and hide one of them using the `when` clause depending on state. Something like this:

```json
    "menus": {
      "view/title": [
        {
          "command": "foam-vscode.toggle-orphans-list-tree",
          "when": "view == foam-vscode.orphans && someState == flat", // here manage the state
          "group": "navigation"
        },
        {
          "command": "foam-vscode.toggle-orphans-list-flat",
          "when": "view == foam-vscode.orphans && someState == tree", // here manage the state
          "group": "navigation"
        }
      ],
      "commandPalette": [
        {
          "command": "foam-vscode.toggle-orphans-list-tree",
          "when": "false"
        },
        {
          "command": "foam-vscode.toggle-orphans-list-flat",
          "when": "false"
        }
      ]
    },
    "commands": [
      {
        "command": "foam-vscode.toggle-orphans-list-tree",
        "title": "Foam: Display orphans grouped by folder",
        "icon": {
          "light": "static/icons/light/list-tree.svg",
          "dark": "static/icons/dark/list-tree.svg"
        }
      },
      {
        "command": "foam-vscode.toggle-orphans-list-flat",
        "title": "Foam: Display all orphans",
        "icon": {
          "light": "static/icons/light/list-flat.svg",
          "dark": "static/icons/dark/list-flat.svg"
        }
      }
    ],
```

Thing is, I don't think we can manage the state with the `when` clause, as we can't access any state managed by the extension from there: https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts Or at least I don't see any way yet

What it can do is, it can read from the whole `config` object, but I think that'd mean we'd overwrite the user defined setting for `groupBy` :thinking: and that's not what we want.

If we could access `globalState` or `workspaceState` from the `when` clause though we'd be golden and could have the toggle: https://code.visualstudio.com/api/references/vscode-api#ExtensionContext

riccardoferretti: I have seen that this extension (https://github.com/Gruntfuggly/todo-tree/) is achieving the button switch, and here is the approach they seem to be taking: https://github.com/Gruntfuggly/todo-tree/search?q=todo-tree-flat - what do you think?

leonhfr: That looks like exactly what we needed, seems like the docs for this were well hidden: https://code.visualstudio.com/api/extension-guides/command#using-a-custom-when-clause-context

leonhfr: @riccardoferretti Should be all good now! Just pushed the toggle and added a GIF

packages/foam-vscode/src/features/orphans.ts
```diff
@@ -0,0 +1,159 @@
...
+      vscode.commands.registerCommand(
+        'foam-vscode.group-orphans-by-folder',
+        () => provider.setGroupBy(OrphansConfigGroupBy.Folder)
+      ),
+      vscode.commands.registerCommand('foam-vscode.group-orphans-off', () =>
+        provider.setGroupBy(OrphansConfigGroupBy.Off)
+      )
+    );
+
+    foam.notes.onDidUpdateNote(() => provider.refresh());
```
riccardoferretti, line 32: Should add this to the subscription? so the listener is properly disposed

riccardoferretti (commented):
everything looks good, I just left a question around an event listener disposal.
I think we only need tests and adding the feature to the docs - am I missing anything?

SanketDG:
This looks great, @leonhfr 

I just have concern, does this necessarily need to be a pane, given that this just needs to be checked that whether a note is orphaned or not, can we just do a notification message: `These are orphaned notes: <Note1> <Note2>`

Panes, I personally feel, are for introspection and browsing. Given we willl have Backlinks, Tag Explorer, and Outline in any Foam setup do we need one more?

leonhfr:
@SanketDG

I think the point is not to check that a given note is orphaned or not, but to be able to reference all current orphans, so they can be discovered / rediscovered / referenced with others. It does make sense then to use a panel instead of message or other solutions.

riccardoferretti (approved):
Fantastic work!

riccardoferretti:
@allcontributors add @leonhfr for code

PR #460, opened 2021-01-26
Welcome screen for Tag Explorer

leonhfr:
Adds a welcome screen for the Tag Explorer when no tags are found.

tag-explorer


riccardoferretti:
I like this, I would just include both ways in which tags can be added (the other one being via the `tags:` front matter property)

leonhfr:
@riccardoferretti added!

riccardoferretti (approved):

PR #467, opened 2021-01-29
Refactor core workspace model 

riccardoferretti:
This PR improves the workspace handling by `foam-core`. It has several goals:
- improve the quality of the code and its readability 
- improve support for placeholders (which in the current implementation have serious limitations)
- add support for other resource types, like images/attachments

---

Current implementation, with `NoteGraph`: all md files are mapped into a `Note` object and these objects have relationships across them, defined with `NoteLink` objects (mind that a `NoteLink` simply describes a link found in a note, it can't know what it connects to, because it doesn't know what other notes are present in the system).
`NoteGraph` brings all of this together by managing the connections amongst these objects.

New implementation, with `FoamWorkspace`: 
- we can map all files in the workspace as `Resource` objects, which can be `Note`, `Attachment` and `Placeholder`
  - that means we can very easily support links and analysis on more than just md files (I haven't enabled that yet, and there are some considerations around performance)
- `Placeholder` is a type of resource that identifies `[[wikilinks]]` that don't point to any note
  - by treating them as first class resources, handling them is simplified (e.g. look at `dataviz.ts`)

---

Here are the major differences between `FoamWorkspace` (the new implementation) and `NoteGraph` (the old one):
- `NoteGraph` computes connections amongst notes as new notes are added to it
  - this has the advantage of always being up to date
  - this has the massive disadvantage that not all information is available when doing link lookups, and today results in various limitations of the model
- `FoamWorkspace` has two phases: one during which the workspace is populated but no links are established, and one where the workspace is resolved (that is, the connections across resources are created)
  - it is possible to keep the connection graph up to date, which is faster and less prone to forgetting to update, but requires solid testing of various edge cases to avoid inconsistencies over time
  - Alternatively the graph can be recomputed when needed, which might be simpler to deal with but we need to check if the computation cost is high
- removed dependency on external graph library (which was doing some magic that was hard to understand and account for), all relationships are managed locally by the class with hashmaps

TODO:
- [x] consolidate the way a URI is converted to key for resources/placeholders (put that in a fn)
- [ ] ~~add plugin system to `FoamWorkspace`~~
- [x] update features to use `FoamWorkspace`
- [ ] TBD support for slugification (cc @SanketDG)
- [x] better support for multiple notes mapped to same key (e.g. `/path/to/file-a.md` and `/another/file-a.md`, which both resolve to `[[file-a.md]]`)
- [x] improve error handling of direct links that don't point to anything (broken links)
- [x] remove `NoteGraph` implementation + graph dep


packages/foam-core/src/model/note.ts
```diff
@@ -30,8 +31,20 @@ export interface NoteLinkDefinition {
...
+export interface Attachment extends BaseResource {
+  type: 'attachment';
+}
+
+export interface Placeholder extends BaseResource {
+  type: 'placeholder';
+}
+
+export interface Note extends BaseResource {
+  type: 'note';
```
leonhfr, line 47: Suggestion here, using an enum for the different types. Something like:

```ts
enum ResourceType {
  Attachement = 'ATTACHEMENT',
  Placeholder = 'PLACEHOLDER',
  Note = 'NOTE',
}
```

IMHO it makes it easier to reference and to define behaviors depending on the type, if needed. For example:

```ts
function renderResource(resource: Resource): void {
  switch (resource.type) {
    case ResourceType.Attachement:
      renderAttachement(resource); // resource is already type Attachement
      break;
    case ResourceType.Placeholder:
      renderPlaceholder(resource);
      break;
    case ResourceType.Note:
      renderNote(resource);
  }
}
```

jmg-duarte: Adding a detail, why doesn't `BaseResource` have a type which then gets overriden?

riccardoferretti: because that's the pattern to do type unions in typescript: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions

Having the field on `BaseResource` as string would allow any value there. having it as a union of the 3 strings is no different than what I have written, but making changes would require the two sets to be consistent.

The reason why `BaseResource` exists is because `uri` is everywhere, so instead of adding it 3 times I put it in the base class, but I'd rather move `uri` to the 3 resource types than move type to `BaseResource` - do you know what I mean?

Regarding using the enum, I am not opposed to it per-se, it just feels heavier than a simple union type, and I am not sure it has advantages (you can do the switch with the union type as well, and TS also detects the `never` type).
That being said, I also see VSCode uses that a lot (although I don't love it), and am open to other opinions.

Just came across this, which describes pretty well my preference for union types: https://blog.bam.tech/developer-news/should-you-use-enums-or-union-types-in-typescript

leonhfr: Some valid points in this article indeed. I guess both are fine. Just wondering - are you planning on exporting those types and using them from `foam-vscode` as well? In this case enums might make it clearer on the possible type values, as you'd have to import it as well :thinking: But maybe that's just the pattern I'm used to.

packages/foam-core/src/model/workspace.ts
```diff
@@ -0,0 +1,401 @@
...
+  ): FoamWorkspace {
+    workspace.links = {};
+    workspace.backlinks = {};
+    workspace.placeholders = {};
+
+    workspace = Object.values(workspace.list()).reduce(
+      (w, resource) => FoamWorkspace.addLinksForResource(w, resource),
+      workspace
+    );
+    if (keepMonitoring) {
```
leonhfr, line 178: This is the part where you said:

> it is possible to keep the connection graph up to date,
> Alternatively the graph can be recomputed when needed

right?

packages/foam-core/src/model/workspace.ts
```diff
@@ -0,0 +1,401 @@
...
+function normalizePath(pathValue: string) {
+  const { ext } = path.parse(pathValue);
+  return ext.length > 0 ? pathValue : pathValue + '.md';
+}
+
+function normalizeKey(pathValue: string) {
+  return path.parse(pathValue).name;
+}
+
+export class FoamWorkspace implements IDisposable {
```
leonhfr, line 45: So I compared this `FoamWorkspace` implementation with the `NoteGraph` one and those are my observations - you might have to take them with a grain of salt as it's my first time looking at those.

First, `FoamWorkspace` was easier to understand, so good job on that!

Second I'm not sure I understand the reason for having static methods here. The static methods all take a `FoamWorkspace` instance as an argument, so why not have the instance methods directly perform the role of the static methods?

I see having the graph state in hashmaps as interesting, the `graphlib` dependency did seem overkill. I don't think Foam will need to perform complex or multi hop queries so it does seem fine and easier to understand like this. Maybe it'd be interesting to benchmark the two implementations? :thinking: Though I don't expect a huge difference. 

What I'm wondering about is: with a key value store implementation like this, can we guarantee ACID properties? and how does it handle concurrent operations? Maybe it's not that important as since it's an extension, the graph database should only have on client :thinking:

leonhfr: Ah yeah, and I didn't get what you meant with:

> add plugin system to `FoamWorkspace`

?

riccardoferretti: currently `bootstrap` uses the `NoteGraph`, and provides hooks for local plugins to augment the note graph. That needs to be ported to the `FoamWorkspace` API. see https://github.com/foambubble/foam/blob/master/docs/features/foam-local-plugins.md

leonhfr: Thanks for the link, it makes sense now :+1:

packages/foam-core/src/model/workspace.ts
riccardoferretti: yup, that's right

packages/foam-core/src/model/workspace.ts
riccardoferretti: I actually missed your previous comment, let me reply.

> First, FoamWorkspace was easier to understand, so good job on that!

I really think NoteGraph is hard to understand, and graphlib was doing some magic that was creating more issues than it was solving. I am glad in that regard it is a success :)

> Second I'm not sure I understand the reason for having static methods here

A few reasons:
- I prefer the functional approach, I think it's easier to read in general (even though this is not pure functional because data is not immutable for performance reasons, I am "treating" it as if it were in the API, which still gives the readability benefit)
- A functional-only approach is a bit heavy imo when doing several operations (e.g. look at the tests when I set up the workspace). The class both provides a namespace for the operation, and a coherent grouping (I think it helps to think about the API)
- I have seen this pattern in the VSCode codebase (see `uri.ts`); mine is a bit stricter (all public fn have both static and instance methods) but the idea is the same (at times one approach works better than the other)
- I'd say the overhead of doing this hybrid approach is low (it would have actually been very low if typescript were a bit smarter, my original approach was to have all instance methods like `getNote = FoamWorkspace.getNote.bind(null, this)`, which is semantically correct but TS struggles to give me the proper type inferences when I later call that method :( )

Still, I am open to other views, I can see it's heavier than a pure OO or FP approach

> Maybe it'd be interesting to benchmark the two implementations? 

Yes I really should invest a bit of time creating some benchmarks. Both for simple in-memory operations, but also more real life e2e (generate a big repo with various types of connections)

> can we guarantee ACID properties?

node is single-threaded so I would expect that to just work? maybe I am missing something though (it's been a while since I have done serious multithreading)
also, I can see how some batch updates (e.g. rename a directory, which affects many notes) might require some special handling (but no different I think than when using graphlib)

jmg-duarte: IMO most of this discussion would be solved with adhere to VSCode API style whenever possible, this makes it consistent and easier for people to pick since the documentation examples match the code.

Regarding the ACID properties, I am unsure why that is a concern. Regardless, once again, we should take advantage of the offered APIs instead of building our own from the ground up.

leonhfr: > A few reasons

Thanks for the explanation. I see it interesting & working, just not used to seeing this

> Yes I really should invest a bit of time creating some benchmarks

I was thinking it's be simple to use an existing knowledge base for benchmarking; there are plenty of them out there that are public, huge, and using wikilinks. 

---

On another note, have you considered using the [workspace state](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext) as a key value store instead of having it directly in the class (probably this change is out of scope here)? I understand there is separation of concern between foam-core and foam-vscode, but maybe a reference to it could be passed to `FoamWorkspace`.

This way Foam could persist the data between sessions. It wouldn't have to recompute the whole graph on startup. I'm not 100% sure whether VS Code puts the storage in memory though, so maybe there are performance concerns.

Another big change I can see is that it'd make any write operation asynchronous (see `update` on [memento stores](https://code.visualstudio.com/api/references/vscode-api#Memento)).

riccardoferretti: interesting points! definitely worth investigating.
I am not familiar with public knowledge bases that could work for this, what did you have in mind?

leonhfr: @riccardoferretti I was thinking it'd be fine to pick someone's repo, no? Here are some public ones: https://github.com/KasperZutterman/Second-Brain

packages/foam-core/src/model/note.ts
riccardoferretti: yes probably that will be part of the Foam API that will be available to other extensions built on top of us.
what do you mean by enum being easier? asking because at times in vscode I actually find them hard to deal with, but also in my case I feel it's just a matter of habit

packages/foam-core/src/model/workspace.ts
riccardoferretti: do you think this is clear enough in terms of API? I was thinking of otherwise having an options object, something like
`resolveLinks({ keepMonitoring: true })`
at the moment this would be the only param, but who knows if in the future there might be more. more importantly now though, it's that the API feels more explicit then `resolveLinks(true)`

riccardoferretti:
ok, I have done the migration.
This is basically the bulk of the PR, and can be tested locally if you want to play with it.

Still have a few todo-items, but if the direction looks good, I am happy to move forward and wrap it up.

Thoughts @jmg-duarte @leonhfr @ingalless @SanketDG ?

riccardoferretti:
This is now ready to be merged, waiting for comments and blessings

packages/foam-core/src/model/note.ts
leonhfr: I meant enum are easier to import/export, so you don't have to retype the type in another package and can rely solely on the enum instead.

packages/foam-core/src/model/workspace.ts
leonhfr: Yeah I agree that something like
```
resolveLinks(workspace, { keepMonitoring: true })
```
is clearer and more easily extensible.

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -16,7 +16,9 @@ const feature: FoamFeature = {
         provider
       )
     );
-    foam.notes.onDidUpdateNote(() => provider.refresh());
+    foam.workspace.onDidUpdate(() => provider.refresh());
+    foam.workspace.onDidAdd(() => provider.refresh());
+    foam.workspace.onDidDelete(() => provider.refresh());
```
leonhfr, line 21: Thinking this will fix a bug reported recently :ok_hand:

leonhfr (approved):
Looks good to me!

jmg-duarte:
Besides the Actions linter having some warnings it looks good to me

jmg-duarte (approved):

PR #468, opened 2021-01-29
[FEATURE] Notes preview

### LINKED ISSUE #459: Note previews in the panel views
Following the implementation of the [Orphans](https://github.com/foambubble/foam/pull/457) panel and considering the existing Backlinks & Tag Explorer panels, I'm thinking it could be good to update the tooltip, currently the same as the description for the 3 panels, to show a preview of the note instead.

I don't have an example to show in a panel on hand, but think of the hover VS code does when hovering over typescript functions and types and displaying the formatted JSDoc references.

If deemed interesting I might take a shot at it.

leonhfr:
Closes #459 

This PR implements note previews in the tooltips for the Tag Explorer and Orphans panels (the ones controlled by Foam).

Because the API for markdown tooltips is only available from VS code 1.52.1 onward, and Foam supports VS code from 1.45.1, the behavior is slightly different according to the current version.

https://github.com/foambubble/foam/blob/1d579b4f473865ff64a8d3343940239616b93010/packages/foam-vscode/package.json#L13-L15

- `< 1.52.1`: simple tooltip with the first words from the note

note-preview-simple

- `>= 1.52.1`: markdown excerpt from the note

note-preview-md

Note that:
- we remove the front matter
- we remove the title (redundant as it is already present in the panels)
- we truncate notes that are too long

Tagging @riccardoferretti as discussed

### DIFF
--- a/packages/foam-vscode/src/extension.ts
+++ b/packages/foam-vscode/src/extension.ts
@@ -27,10 +27,14 @@ export async function activate(context: ExtensionContext) {
     const services: Services = {
       dataStore: dataStore,
     };
+    const foamContext = {
+      ...context,
+      dataStore,
+    };
     const foamPromise: Promise<Foam> = bootstrap(config, services);
 
     features.forEach(f => {
-      f.activate(context, foamPromise, dataStore);
+      f.activate(foamContext, foamPromise);
     });
 
     const foam = await foamPromise;
--- a/packages/foam-vscode/src/extension.ts
+++ b/packages/foam-vscode/src/extension.ts
@@ -30,7 +30,7 @@ export async function activate(context: ExtensionContext) {
     const foamPromise: Promise<Foam> = bootstrap(config, services);
 
     features.forEach(f => {
-      f.activate(context, foamPromise);
+      f.activate(context, foamPromise, dataStore);
     });
 
     const foam = await foamPromise;
--- a/packages/foam-vscode/src/features/orphans.ts
+++ b/packages/foam-vscode/src/features/orphans.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { Foam, Note, URI } from 'foam-core';
+import { Foam, FileDataStore, Note, URI } from 'foam-core';
 import micromatch from 'micromatch';
 import {
   getOrphansConfig,
@@ -13,13 +13,14 @@ import { getNoteTooltip, getContainsTooltip } from '../utils';
 const feature: FoamFeature = {
   activate: async (
     context: vscode.ExtensionContext,
-    foamPromise: Promise<Foam>
+    foamPromise: Promise<Foam>,
+    dataStore: FileDataStore
   ) => {
     const foam = await foamPromise;
     const workspacesFsPaths = vscode.workspace.workspaceFolders.map(
       dir => dir.uri.fsPath
     );
-    const provider = new OrphansProvider(foam, {
+    const provider = new OrphansProvider(foam, dataStore, {
       ...getOrphansConfig(),
       workspacesFsPaths,
     });
@@ -54,7 +55,11 @@ export class OrphansProvider
   private orphans: Note[] = [];
   private root = vscode.workspace.workspaceFolders[0].uri.fsPath;
 
-  constructor(private foam: Foam, config: OrphansProviderConfig) {
+  constructor(
+    private foam: Foam,
+    private dataStore: FileDataStore,
+    config: OrphansProviderConfig
+  ) {
     this.groupBy = config.groupBy;
     this.exclude = this.getGlobs(config.workspacesFsPaths, config.exclude);
     this.setContext();
@@ -99,6 +104,14 @@ export class OrphansProvider
 
     const orphans = this.orphans.map(o => new Orphan(o));
     return Promise.resolve(orphans);
+  }
+
+  async resolveTreeItem(item: OrphanTreeItem): Promise<OrphanTreeItem> {
+    if (item instanceof Orphan) {
+      const content = await this.dataStore.read(item.note.uri);
+      item.tooltip = getNoteTooltip(content);
+    }
+    return item;
   }
 
   private computeOrphans(): void {
@@ -162,7 +175,7 @@ class Orphan extends vscode.TreeItem {
   constructor(public readonly note: Note) {
     super(note.title, vscode.TreeItemCollapsibleState.None);
     this.description = note.uri.path;
-    this.tooltip = getNoteTooltip(note);
+    this.tooltip = undefined;
     this.command = {
       command: 'vscode.open',
       title: 'Open File',
@@ -179,7 +192,8 @@ export class Directory extends vscode.TreeItem {
     super(dir, vscode.TreeItemCollapsibleState.Collapsed);
     const s = this.notes.length > 1 ? 's' : '';
     this.description = `${this.notes.length} orphan${s}`;
-    this.tooltip = getContainsTooltip(notes);
+    const titles = this.notes.map(n => n.title);
+    this.tooltip = getContainsTooltip(titles);
   }
 
   iconPath = new vscode.ThemeIcon('folder');
--- a/packages/foam-vscode/src/features/orphans.ts
+++ b/packages/foam-vscode/src/features/orphans.ts
@@ -8,6 +8,7 @@ import {
   OrphansConfigGroupBy,
 } from '../settings';
 import { FoamFeature } from '../types';
+import { getNoteTooltip, getContainsTooltip } from '../utils';
 
 const feature: FoamFeature = {
   activate: async (
@@ -161,7 +162,7 @@ class Orphan extends vscode.TreeItem {
   constructor(public readonly note: Note) {
     super(note.title, vscode.TreeItemCollapsibleState.None);
     this.description = note.uri.path;
-    this.tooltip = this.description;
+    this.tooltip = getNoteTooltip(note);
     this.command = {
       command: 'vscode.open',
       title: 'Open File',
@@ -178,7 +179,7 @@ export class Directory extends vscode.TreeItem {
     super(dir, vscode.TreeItemCollapsibleState.Collapsed);
     const s = this.notes.length > 1 ? 's' : '';
     this.description = `${this.notes.length} orphan${s}`;
-    this.tooltip = this.description;
+    this.tooltip = getContainsTooltip(notes);
   }
 
   iconPath = new vscode.ThemeIcon('folder');
--- a/packages/foam-vscode/src/features/orphans.ts
+++ b/packages/foam-vscode/src/features/orphans.ts
@@ -1,26 +1,25 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { Foam, FileDataStore, Note, URI } from 'foam-core';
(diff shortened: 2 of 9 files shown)


packages/foam-vscode/src/utils.ts
```diff
@@ -184,3 +188,55 @@ export async function focusNote(notePath: string, moveCursorToEnd: boolean) {
...
+  if (version >= STABLE_MARKDOWN_STRING_API_VERSION) {
+    return formatMarkdownTooltip(note) as any;
+  }
+
+  return formatSimpleTooltip(note);
+}
+
+export function formatMarkdownTooltip(note: Note): MarkdownString {
+  const LINES_LIMIT = 15;
+  const markdown = firstNLines(stripFrontMatter(note.source.text), LINES_LIMIT);
```
riccardoferretti, line 216: what happens if there is an image? and if it's a tall image?

packages/foam-vscode/src/utils.ts
```diff
@@ -184,3 +188,55 @@ export async function focusNote(notePath: string, moveCursorToEnd: boolean) {
     editor.selection = new Selection(range.end, range.end);
   }
 }
+
+export function getContainsTooltip(notes: Note[]): string {
```
riccardoferretti, line 192: nice touch!

packages/foam-vscode/src/utils.ts
riccardoferretti: what happens if there are 15 very long lines?

It might not be easy to "ensure" the tooltip is under control (I guess we can't limit its height as a container), but as a heuristic lines, characters, and images should be enough

packages/foam-vscode/src/utils.ts
```diff
@@ -184,3 +188,55 @@ export async function focusNote(notePath: string, moveCursorToEnd: boolean) {
...
+  const markdown = firstNLines(stripFrontMatter(note.source.text), LINES_LIMIT);
+  return new MarkdownString(markdown);
+}
+
+export function formatSimpleTooltip(note: Note): string {
+  const WORDS_LIMIT = 30;
+  const words = removeMarkdown(stripFrontMatter(note.source.text))
+    .replace(/\r?\n|\r/g, ' ')
+    .replace(/\s+/g, ' ')
+    .replace(note.title, '')
```
riccardoferretti, line 225: what do you want to achieve by removing the title?

riccardoferretti (commented):
Looks pretty good overall! I left a few questions for you

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -29,7 +30,7 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
 
   private tags: {
     tag: string;
-    noteUris: vscode.Uri[];
+    notes: Note[];
```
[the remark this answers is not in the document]
leonhfr: @riccardoferretti Ok that sounds like a good argument. By reading the file content you mean directly something like `fs.read()`, right? Or is there a foam way to access the content? Though that would require probably passing the `foam` instance to the tree children which might not be ideal.

I'm kind of hesitant to use `fs.read` directly as it feels like an expensive operation to perform over an unknown and possibly large number of files. However there is a way for the tooltip to resolve to a promise instead of having to be pre computed, so we would only read the file content on hover. That'd mean having a min version of 1.52.1 though.

packages/foam-vscode/src/utils.ts
leonhfr: I thought it felt redundant as the mouse is hovering over it anyway, so it made sense to remove the first instance of the title. But we can keep it for sure, it's only for the tooltips < 1.52.1 anyway.

packages/foam-vscode/src/utils.ts
leonhfr: @riccardoferretti actually here I feel it's missing some sort of sorting, as now it's just from the order of the provided array. A simple sorting would be alphabetical but ideally we'd have something based on maybe the number of note connections (for tags. for orphans we can keep it alphabetical I guess).

packages/foam-vscode/src/utils.ts
leonhfr: Images are not shown. For now there's an ugly missing image icon instead so I was thinking of replacing the images by their alt text instead. So:

```
![This is an alt text](some/kind/of/uri.png)
```
could be replaced by:
```
Image: "This is an alt text"
```

And completely removed if no alt text is provided. wdyt?

leonhfr: But for sure the excerpt func has to be refined, any inputs are welcomed

packages/foam-vscode/src/features/tags-tree-view/index.ts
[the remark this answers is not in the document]
riccardoferretti: the right way to do that would be via the `Datastore`, but yeah, that's probably not being passed in all the way.

I am a big fan of dependency injection, but I gotta admit that at times it gets in the way 😅 
Open to other approaches (after all that's also what vscode does with its packages), although I wouldn't make that change part of this PR as a whole

packages/foam-vscode/src/utils.ts
riccardoferretti: Those are valid ideas, but I think it's ok to keep it simple for now - let's see once we start using the feature if we think it can be improved

packages/foam-vscode/src/utils.ts
riccardoferretti: yeah let's keep it simple, there might be some odd edge cases here

packages/foam-vscode/src/utils.ts
riccardoferretti: Here are my thoughts:
- re image, I believe it's fine to leave the original text, it's understandable by the user what's going on
- re size, the best pareto effort I can think of is either what you have, or then also limit each line to 100 char, or in general pick the first 1500 chars of the string and do the line limiting on those. It's a bit hard to judge without seeing the actual results, each of them have pros and cons. wdyt?

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -83,15 +85,12 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
...
-  ) {
+  constructor(public readonly tag: string, public readonly notes: Note[]) {
     super(tag, vscode.TreeItemCollapsibleState.Collapsed);
-    this.description = `${this.noteUris.length} reference${
-      this.noteUris.length !== 1 ? 's' : ''
+    this.description = `${this.notes.length} reference${
+      this.notes.length !== 1 ? 's' : ''
     }`;
-    this.tooltip = this.description;
+    this.tooltip = getContainsTooltip(notes);
```
riccardoferretti, line 93: Ok, I thought of this crazy idea, I can't not share it :)

Here are a couple of considerations for tags, which might be interesting to think about.
When I look at a tag probably I want to know where it shows up in the target note, in what context.
What if we showed in the description the list of lines where the tag shows up in the file. e.g.

```
12: - I need to check whether this would work #follow-up
25: - Check-in with Max #follow-up
```

where `12` and `25` are the line numbers.


Might be out of scope for this PR, but thought it'd be good to share.

This wouldn't change anything in the way we deal with the preview for orphans.

packages/foam-vscode/src/features/tags-tree-view/index.ts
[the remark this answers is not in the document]
leonhfr: I refactored by passing the data store to the tree provider, see latest commit

packages/foam-vscode/src/utils.ts
leonhfr: I improved the excerpt function, should be pretty good now!

packages/foam-vscode/src/features/tags-tree-view/index.ts
leonhfr: I like this idea, it could be very useful indeed!

I think it's outside the scope of the PR though, so I suggest we keep it as a future improvement.

leonhfr:
@riccardoferretti 

In the latest commit:

- refactored by passing the data store to the tree providers. We can make use of the `resolveTreeItem` function so that the tooltip is only computed on hover, and we read from file from there so we save the expensive operation. This function is also only called once so even if the user hovers over the same tooltip several times the file is only read once. I updgraded the min vs code version to 1.46.1 to use this API.
- I improved the excerpt function. It also adds an ellipsis indicating the remaining number of lines, and replaces images by their alt text (or by an empty string if no alt text is present).

note-previ

packages/foam-vscode/src/extension.ts
```diff
@@ -30,7 +30,7 @@ export async function activate(context: ExtensionContext) {
     const foamPromise: Promise<Foam> = bootstrap(config, services);
 
     features.forEach(f => {
-      f.activate(context, foamPromise);
+      f.activate(context, foamPromise, dataStore);
```
riccardoferretti, line 33: to pass the data store along, I would suggest either using the `context`, or passing the whole `services` object, so that in the future we don't need to keep adding fields to this call - I'd suggest the former (after all that's what the context is meant for), what do you think?

packages/foam-vscode/src/features/orphans.ts
```diff
@@ -1,24 +1,26 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { Foam, Note, URI } from 'foam-core';
+import { Foam, FileDataStore, Note, URI } from 'foam-core';
```
riccardoferretti, line 3: can we import the interface instead?

riccardoferretti (approved):
Looks good to me, left a couple of minor comments

packages/foam-vscode/src/extension.ts
leonhfr: By using the context, how would you pass it? By finding the data store in `context.subscriptions`? Because if not we only have a global and a workspace data store. So maybe passing the `services` is best?

riccardoferretti: what I meant is that we can have a `FoamContext` which extends the `ExtensionContext` that we are getting.

```
c: FoamContext = {
...context,
dataStore: ..
}
```

and that's what we pass to the various feature activations, does that make sense?

leonhfr: makes sense! both of your comments should be addressed now!

leonhfr:
@riccardoferretti can I ask you to rerun the CI on this? I think the failure on Windows might be a random hiccup

riccardoferretti:
ok, I am not sure what the issue is, but it's also been happening on @ingalless branch

I have rerun the tests a few times but to no avail.

I'd say let's remove that test (funny, I just added it as a blueprint when i first introduced the e2e setup) and unblock both branches

ingalless:
> ok, I am not sure what the issue is, but it's also been happening on @ingalless branch
> 
> I have rerun the tests a few times but to no avail.
> 
> I'd say let's remove that test (funny, I just added it as a blueprint when i first introduced the e2e setup) and unblock both branches

Sweet! I'll wait for this to merge, rebase master into my branch and see if I continue to see issues. That will at least prove the error to be my tests and not the copy to clipboard ones!

leonhfr:
@riccardoferretti you were right, disabling the test did the trick!

Issue #451, opened 2021-01-15
Numeric tags not being recognized in tag explorer


bserrao:
- Foam version: 0.7.7
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**
When using full numeric tags (#2021), those get correctly highlighted in VSCode but don't get listed on tag explorer

**Steps to reproduce**
1. Edit any markdown document inside Foam with a numeric tag (#2021)
2. Save file, and check tag explorer

**Additional information**
This only happens when using a fully numeric tag like #2021. If i include letters in any position of the tag, it works as expected.
For instance: #2021y , #20y21 are correctly picked up by tag explorer

Also look at #382 where changes where made related to tag parsing.

Thanks

mngad:
So I thought I'd approach this as a `good first issue`. Turns out this can be added by extending the regex from [a-zA-Z] to [a-zA-Z0-9]. 

However, there's a test that checks that tags with only numbers in them are ignored:

      it('ignores tags that only have numbers in text', () => {
        expect(
          extractHashtags('this #123 tag should be ignore, but not #123four')
        ).toEqual(new Set(['123four']));
      });

So this is obviously a desired feature and more of a discussion point than a `good first issue`.

pderaaij:
@riccardoferretti I've had the same thing as @mngad ran into. I was looking into solving this issue, but it looks the behaviour is intentional. Any thoughts how you'd like to see this continued?

riccardoferretti:
yup, numeric tags will not be supported for the time being

PR #470, opened 2021-01-31
Numeric tags

njnygaard:
image

```json
        "foam.tags.includeNumbers": {
          "type": "boolean",
          "description": "Index solely numeric tags e.g. #2021",
          "default": false
        }
```

Tests are passing. I've included one for just the numeric tags.
Please let me know if this follows standards. This is my first time working with a VSCode extension and passing settings values from `foam-vscode` to `foam-core` was challenging because I only saw it done with the ignore blobs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -21,15 +21,5 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "prettier.requireConfig": true,
   "editor.formatOnSave": true,
-  "jest.debugCodeLens.showWhenTestStateIn": [
-    "fail",
-    "unknown",
-    "pass"
-  ],
-  "cSpell.words": [
-    "exluding",
-    "frontmatter",
-    "precendence",
-    "unist"
-  ]
+  "jest.debugCodeLens.showWhenTestStateIn": ["fail", "unknown", "pass"]
 }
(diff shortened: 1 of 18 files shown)


njnygaard:
image
Sorry about the CI gauntlet.

njnygaard:
#451 Tagging Issue

jmg-duarte:
I don't think tagging targets should be optional. It's either all or nothing. Some users will expect numeric tags to work out of the box and search for an option is not a viable alternative.

njnygaard:
Hi @jmg-duarte, I neglected to include the previous closed PR: #469 
There was discussion there about making numeric tagging optional like this.
I'll let you, @ingalless, @SanketDG, and @riccardoferretti weigh in on direction.
Suffice it to say, it was easier to do without the vscode config :smile:

riccardoferretti:
@jmg-duarte why do you think that to be the case? so long as the "strategy" is consistent within one's own repo, what matters if other people consider numbers to be tags or not?

jmg-duarte:
@njnygaard I read the previous discussion. I simply moved here since the PR/Issue was closed.

@riccardoferretti adding these kinds of options only adds to the overall complexity of the project without providing a meaningful feature. At best we could add an ignore setting.

riccardoferretti:
After review and discussion with @jevakallio this doesn't feel like a big enough priority atm so we won't move forward

Issue #455, opened 2021-01-22
Graph nodes are not properly highlighted on light themes 


riccardoferretti:
- Foam version: 0.8.0
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes 

**Summary**
When a node is selected in the graph, the other nodes don't become less prominent.

I believe the reason is that when we get the style for the nodes that are "lessened", we darken the color.
That works on a dark background, but doesn't on a light one.
The graph needs to know what type of theme it's on, and needs to darken or lighten the base color as a consequence.

**Steps to reproduce**
1. Select a light theme
2. open the graph
3. highlight a node

You will see that the other nodes are still in their "full color".
![image](https://user-images.githubusercontent.com/457005/105512777-d536f100-5cd1-11eb-84d3-a6b18b5f8659.png)

![image](https://user-images.githubusercontent.com/457005/105512848-e8e25780-5cd1-11eb-816a-3818a0978d4a.png)

riccardoferretti:
this has been fixed in #479

AliArdaOrhan:
This still not working in 0.10.3 despite there isn't any problem with 0.9.x

riccardoferretti:
@AliArdaOrhan what problem are you seeing? can you add a screenshot?

AliArdaOrhan:
All nodes seems highlighted in new versions. That makes it really hard to navigate. First one is from 0.9.1 version. Second ss is from 0.10.3 version. At second SS you can see that **Portfolio Management** node is selected, however all the other as well are looked highlighted.

![image](https://user-images.githubusercontent.com/8167817/109507405-c05b3380-7aaf-11eb-8956-664554133d37.png)

![image](https://user-images.githubusercontent.com/8167817/109507659-02847500-7ab0-11eb-99be-bee41416d783.png)

riccardoferretti:
Can you please create a new issue to track this @AliArdaOrhan ?

PR #479, opened 2021-02-05
Added line width and line color to graph settings

nitwit-se:
Suggested changes inspired by [Issue 455](https://github.com/foambubble/foam/issues/455)

- Added `foreground` setting to define line colors 
- Use transparency to shade lines and nodes instead of just using d3...darker: my suggested approach for handling highlighting regardless of dark or light mode themes
- Added `lineWidth` setting to define line width and particle width, can be useful when using light mode themes
- Bug workaround for placeholder notes not being highlighted correctly - this bug workaround is presumably caused by a deeper bug that I couldn't track down

First time working with VS Code extensions - so appologies if this PR is of poor quality, feedback most welcome!

### DIFF
--- a/packages/foam-vscode/static/graphs/default/graph.js
+++ b/packages/foam-vscode/static/graphs/default/graph.js
@@ -4,7 +4,6 @@ const CONTAINER_ID = 'graph';
 const styleFallback = {
   background: '#202020',
   fontSize: 12,
-  lineColor: '#e0e0e0',
   lineWidth: 0.2,
   particleWidth: 1.0,
   highlightedForeground: '#f9c74f',
@@ -34,7 +33,7 @@ const defaultStyle = {
   background: getStyle(`--vscode-panel-background`) ?? styleFallback.background,
   fontSize:
     parseInt(getStyle(`--vscode-font-size`) ?? styleFallback.fontSize) - 2,
-  lineColor: getStyle('--vscode-editor-foreground') ?? styleFallback.lineColor,
+  lineColor: getStyle('--vscode-editor-foreground') ?? styleFallback.node.note,
   lineWidth: parseFloat(styleFallback.lineWidth),
   particleWidth: parseFloat(styleFallback.particleWidth),
   highlightedForeground:
--- a/packages/foam-vscode/static/graphs/default/graph.js
+++ b/packages/foam-vscode/static/graphs/default/graph.js
@@ -4,8 +4,9 @@ const CONTAINER_ID = 'graph';
 const styleFallback = {
   background: '#202020',
   fontSize: 12,
-  foreground: '#e0e0e0',
+  lineColor: '#e0e0e0',
   lineWidth: 0.2,
+  particleWidth: 1.0,
   highlightedForeground: '#f9c74f',
   node: {
     note: '#277da1',
@@ -33,10 +34,10 @@ const defaultStyle = {
   background: getStyle(`--vscode-panel-background`) ?? styleFallback.background,
   fontSize:
     parseInt(getStyle(`--vscode-font-size`) ?? styleFallback.fontSize) - 2,
-  foreground:
-    getStyle('--vscode-list-deemphasizedForeground') ??
-    styleFallback.foreground,
-  lineWidth: styleFallback.lineWidth,
+  lineColor:
+    getStyle('--vscode-list-deemphasizedForeground') ?? styleFallback.lineColor,
+  lineWidth: parseFloat(styleFallback.lineWidth),
+  particleWidth: parseFloat(styleFallback.particleWidth),
   highlightedForeground:
     getStyle('--vscode-list-highlightForeground') ??
     styleFallback.highlightedForeground,
@@ -166,17 +167,10 @@ function initDataviz(channel) {
     .d3Force('x', d3.forceX())
     .d3Force('y', d3.forceY())
     .d3Force('collide', d3.forceCollide(graph.nodeRelSize()))
-    .linkWidth(
-      () => parseFloat(model.style.lineWidth) || styleFallback.lineWidth
-    )
+    .linkWidth(() => model.style.lineWidth || styleFallback.lineWidth)
     .linkDirectionalParticles(1)
-    .linkDirectionalParticleWidth(link =>
-      getLinkState(link, model) === 'highlighted'
-        ? Math.max(
-            1.0,
-            parseFloat(model.style.lineWidth) || styleFallback.lineWidth
-          )
-        : 0
+    .linkDirectionalParticleWidth(
+      () => model.style.particleWidth || styleFallback.particleWidth
     )
     .nodeCanvasObject((node, ctx, globalScale) => {
       const info = model.nodeInfo[node.id];
@@ -225,11 +219,8 @@ function augmentGraphInfo(data) {
   data.links.forEach(link => {
     const a = data.nodes[link.source];
     const b = data.nodes[link.target];
-    // Workaround: nodes of type placeholder have mixed up id and uri
-    if (b.type == 'placeholder') a.neighbors.push(b.id.path);
-    else a.neighbors.push(b.id);
-    if (a.type == 'placeholder') b.neighbors.push(a.id.path);
-    else b.neighbors.push(a.id);
+    a.neighbors.push(b.id);
+    b.neighbors.push(a.id);
     a.links.push(link);
     b.links.push(link);
   });
@@ -262,11 +253,11 @@ function getLinkColor(link, model) {
   const style = model.style;
   switch (getLinkState(link, model)) {
     case 'regular':
-      return style.foreground;
+      return style.lineColor;
     case 'highlighted':
       return style.highlightedForeground;
     case 'lessened':
-      return d3.hsl(style.foreground).copy({ opacity: 0.5 });
+      return d3.hsl(style.lineColor).copy({ opacity: 0.5 });
     default:
       throw new Error('Unknown type for link', link);
   }
--- a/packages/foam-vscode/static/graphs/default/graph.js
+++ b/packages/foam-vscode/static/graphs/default/graph.js
@@ -4,6 +4,8 @@ const CONTAINER_ID = 'graph';
 const styleFallback = {
   background: '#202020',
   fontSize: 12,
+  foreground: '#e0e0e0',
+  lineWidth: 0.2,
   highlightedForeground: '#f9c74f',
   node: {
     note: '#277da1',
@@ -31,6 +33,10 @@ const defaultStyle = {
   background: getStyle(`--vscode-panel-background`) ?? styleFallback.background,
   fontSize:
     parseInt(getStyle(`--vscode-font-size`) ?? styleFallback.fontSize) - 2,
+  foreground:
+    getStyle('--vscode-list-deemphasizedForeground') ??
+    styleFallback.foreground,
+  lineWidth: styleFallback.lineWidth,
   highlightedForeground:
     getStyle('--vscode-list-highlightForeground') ??
     styleFallback.highlightedForeground,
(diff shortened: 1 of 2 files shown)


nitwit-se:
Jest tests seem to be failing with vscode 1.53.0 - even on master, suite.ts:23 is now causing a stack overflow. Possibly stream flushing causing a recursive call back to suite.ts's redirect of stderr? Ran out of time to investigate further today.

riccardoferretti:
yup, the issue is not related to your changes, will look properly into the problem tomorrow and keep you posted

riccardoferretti:
To unblock development, I am now forcing version 1.52 when running tests. 
I couldn't figure out the cause of the issue, if you have any idea about it let us know!
In the meantime, if you rebase from master, things will be fine now.

nitwit-se:
Thanks - no sorry I didn't have any luck either. It seems to be specifically output to stderr that triggers it - swapping out console.error with console.log works but then output is flooded with repeated lines of "START_NATIVE_LOG". Commenting out the code completely also works fine - but then no way of getting feedback on which tests fail if a test fails...   
Re-based and green now!

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -212,8 +225,11 @@ function augmentGraphInfo(data) {
   data.links.forEach(link => {
     const a = data.nodes[link.source];
     const b = data.nodes[link.target];
-    a.neighbors.push(b.id);
-    b.neighbors.push(a.id);
+    // Workaround: nodes of type placeholder have mixed up id and uri
```
riccardoferretti, line 228: what is happening here? not sure I understand

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -223,17 +239,19 @@ function augmentGraphInfo(data) {
...
+  const typeFill = d3.rgb(
+    style.node[info.type ?? 'note'] ?? style.node['note']
+  );
   switch (getNodeState(nodeId, model)) {
     case 'regular':
       return { fill: typeFill, border: typeFill };
     case 'lessened':
-      const darker = d3.hsl(typeFill).darker(3);
-      return { fill: darker, border: darker };
+      const transparent = d3.rgb(typeFill).copy({ opacity: 0.5 });
```
riccardoferretti, line 249: yes! thanks for this. it think it will fix #455

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -174,7 +187,7 @@ function initDataviz(channel) {
       const size = sizeScale(info.neighbors.length);
       const { fill, border } = getNodeColor(node.id, model);
       const fontSize = model.style.fontSize / globalScale;
-      let textColor = d3.rgb(fill);
+      let textColor = fill.toString();
```
riccardoferretti, line 190: do I understand correctly that colors are now d3 objects, and only at the final rendering stage they are being converted to strings? I like this

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -160,10 +166,17 @@ function initDataviz(channel) {
...
     .d3Force('collide', d3.forceCollide(graph.nodeRelSize()))
-    .linkWidth(0.2)
+    .linkWidth(
+      () => parseFloat(model.style.lineWidth) || styleFallback.lineWidth
+    )
     .linkDirectionalParticles(1)
     .linkDirectionalParticleWidth(link =>
-      getLinkState(link, model) === 'highlighted' ? 1 : 0
+      getLinkState(link, model) === 'highlighted'
+        ? Math.max(
```
riccardoferretti, line 175: what are you trying to convey here?
not sure this would achieve the desired particle effect, wouldn't it create particles that are as big as the line (so, invisible from a user's POV)?
Maybe we can make this an explicit setting, or computing it from `lineWidth`?

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -160,10 +166,17 @@ function initDataviz(channel) {
     .d3Force('x', d3.forceX())
     .d3Force('y', d3.forceY())
     .d3Force('collide', d3.forceCollide(graph.nodeRelSize()))
-    .linkWidth(0.2)
+    .linkWidth(
+      () => parseFloat(model.style.lineWidth) || styleFallback.lineWidth
```
riccardoferretti, line 170: shouldn't the parsing happen when we assign the value to `model.style.lineWidth`?

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -4,6 +4,8 @@ const CONTAINER_ID = 'graph';
 const styleFallback = {
   background: '#202020',
   fontSize: 12,
+  foreground: '#e0e0e0',
```
riccardoferretti, line 7: where is this value coming from?

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -31,6 +33,10 @@ const defaultStyle = {
   background: getStyle(`--vscode-panel-background`) ?? styleFallback.background,
   fontSize:
     parseInt(getStyle(`--vscode-font-size`) ?? styleFallback.fontSize) - 2,
+  foreground:
+    getStyle('--vscode-list-deemphasizedForeground') ??
```
riccardoferretti, line 37: do I understand correctly that this is the color of edges? in that case we should be more specific with the name

riccardoferretti (commented):
thanks for exposing these properties, I have left a few comments, let me know what you think!

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: It's an ugly workaround to a bug in dataviz.ts that is setting id to a dictonary and not a string for placeholder nodoes - sorry :) At the time I hadn't dug deep enough into the code to find the underlying bug. But I think I can now see how to fix this in datavviz.ts  instead - let me work on a better fix and update this PR!

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Yes - d3.rgb() seems to strip away any opacity for some reason, but returning d3 objects in getNodeColor() and then calling .toString() on them seems to preserve the alpha channel.

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: I didn't dig deep enough into how force-graph is implemented but it seems that when the line width is < 1, then particle width needs to be quite large relatively to be visible - but once line width is > 1 then setting particleWidth = lineWidth produced particles slightly wider than the line. But I like your idea of having it explicit - leaving people to play with the setting to achieve their preferred look. I'll udate this too!

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Good catch - I'll fix that, thanks!

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Thanks - yes, I will go with particleWidth, lineWidth and lineColor as settings.

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Reversed this change and fixed in dataviz.ts now.

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -31,6 +34,10 @@ const defaultStyle = {
   background: getStyle(`--vscode-panel-background`) ?? styleFallback.background,
   fontSize:
     parseInt(getStyle(`--vscode-font-size`) ?? styleFallback.fontSize) - 2,
+  lineColor:
+    getStyle('--vscode-list-deemphasizedForeground') ?? styleFallback.lineColor,
```
riccardoferretti, line 38: before this PR we were computing the line color with `d3.hsl(style.node.note).darker(2);`. 
Can we keep the same behavior in the default style? I am not sure that `--vscode-list-deemphasizedForeground` is a good enough approximation across themes

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -160,10 +167,10 @@ function initDataviz(channel) {
     .d3Force('x', d3.forceX())
     .d3Force('y', d3.forceY())
     .d3Force('collide', d3.forceCollide(graph.nodeRelSize()))
-    .linkWidth(0.2)
+    .linkWidth(() => model.style.lineWidth || styleFallback.lineWidth)
     .linkDirectionalParticles(1)
-    .linkDirectionalParticleWidth(link =>
-      getLinkState(link, model) === 'highlighted' ? 1 : 0
+    .linkDirectionalParticleWidth(
+      () => model.style.particleWidth || styleFallback.particleWidth
```
riccardoferretti, line 173: does this mean that particles will be always visible? (as opposed to only for selected nodes)

riccardoferretti (commented):
thanks for implementing the changes, I have a couple of extra comments, but overall this looks really good

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Good catch - that was embarassing that I didn't notice this when testing, thanks! Fixed shortly.

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Yes - did I understand right that, set it to `--vscode-editor-foreground` by default instead?  I'll fix that right away too.

riccardoferretti: originally I thought we should by default have `lineColor` be based off of the value of `node.note`, not `--vscode-editor-foreground`, so that if a user only changes `node.note`, it automatically changes the line color accordingly. 
The user can still manually override `lineColor` of course. 
Still debating whether that's a good idea, both because I think it makes things a little bit more complicated (nothing too crazy) and because maybe the user wouldn't expect that change (which, again, they could still reverse by providing a `lineColor` value. What do you think?

Instead of `.darker(2)` we can use the same transparency trick you have used everywhere else (e.g. `.opacity(.8)`) to make this work across light and dark themes.

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: removed `foreground` now

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: Sorry for the delay - day job got in the way. I agree that this is a much better approach - allowing the user to basically get away with only having to set background color and node color and most settings will fall into place. I've made the change and pushed it - let me know if I misunderstood anything :)

packages/foam-vscode/static/graphs/default/graph.js
```diff
@@ -244,11 +252,15 @@ function getLinkColor(link, model) {
...
     case 'regular':
-      return d3.hsl(style.node.note).darker(2);
+      return (
+        style.lineColor || d3.hsl(style.node['note']).copy({ opacity: 0.8 })
+      );
     case 'highlighted':
       return style.highlightedForeground;
     case 'lessened':
-      return d3.hsl(style.node.note).darker(4);
+      return style.lineColor
```
riccardoferretti, line 261: this approach makes perfect sense and conveys exactly the semantics we discussed, but there is something that doesn't feel right to me, for the following reasons:

- I feel that `styleFallback` and `defaultStyle` should have  the `lineColor` property, even if as `null`/`undefined` (not sure whether it would work), as that property is part of the style API. if the default value is `node['note']` even better because it conveys the semantics of what we want to achieve

- to me the logic in this line feels like it should live where we update all other styles, and in this part of the code we should simply read the property that we computed at "update time" (I see this logic more as BL than rendering logic, mostly to keep in line with the rest of the settings updating logic. I understand this is a personal preference, so I am not too fussed about it)

Just looking again at the code, I realized we are using the same logic a few lines above. This to me confirms that this check shouldn't live here but be part of the update logic.

thoughts?

riccardoferretti (commented):
we are getting there, now the code functionally does what we need, but I left a comment about the semantics that I would like to improve, wdyt?

packages/foam-vscode/static/graphs/default/graph.js
nitwit-se: See latest commit: I think you are right that it cleans things up and moves logic to a better place. Not sure if the change to updateStyle is the best way to do it though, let me know if you see a better way.

riccardoferretti (approved):
looks great!

riccardoferretti:
@allcontributors add @nitwit-se for code

nitwit-se:
Thanks for the great feedback and your patience @riccardoferretti

PR #482, opened 2021-02-08
Allow absolute paths in `openDailyNote.directory`

### LINKED ISSUE #481: `foam.openDailyNote.directory` should allow absolute paths, not only relative ones
I'm hoping to be able to open the daily note from any instance of VSCode (ie. from any directory) instead of having to open the `foam-template` cloned repository. When I press the `alt-d` shortcut, I want the daily page to open.

---

To that end, I set `foam.openDailyNote.directory` to the absolute path, instead of the default relative path:

```
`foam.openDailyNote.directory`: "/Users/michael/src/github.com/movermeyer/foam-data-repo/journal"
``` 

However, when I then press the `alt-d` shortcut, it treated the entire absolute path as a relative path, creating the very long path:

```
/Users/michael/src/github.com/movermeyer/other_project/Users/michael/src/github.com/movermeyer/foam-data-repo/journal/2021-02-07.md
```

instead of using the absolute path to make the expected note path: 

```
Users/michael/src/github.com/movermeyer/foam-data-repo/journal/2021-02-07.md
```

movermeyer:
Fixes #481 

This allows users to press the `alt-d` shortcut to open the daily note from any instance of VSCode, not just within the `foam-template` repo.

### DIFF
--- a/packages/foam-vscode/src/dated-notes.test.ts
+++ b/packages/foam-vscode/src/dated-notes.test.ts
@@ -1,0 +1,74 @@
+import { workspace } from 'vscode';
+import { getDailyNotePath } from './dated-notes';
+
+describe('getDailyNotePath', () => {
+  test('Adds the root directory to relative directories (Posix paths)', async () => {
+    const date = new Date('2021-02-07T00:00:00Z');
+    const year = date.getFullYear();
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
+    const isoDate = `${year}-0${month}-0${day}`;
+
+    await workspace
+      .getConfiguration('foam')
+      .update('openDailyNote.directory', 'journal/subdir');
+    const foamConfiguration = workspace.getConfiguration('foam');
+    expect(getDailyNotePath(foamConfiguration, date).path).toMatch(
+      new RegExp(`journal[\\\\/]subdir[\\\\/]${isoDate}.md$`)
+    );
+  });
+
+  test('Uses absolute directories without modification (Posix paths)', async () => {
+    const date = new Date('2021-02-07T00:00:00Z');
+    const year = date.getFullYear();
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
+    const isoDate = `${year}-0${month}-0${day}`;
+
+    await workspace
+      .getConfiguration('foam')
+      .update('openDailyNote.directory', '/absolute_path/journal');
+    const foamConfiguration = workspace.getConfiguration('foam');
+    expect(getDailyNotePath(foamConfiguration, date).path).toMatch(
+      new RegExp(`^/absolute_path/journal/${isoDate}.md`)
+    );
+  });
+
+  test('Adds the root directory to relative directories (Windows paths)', async () => {
+    const date = new Date('2021-02-07T00:00:00Z');
+    const year = date.getFullYear();
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
+    const isoDate = `${year}-0${month}-0${day}`;
+
+    await workspace
+      .getConfiguration('foam')
+      .update('openDailyNote.directory', 'journal\\subdir');
+    const foamConfiguration = workspace.getConfiguration('foam');
+    expect(getDailyNotePath(foamConfiguration, date).path).toMatch(
+      new RegExp(`journal[\\\\/]subdir[\\\\/]${isoDate}.md$`)
+    );
+  });
+
+  test('Uses absolute directories without modification (Windows paths)', async () => {
+    // While technically the test passes on all OS's, it's only because the test is overly loose.
+    // On Posix systems, this test actually does modify the path, since Windows style paths are
+    // considered to be relative paths. So while this test passes on Posix systems, it is not
+    // because it treats it as an absolute path, but rather that the test doesn't check the same thing.
+    // This was considered "good enough" instead of introducing a dependency like `skip-if` to skip the
+    // test on Posix systems.
+    const date = new Date('2021-02-07T00:00:00Z');
+    const year = date.getFullYear();
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
+    const isoDate = `${year}-0${month}-0${day}`;
+
+    await workspace
+      .getConfiguration('foam')
+      .update('openDailyNote.directory', 'C:\\absolute_path\\journal');
+    const foamConfiguration = workspace.getConfiguration('foam');
+    expect(getDailyNotePath(foamConfiguration, date).path).toMatch(
+      new RegExp(`/C:[\\\\/]absolute_path[\\\\/]journal[\\\\/]${isoDate}.md`)
+    );
+  });
+});
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -1,8 +1,8 @@
 import { workspace, WorkspaceConfiguration } from 'vscode';
-import { dirname, join } from 'path';
 import dateFormat from 'dateformat';
 import * as fs from 'fs';
-import { docConfig, focusNote, pathExists } from './utils';
+import { docConfig, focusNote, getDirname, pathExists } from './utils';
+import { URI } from 'foam-core';
 
 async function openDailyNoteFor(date?: Date) {
   const foamConfiguration = workspace.getConfiguration('foam');
@@ -17,13 +17,16 @@ async function openDailyNoteFor(date?: Date) {
   );
   await focusNote(dailyNotePath, isNew);
 }
-function getDailyNotePath(configuration: WorkspaceConfiguration, date: Date) {
-  const rootDirectory = workspace.workspaceFolders[0].uri.fsPath;
+
+function getDailyNotePath(
+  configuration: WorkspaceConfiguration,
+  date: Date
+): URI {
+  const rootDirectory = workspace.workspaceFolders[0].uri;
   const dailyNoteDirectory: string =
     configuration.get('openDailyNote.directory') ?? '.';
   const dailyNoteFilename = getDailyNoteFileName(configuration, date);
-
-  return join(rootDirectory, dailyNoteDirectory, dailyNoteFilename);
+  return URI.joinPath(rootDirectory, dailyNoteDirectory, dailyNoteFilename);
 }
 
 function getDailyNoteFileName(
@@ -42,7 +45,7 @@ function getDailyNoteFileName(
 
 async function createDailyNoteIfNotExists(
   configuration: WorkspaceConfiguration,
-  dailyNotePath: string,
+  dailyNotePath: URI,
   currentDate: Date
 ) {
   if (await pathExists(dailyNotePath)) {
@@ -56,7 +59,7 @@ async function createDailyNoteIfNotExists(
     configuration.get('openDailyNote.filenameFormat');
 
   await fs.promises.writeFile(
-    dailyNotePath,
+    dailyNotePath.fsPath,
     `# ${dateFormat(currentDate, titleFormat, false)}${docConfig.eol}${
       docConfig.eol
     }`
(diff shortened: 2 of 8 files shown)


riccardoferretti:
can you please rebase from master? the issue with the tests has been fixed

riccardoferretti:
I see the checks are still failing, but now the error seems related to the changeset

movermeyer:
> but now the error seems related to the changeset

@riccardoferretti Yup. As noted in the PR description, I haven't figured out the code needed to mock the `WorkspaceConfiguration` objects in the tests yet. I'll try again later this week.

Thanks for fixing the CI!

package.json
```diff
@@ -25,7 +25,9 @@
   },
   "devDependencies": {
     "all-contributors-cli": "^6.16.1",
-    "lerna": "^3.22.1"
+    "lerna": "^3.22.1",
+    "set-tz": "0.2.0",
```
movermeyer, line 29: This allows me to fix the timezone, so that the tests work when run locally, regardless of which OS and which timezone the developers machine is using.

package.json
```diff
@@ -25,7 +25,9 @@
...
+    "skip-if": "^1.1.1"
```
movermeyer, line 30: This allows use to skip tests based on the OS.

packages/foam-vscode/src/test/suite.ts
```diff
@@ -3,7 +3,7 @@
  * We use the following convention in Foam:
  * - *.test.ts are unit tests
  *   they might still rely on vscode API and hence will be run in this environment, but
- *   are fundamentally about testing functions in isolations
+ *   are fundamentally about testing functions in isolation
```
movermeyer, line 6: Fixes a minor typo I noticed.

movermeyer:
@riccardoferretti I finally managed to get everything sorted. It should be ready for your review now 🎉 🙏

packages/foam-vscode/src/dated-notes.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  skipIfWindows(
+    'Uses absolute directories without modification (Unix systems)',
+    async () => {
+      const date = new Date('2021-02-07T00:00:00Z');
+      await workspace
+        .getConfiguration('foam')
+        .update('openDailyNote.directory', '/absolute_path/journal');
+      const foamConfiguration = workspace.getConfiguration('foam');
+      expect(getDailyNotePath(foamConfiguration, date)).toMatch(
```
riccardoferretti, line 28: iiuc you are doing all this to deal with the different path format.

I understand the issue, that why we have started using `URI` across the code, except for the lowest fs service level (basically all platform specific translation is managed by the `FileDataStore`.

I would suggest you change `getDailyNotePath` to use URI instead, and consolidate the test by accessing the `path` property on the URI (which is not platform specific).

Beside making the tests easier to read/follow, and removing the skip dependency, this will make the code more solid in the first place.

thoughts?

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -18,12 +18,15 @@ async function openDailyNoteFor(date?: Date) {
...
-  const rootDirectory = workspace.workspaceFolders[0].uri.fsPath;
   const dailyNoteDirectory: string =
     configuration.get('openDailyNote.directory') ?? '.';
   const dailyNoteFilename = getDailyNoteFileName(configuration, date);
-
-  return join(rootDirectory, dailyNoteDirectory, dailyNoteFilename);
+  if (isAbsolute(dailyNoteDirectory)) {
+    return join(dailyNoteDirectory, dailyNoteFilename);
+  } else {
+    const rootDirectory = workspace.workspaceFolders[0].uri.fsPath;
```
riccardoferretti, line 27: in connection to my comment above, we should always be _very_ wary of using the `fsPath` property anywhere outside of the `DataStore`, as it immediately makes the code platform dependent. 
`URI` is the abstraction we want to use instead in all the business logic.

packages/foam-vscode/global-setup.js
```diff
@@ -0,0 +1,5 @@
+const setTZ = require('set-tz')
+
+module.exports = async () => {
+    setTZ('UTC')
```
riccardoferretti, line 4: what are you trying to achieve here and what problem is it solving?

riccardoferretti (commented):
thanks @movermeyer, I left a couple of comments, mostly around the use of `URI`, let me know your thoughts

packages/foam-vscode/global-setup.js
movermeyer: Without this code, if the tests are run locally on a developer's machine, and their computer's timezone is not 'UTC', the tests may not pass.

This is because `dateFormat` will use the date of the local timezone when formatting the filename that is used for the daily note's filepath.

For example, if the time is `2021-02-07T00:00:00Z`, it is only `2021-02-06T19:00:00-05:00` in `Americas/Toronto`, so if a developer in `Americas/Toronto` runs the tests locally, the test will fail. The test expects the filename to be `2021-02-07.md`, but the developer's local timezone will cause `dateFormat` to produce a filename of `2021-02-06.md`.

In order to have tests that succeed regardless of the developer's timezone, we need to fix the timezone used by the tests. `set-tz` does this in a way that is cross-platform (since fixing the timezone is not the same in Windows (much harder) as it is on Mac or Linux (much easier)).

packages/foam-vscode/src/dated-notes.test.ts
movermeyer: That seems to make sense. This is actually my first time writing Typescript, so I didn't know about `URI` 😅 .

I'll take a look and rework this PR later this week. Thanks for the great feedback! 🎉

riccardoferretti: hehe JS/TS are weird languages in many ways, welcome to the club :)

`URI` is actually not built into JS/TS, it's a utility class (that we shamelessly copied from the vscode repo to maintain compatibility) that we use exactly to abstract the platform specific details.
As it's compatible with vscode's, here you can decide whether to use our `URI` (see `foam-core/src/common/uri.ts` or the one coming from vscode (with a slight preference for our implementation for consistency))

packages/foam-vscode/global-setup.js
riccardoferretti: thanks for the explanation, sometimes I wonder whether we should use moment.js, but so far it's not been super necessary.

Back to my original point, the reason why I was asking was because we don't use TZ in our code.
So, I feel that we are now introducing TZ in the test code, which then we need to account for.

Let me elaborate:
what if instead of using `const date = new Date('2021-02-07T00:00:00Z');`, you used `const date = new Date();`, then generate the expected folder path? (you can also be specific by assigning day/month/year if you want to make the test always about the same value)

This way I think you wouldn't need to deal with TZ issues in the first place.

I am not against adding dependencies per se, but I lean towards avoiding it if possible to keep the code easier to reason about without knowing its ins and outs - "if possible" being the key words there :)

wdyt? open to hearing thoughts both on the specific case and on the general approach

movermeyer: Yeah, makes sense.

My only hesitation is a general reluctance to put non-determinism in test suites.

But in this case, perhaps doesn't justify the new dependency (a reason why JS/TS development is weird... everything is a dependency).

I'll look into it later this week.

packages/foam-vscode/src/dated-notes.ts
movermeyer: `rootDirectory` was already defined this way in the existing code. 🙈 

I will try to see if I can use `URI` for this.

packages/foam-vscode/src/dated-notes.test.ts
movermeyer: I've reworked it to use `foam-core`'s `URI` in _more_ places. Pushing the use of URI up a level allows the tests to avoid using the `skip-if` dependency.

packages/foam-vscode/global-setup.js
movermeyer: Alright, I now manually create the corresponding filename in the the tests instead of fixing the timezone. So I've removed the `set-tz` dependency.

movermeyer:
Alright @riccardoferretti, I've reworked this PR so it no longer uses the `set-tz` and `skip-if` dependencies.

It's ready for another round of review.

package.json
movermeyer: Dropped after rework.

package.json
movermeyer: Dropped after rework.

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -1,14 +1,15 @@
...
 import * as fs from 'fs';
 import { docConfig, focusNote, pathExists } from './utils';
+import { URI } from 'foam-core';
 
 async function openDailyNoteFor(date?: Date) {
   const foamConfiguration = workspace.getConfiguration('foam');
   const currentDate = date !== undefined ? date : new Date();
 
-  const dailyNotePath = getDailyNotePath(foamConfiguration, currentDate);
+  const dailyNotePath = getDailyNotePath(foamConfiguration, currentDate).fsPath;
```
riccardoferretti, line 12: one more comment about `URI`, as we want to use it wherever possible, until the last moment when we actually need the `fsPath`

Basically `createDailyNoteIfNotExists` should receive a `URI`, and when calling `writeFile` translate it into the corresponding `fsPath`

the goal is to never see an `fsPath` in the codebase which is not a direct argument to a `fs` call

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -17,13 +18,24 @@ async function openDailyNoteFor(date?: Date) {
...
+  date: Date
+): URI {
+  const dailyNoteDirectory: URI =
+    URI.file(configuration.get('openDailyNote.directory')) ?? URI.file('.');
   const dailyNoteFilename = getDailyNoteFileName(configuration, date);
 
-  return join(rootDirectory, dailyNoteDirectory, dailyNoteFilename);
+  const rootDirectory = isAbsolute(dailyNoteDirectory.path)
+    ? URI.file('.')
+    : workspace.workspaceFolders[0].uri; // TODO: This isn't guaranteed to exist (They could have no workspace open)
```
riccardoferretti, line 32: remove TODO and handle the case.

also, shouldn't this ternary operator be the other way around?

What does `URI.file('.')` resolves to?

packages/foam-vscode/src/features/open-dated-note.ts
```diff
@@ -213,7 +213,7 @@ const datedNoteCommand = (date: Date) => {
   if (foamNavigateOnSelect === 'createNote') {
     return createDailyNoteIfNotExists(
       foamConfig,
-      getDailyNotePath(foamConfig, date),
+      getDailyNotePath(foamConfig, date).fsPath,
```
riccardoferretti, line 216: same as above, pass the `URI`

riccardoferretti (commented):
it looks really good, I left a couple of extra comments around `URI`

packages/foam-vscode/src/dated-notes.ts
movermeyer: > remove TODO and handle the case.

I was mistaken. I was misreading a different error message at the time. This seems to be fine (indeed, I didn't change this code from what was there).

> shouldn't this ternary operator be the other way around?

No. I'm was trying to get a URI I can prepend as the first element of `URI.joinPath`. In the case of an absolute path, I needed a URI that would have no effect, so I forget how, but I found that `URI.file('.')` can serve that purpose.

I agree that it could be confusing, so I've rewritten this as a forking return statement.

packages/foam-vscode/src/dated-notes.ts
movermeyer: I've added a new commit that changes the signatures of all these methods to use URI instead of string for the paths.
Then the changes related to allowing absolute directories is done in a separate commit on top of that.

movermeyer:
@riccardoferretti Ready for another round. I've made everything in this area use `URI`s whenever possible, and cleaned up the confusing code you mentioned last time.

riccardoferretti (approved):
LGTM thx!

riccardoferretti:
@allcontributors add @movermeyer for code

PR #493, opened 2021-02-18
Feature/blank note explorer view

### LINKED ISSUE #491: New Explorer View for Blank Notes
I think an explorer view which displays notes which have no content (e.g. they are totally blank, or perhaps contain just a title) would be helpful for many of us. 

In my Foam workflow, I typically create a lot of "stub" topics, with intentions of going back and fleshing out those missing items. However, if I create several in a day, it becomes unwieldy to remember which topics are new and need attention.

I've already done the work to get this implemented, and it is essentially a carbon copy of the Orphan's view, but with different filter criteria. 

foam-blank-note

joeltjames:
This PR resolves #491, but creating a new "Placeholder" view.

It also extracts a `FilteredNotesProvider` from the current `OrphansProvider`.  This allows us to very easily create explorer views similar to the Orphans and Blank Notes view. The new path for doing this would just require adding the requisite settings, and then adding a `FoamFeature` which defines the match function.

Please let me know if I've missed anything in this code change.


packages/foam-vscode/src/features/orphans.ts
```diff
@@ -19,9 +13,12 @@ const feature: FoamFeature = {
...
-    const provider = new OrphansProvider(
+    const provider = new FilteredNotesProvider(
       foam.workspace,
       foam.services.dataStore,
+      'orphans',
+      'orphan',
+      (note: Note) => isOrphan(foam.workspace, note),
       {
         ...getOrphansConfig(),
         workspacesFsPaths,
```
joeltjames, line 24: This is the real "secret sauce" of this change. We no longer have a specific `OrphansProvider`, but rather we use a generic `FilteredNotesProvider` which provides all of the same methods, but requires the called to provide a filter function. In the case of the `Orphans` feature, that filter is:
```typescript
export function isOrphan(workspace: FoamWorkspace, note: Note) {
  return workspace.getConnections(note.uri).length === 0;
}
```

riccardoferretti:
this is interesting @joeltjames 
a couple of questions to put it in context:
- currently we use the name "placeholder" for links that have not been materialized as a note, have you considered combining this concept with what you call "blank note"? (in a way, a blank note is a placeholder)
- the logic that defines what a blank note is feels tricky, as e.g. a note with one line of content would show up as blank... maybe we could check if the note is just a heading? thinking out loud..

joeltjames:
>     * currently we use the name "placeholder" for links that have not been materialized as a note, have you considered combining this concept with what you call "blank note"? (in a way, a blank note is a placeholder)
I like the idea of combining these two, perhaps we could just merge the name, call this the "Placeholder" view, and then display placeholders in a slightly different format in the list
 
>     * the logic that defines what a blank note is feels tricky, as e.g. a note with one line of content would show up as blank... maybe we could check if the note is just a heading? thinking out loud..
I like the idea of checking for a title, honestly this was a quick and dirty way to get it running, and it probably makes sense to have something a little more robust.

riccardoferretti:
> we could just merge the name, call this the "Placeholder" view, and then display placeholders in a slightly different format in the list

I like this!

> I like the idea of checking for a title

Sounds good then, let's see how it plays out. it seems a fair argument to say that a note that is only heading doesn't really have any content, and therefore is just a placeholder

joeltjames:
@riccardoferretti I was able to get those changes made. Placeholders and Blank Notes now appear under the "Placeholder"s view. See the attached gif for a preview.
foam-placeholders

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  private sort(a: Resource, b: Resource) {
+    const titleA = getTitle(a);
+    const titleB = getTitle(b);
+    return titleA.localeCompare(titleB);
+  }
+}
+
+export interface FilteredResourcesProviderConfig
+  extends FilteredResourcesConfig {
+  workspacesFsPaths: string[];
```
riccardoferretti, line 197: can we use `URI` here instead?
The goal is to reduce the exposure of platform related strings, until the very last place where they need to be used.
(e.g. here wherever you use `workspacePaths` you'd be using `workspaceFolders.fsPath` internally, and exposing just the URI to the client)

packages/foam-vscode/src/features/orphans.test.ts
```diff
@@ -1,110 +1,37 @@
-import { OrphansProvider, Directory, OrphansProviderConfig } from './orphans';
-import { OrphansConfigGroupBy } from '../settings';
+// import { OrphansProvider, Directory, OrphansProviderConfig } from './orphans';
```
riccardoferretti, line 1: remove commented lines

riccardoferretti (commented):
This looks like a good approach. I have left a few comments, let me know your thoughts.

joeltjames:
@riccardoferretti, I just pushed a few commits to address your feedback. Thanks for the suggestions.

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -149,19 +149,21 @@ export class FilteredResourcesProvider
...
-      let folder = fsPath.replace(/\\/g, '/');
+    for (const fsPath of fsURI) {
+      let folder = fsPath.path.replace(/\\/g, '/');
       if (folder.substr(-1) === '/') {
         folder = folder.slice(0, -1);
       }
       exclude.push(...globs.map(g => `${folder}/${g}`));
     }
 
+    console.log(exclude);
```
riccardoferretti, line 165: remove 😅

packages/foam-vscode/src/features/placeholders.ts
```diff
@@ -0,0 +1,85 @@
...
+        'foam-vscode.group-placeholders-by-folder',
+        () => provider.setGroupBy(FilteredResourcesConfigGroupBy.Folder)
+      ),
+      vscode.commands.registerCommand(
+        'foam-vscode.group-placeholders-off',
+        () => provider.setGroupBy(FilteredResourcesConfigGroupBy.Off)
+      ),
+      foam.workspace.onDidAdd(() => {
+        provider.refresh();
+        foam.workspace.resolveLinks();
```
riccardoferretti, line 46: is there a reason for calling this? I don't believe you need to resolve the links here

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+          }
+        }
+      )
+    );
+  },
+};
+
+export default feature;
+
+export class FilteredResourcesProvider
```
riccardoferretti, line 57: We should call this `GroupedResourcesTreeDataProvider`

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  // prettier-ignore
+  private groupBy: FilteredResourcesConfigGroupBy = FilteredResourcesConfigGroupBy.Folder;
+  private exclude: string[] = [];
+  private filteredResources: Resource[] = [];
+  private root = vscode.workspace.workspaceFolders[0].uri.path;
+
+  constructor(
+    private workspace: FoamWorkspace,
+    private dataStore: IDataStore,
+    private filteredNoteType: string,
```
riccardoferretti, line 72: not sure if note type really describes what this is used for. This to me is almost the ID of the provider (as e.g. it needs to be unique to work with the settings commands)

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+    private filteredNoteContextValue: string,
```
riccardoferretti, line 73: this seems to me the label to use for the items in the tree view, is it correct? if that's the case, we should rename it to `resourceName` or something like that

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  setGroupBy(groupBy: FilteredResourcesConfigGroupBy): void {
+    this.groupBy = groupBy;
+    this.setContext();
+    this.refresh();
+  }
+
+  private setContext(): void {
+    vscode.commands.executeCommand(
+      'setContext',
+      `foam-vscode.${this.filteredNoteType}-grouped-by-folder`,
```
riccardoferretti, line 92: give that this provider requires specific commands, I believe it should also create/export these commands, e.g. by making them accessible via a `.commands` property, which would be used by the client to register them (there is still the issue of those commands needing to be consistent with what is in `package.json`, but at least we go from 3 places to keep consistent to 2)

I am referring to the commands defined in `features/placeholders.ts:36` (and corresponding ones for orphans)

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+    const titleA = getTitle(a);
+    const titleB = getTitle(b);
+    return titleA.localeCompare(titleB);
+  }
+}
+
+export interface FilteredResourcesProviderConfig
+  extends FilteredResourcesConfig {
+  workspacesURIs: URI[];
+  includeLinks?: boolean;
```
riccardoferretti, line 198: what's this setting for?

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+export default feature;
+
+export class FilteredResourcesProvider
+  implements vscode.TreeDataProvider<FilteredResourceTreeItem> {
+  // prettier-ignore
+  private _onDidChangeTreeData: vscode.EventEmitter<FilteredResourceTreeItem | undefined | void> = new vscode.EventEmitter<FilteredResourceTreeItem | undefined | void>();
+  // prettier-ignore
+  readonly onDidChangeTreeData: vscode.Event<FilteredResourceTreeItem | undefined | void> = this._onDidChangeTreeData.event;
+  // prettier-ignore
+  private groupBy: FilteredResourcesConfigGroupBy = FilteredResourcesConfigGroupBy.Folder;
```
riccardoferretti, line 64: `FilteredResourcesConfigGroupBy` => `GroupedResoucesConfigGroupBy`

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  extends FilteredResourcesConfig {
+  workspacesURIs: URI[];
+  includeLinks?: boolean;
+}
+
+type FilteredResourcesByDirectory = { [key: string]: Resource[] };
+
+type FilteredResourceTreeItem = FilteredResource | Directory;
+
+class FilteredResource extends vscode.TreeItem {
```
riccardoferretti, line 205: `ResourceTreeItem`

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+    } else if (isPlaceholder(this.resource)) {
+      iconStr = 'new-file';
+    } else {
+      iconStr = 'note';
+    }
+    this.iconPath = new vscode.ThemeIcon(iconStr);
+  }
+}
+
+export class Directory extends vscode.TreeItem {
```
riccardoferretti, line 237: `DirectoryTreeItem`

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  constructor(
+    public readonly dir: string,
+    public readonly resources: Resource[],
+    contextValue: string
```
riccardoferretti, line 241: this is not the `contextValue`, but the type of resource which is used for the label, we should call this `itemLabel` or something like that

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  private filteredResources: Resource[] = [];
+  private root = vscode.workspace.workspaceFolders[0].uri.path;
+
+  constructor(
+    private workspace: FoamWorkspace,
+    private dataStore: IDataStore,
+    private filteredNoteType: string,
+    private filteredNoteContextValue: string,
+    private filterPredicate: (resource: Resource, index: number) => boolean,
+    config: FilteredResourcesProviderConfig
```
riccardoferretti, line 75: why expanding the provider config? I would keep the `workspacesURIs` on their own (in reality these should be provided by the `FoamWorkspace` itself, but let's not include it in this PR - but it's also a good reason to keep it separate from the provider config)

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+}
+
+export interface FilteredResourcesProviderConfig
+  extends FilteredResourcesConfig {
+  workspacesURIs: URI[];
+  includeLinks?: boolean;
+}
+
+type FilteredResourcesByDirectory = { [key: string]: Resource[] };
```
riccardoferretti, line 201: generally speaking, I don't think we need to use `Filtered...` on all types/methods/params.
The main purpose of the provider is to show us some resources, in a way the fact that they are filtered is an implementation detail

packages/foam-vscode/src/features/filtered-resources.ts
riccardoferretti: I don't think that this class belongs to a feature.
we can put it in a `src/vscode-utils.ts` file or something like that.

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+} from '../settings';
+import { getContainsTooltip, getNoteTooltip } from '../utils';
+import { dirname, join } from 'path';
+import { FoamFeature } from '../types';
+import { commands } from 'vscode';
+
+const feature: FoamFeature = {
+  activate: (context: vscode.ExtensionContext) => {
+    context.subscriptions.push(
+      commands.registerCommand(
```
riccardoferretti, line 26: I understand you are making this a feature to be able to register this command.
But this command in reality is not related to the feature, it's just used by it.
I would move this into a `feature/utility-commands.ts` file

packages/foam-vscode/src/features/filtered-resources.ts
```diff
@@ -0,0 +1,252 @@
...
+      this.command = {
+        command: 'vscode.open',
+        title: 'Open File',
+        arguments: [resource.uri],
+      };
+    }
+    this.contextValue = contextValue;
+
+    let iconStr: string;
+    if (isAttachment(this.resource)) {
```
riccardoferretti, line 226: this is minor, but I think it reads better to have a:
 ```js
switch(note.type) {
case 'attachment':
...
}
```

File: packages/foam-vscode/src/features/placeholders.ts [hunk 19]
```diff
@@ -0,0 +1,85 @@
...
+      })
+    );
+  },
+};
+
+export default feature;
+
+export function isPlaceholderResource(resource: Resource) {
+  if (isAttachment(resource)) {
```
riccardoferretti, line 63: I don't think you need this, it's handled by the main return statement

packages/foam-vscode/src/features/placeholders.ts
joeltjames: If we don't resolve the links, then placeholders do not get updated. This is an issue when we click a true "placeholder" in the explorer view, which creates a blank note for that "placeholder". The blank note then shows up, but the "placeholder" link is still present. Resolving links cleans up the placeholders.

packages/foam-vscode/src/features/filtered-resources.ts
riccardoferretti: this class needs to be documented, in how it works and what it needs.
e.g. we need to say that:
- it will create commands in the shape of `foam-vscode.${this.filteredNoteType}-grouped-by-folder` and that they need to be defined in `package.json`
- the IDs of the provider (see comment below) need to be unique
- ...

riccardoferretti (commented):
There are some design considerations around the `FilteredResourcesProvider` class that I have been holding off to share because I wanted to clarify them in my head, but in the end I think are necessary for good integration with the codebase, let me know your thoughts - thanks!

packages/foam-vscode/src/features/placeholders.ts
riccardoferretti: can you check now whether this is still needed? 
I believe the behavior you are describing is due to a bug in 0.10.0 that was fixed in 0.10.1

packages/foam-vscode/src/features/filtered-resources.ts
joeltjames: I ended up moving it to `src/utils/grouped-resources-tree-data-provider.ts`.

packages/foam-vscode/src/features/filtered-resources.ts
joeltjames: I went with `providerId`

packages/foam-vscode/src/features/filtered-resources.ts
joeltjames: Yes, I like this a lot. I defined the commands as a getter on the provider, and then use a deconstructed array to insert those into the subscription.

packages/foam-vscode/src/features/filtered-resources.ts
joeltjames: This was ported over from the old oprhans code, I'm not sure what this setting was intended to do, but it has been removed.

packages/foam-vscode/src/features/filtered-resources.ts
joeltjames: This was ported over from the original OrphanProvider, so I didn't give it much thought. I have extracted it from the settings and just include it as a different parameter to the constructor.

packages/foam-vscode/src/features/placeholders.ts
joeltjames: Yes, pulling in 0.10.1 seems to have resolved this.

joeltjames:
@riccardoferretti I think I managed to clean up all of these items. Thanks again.

riccardoferretti (approved):
this looks good! I just left a comment about the naming used in the tests, but beyond that I think it's ready for merging

riccardoferretti:
@allcontributors add @joeltjames for code

PR #501, opened 2021-02-26
Add ability to launch a daily note on startup

ingalless:
I had this idea this morning that I'd like my foam notes to open on startup

It then got me thinking "hey, it would be cool if it would land on my daily note on startup too!"

So, here's that feature.

Update to docs following shortly...

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -169,6 +169,11 @@
             "Disable wikilink definitions generation"
           ]
         },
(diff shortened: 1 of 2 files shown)


ingalless:
@riccardoferretti Reading the docs I have just noticed that it was marked as "future" functionality, so if I'm missing something here let me know! I won't be hurt if you decline the PR 😆

riccardoferretti (approved):
LGTM!

PR #502, opened 2021-02-27
Improved wikilink definition resolution (fixes #499)

### LINKED ISSUE #499: Foam Graph is blank
`Open Foam Graph` opens the panel, but it is blank. My best guess is this is because I am on the VSCode for Insiders build in order to get native M1 Apple Silicon support. (Foam Graph worked fine on previous Intel laptop using standard VSCode)

riccardoferretti:
Wikilinks "redirected" via a wikilink definition to a path were not being properly checked for the existence of the target node. This PR addresses that.

In adding the tests I noticed that also for direct links where not creating the placeholder to the correct location, so I fixed that too.

Also improved the organization of the workspace tests.

### DIFF
--- a/packages/foam-core/src/model/workspace.ts
+++ b/packages/foam-core/src/model/workspace.ts
@@ -129,14 +129,17 @@ export class FoamWorkspace implements IDisposable {
     note: Note,
     link: NoteLink
   ): URI {
-    let targetUri: URI | null = null;
+    let targetUri: URI | undefined;
     switch (link.type) {
       case 'wikilink':
         const definitionUri = note.definitions.find(
           def => def.label === link.slug
         )?.url;
         if (isSome(definitionUri)) {
-          targetUri = parseUri(note.uri, definitionUri!);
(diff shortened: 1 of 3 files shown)


ingalless (approved):
Good stuff @riccardoferretti! Nice tests

packages/foam-core/src/model/workspace.ts
```diff
@@ -129,14 +129,17 @@ export class FoamWorkspace implements IDisposable {
     note: Note,
     link: NoteLink
   ): URI {
-    let targetUri: URI | null = null;
+    let targetUri: URI | undefined;
```
jmg-duarte, line 132: - Is this union on `undefined` really necessary? Couldn't we just add an empty URI?
- If it is necessary, shouldn't line 157 be updated to deal with it?

riccardoferretti: typescript is smart enough to see that by line 157 the type has become just URI. I like that, it better describes what is actually happening

PR #507, opened 2021-03-02
fix: fixing casing issue for windows

### LINKED ISSUE #488: Path bug in Windows
- Foam version: v0.9.1
- Platform: Windows 
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Not sure

**Summary**

Windows wiki-link bug

The Windows path is very weird. It's not at all portable because it is not a relative path.

**Steps to reproduce**
1.  Add a link to todo.md by [[todo]]
2.  Ctrl + S

**Additional information**

Feel free to attach any of the following that might help with debugging the issue:
- screenshots
- a zip with a minimal repo to reproduce the issue
- the Foam log in VsCode (see [instructions](https://github.com/foambubble/foam/blob/master/docs/foam-logging-in-vscode.md))

kenotron:
Fixes #488 

On Windows, the `uri` util file contains a compare relative paths with posix functions. The drive letters might not match on Windows (unique problem on win32 platform).

This fix introduces a small utility that is only relevant to Windows platform that would normalize drive letters in paths.


joshkreud:
Can confirm this works!

riccardoferretti (commented):
Thanks for contributing this @kenotron 

I see how the approach works, what I am concerned about is that it addresses the issue downstream instead of e.g. ensuring consistency across the app.

I would like to understand why in one case we get an uppercase path, and in the other we don't.

It feels like this normalization should live in the core model itself, as opposed to being applied "ad-hoc" when accessing the path.

wdyt?

joshkreud:
Hey, 
@riccardoferretti Yeah I agree. This cloud lead to other bugs in other places. 

Ao far I traced the lower case letters to: https://github.com/foambubble/foam/blob/32e443bbaea05c748396e37a126324633b8fabe5/packages/foam-core/src/bootstrap.ts#L16

But one interesting thing I noticed: 
After I applied this PR, everything ran fine.
Now, even though I reset to `master` all Drive letters are lower case and it works even without the Fix. 

Smells like a cache thing?

riccardoferretti:
yeah I think the change would need to be somewhere in the data store (which is the sole interface with the OS/platform), so that all clients would get a consistent model

Not sure about why it's not happening anymore, I did change in 0.10.3 the wikilink resolution in the graph to better take into account wikilink definitions, might be related to that one, although I am not positive as to where in the flow that would have happened

let's add a comment to the original issue and see if people are still experiencing the problem post 0.10.3

kenotron:
I still see this bug in 0.10.3. I suppose I can dig deeper in the layer (I'm pretty new to the codebase here, so maybe I can fix this PR to address the core).

riccardoferretti:
@kenotron did you have any luck here?

joshkreud:
Hi Again, 
Got back to my windows PC. 

I can see that the VSCode API seems to provide a capital drive letter.
https://github.com/foambubble/foam/blob/4e661aa6b5fe76aa892fb34828bdbd97f53b0f8e/packages/foam-vscode/src/features/wikilink-reference-generation.ts#L103-L104

The difference between `_fsPath` and `path` seems a bit weird to me: 
image

But in the Datastore the `findAllFiles` procedure returns lower case drive letters:
https://github.com/foambubble/foam/blob/6baeec4db6db0de8c7820d0aa2c54be27bfeff0e/packages/foam-core/src/services/datastore.ts#L149-L151

When it maps the Path strings to `URI` objects in L151, it runs into a Windows specific part here: 
https://github.com/foambubble/foam/blob/6baeec4db6db0de8c7820d0aa2c54be27bfeff0e/packages/foam-core/src/common/uri.ts#L331-L336


So if we assume that the Capital letter from VSCode is not a bug, this would be the place to replace it to capital?

riccardoferretti:
Thanks for looking into this.
There is something odd going on here: the `foam/packages/foam-core/src/common/uri.ts` is taken straight from VS Code, so in VS Code the transformation must happen somewhere else.

I feel I am losing track of the issue, let's take a step back: what problem are we solving? 

In https://github.com/foambubble/foam/issues/488#issuecomment-787865982 we talk about `noteUri`, tracking back the object to `foam-core/src/features/janitor.ts`, it also comes from the `FoamWorkspace`.
So, how come are they different?

joshkreud:
I actually haven't looked at the janitor yet. 

But stepping a bit back to the issue. 
It happens when updating the wiki-links. 

The URI for the document which is being saved right now has a Capital drive letter. (`window.activeTextEditor`)
All URIs in the foam datastore have a lower case drive letter. (`findAllFiles` returns a list of lower case drive letter strings that get mapped to `URI`, retaining the lower case letter)

The issue is, that since the drive letter case is different in both, `computeRelativePath` fails, because it cannot find the common path slice. 
This then results in Wiki-Links looking like in #488 or as seen in the following screenshot: 
image

joshkreud:
Hey
Just found this thing: https://github.com/microsoft/vscode/issues/116298
I think the root issue really is within VSCode and how events handle `URI` objects

riccardoferretti:
That's a really good find!

kenotron:
Okay I was able to update my change to @joshkreud 's suggestion. It is localizing the fix to that part for Windows only. If you either of you have a Windows box... please play with this branch!

It works locally on my Windows box now!

riccardoferretti:
is there a test you can add for this?

kenotron:
I don't really understand the build failure of that win-2019 ?

kenotron:
updated with test

riccardoferretti:
also, I noticed that in https://github.com/foambubble/foam/pull/507/files#diff-0f9418ec65d5cd4ec04508869ac06a3d03393e9cf47da10c35ab2949b3c391deL635 (`common/urits:635`) there is a manipulation of the drive letter - had you seen this?

kenotron:
Alright, so I'm going in circles here with this PR. I can revert back to that if you like keeping the tests. But please make a decision so I don't go back and forth on fixing this tiny bug. I actually believe my original code was the most elegant and preserves the vscode-uri bits as much as possible.

I can either fix it in utils/uri.ts like I had the first commit (https://github.com/foambubble/foam/pull/507/commits/a2223cda4e2b9ea491894fe7c0c2807981fe8064),

-or-

I can fix it in common/uri.ts (which is sort of "external" copy of vscode-uri). 

I've done both. I could replace the implementation of the "normalizeDriveLetters" to something like `uriToFsPath`, but that returns a Windows path. This won't work actually because `path.posix.relative()` which doesn't work with the Windows pathing. 

All in all, neither is ideal because node's built-in path.posix wasn't really up for this kind of task :(

riccardoferretti:
@kenotron I appreciate the help and I understand the frustration, please bear with me.

Given that:
- I can't replicate the issue on my system
- I have not dug deep enough in the problem to evaluate the solution
- The reason/cause of the issue itself is somewhat of a mystery and this is a brute force solution

My comments and questions which might look like a merry-go-round are meant to explore the solution space and lower the risk when touching such a core part of the model. 

Also, as part of the API v1, we also need to make adjustments to `URI` object/function (one more reason why I needed a test for the bug, to avoid regressions).

So, let's keep this on hold for a sec, and make sure it's included as part of the URI refactoring for API v1. I will make sure to include you when that PR is ready as you understand this issue better than me by now.

Feel free to let me know your thoughts

riccardoferretti:
@kenotron here is the PR: https://github.com/foambubble/foam/pull/537

I need to add your test, see it fail, and add your fix.
In the meantime if you have comments there, would love to hear your thoughts

riccardoferretti:
I have added the test, it fails (as expected).
I am still unclear whether this is something that needs to be normalized in the `URI` itself, or in the functions that manipulate it (e.g. `computeRelativePath`).

In Windows, are drive letters always upper/lower case? if we assume them to be one or the other, would it fail when that's not the case (is the drive letter case sensitive)?

Thanks for your help on this!

kenotron:
> I have added the test, it fails (as expected).
> I am still unclear whether this is something that needs to be normalized in the `URI` itself, or in the functions that manipulate it (e.g. `computeRelativePath`).
> 
> In Windows, are drive letters always upper/lower case? if we assume them to be one or the other, would it fail when that's not the case (is the drive letter case sensitive)?
> 
> Thanks for your help on this!

I don't want to give the impression that I'm upset here - thanks for the FREE software and the support! I think the issue is that I found that the bug affects a rather large portion of the user base (all Windows users will experience this) and the fix seems to me quite trivial. Given that you're already redoing the URI object:

1. will you be release that API v1 soon? this is a bleeding problem, and one that is steering me actively away from using Foam as my tool of choice... there are some alternatives I guess but I'm tending towards using a wikilinks extension for now as a stop gap
2. what's the harm in releasing something that isn't optimal if you already have something rewritten with a test case that will prove it is working (if v1 API will take a while)

To answer your question about drive letters. I hate to answer, but I have to: it's stupid. The answer should frighten you. It should horrify you. Yes it should, indeed: drive letters can be bOtH. The entire FS is case insensitive, so either is suppose to work... but string comparison and relative path calculation functions all assume case sensitivity for posix compatibility. Based on all this, the fix is to have functions that manipulate paths to understand case-insensitivity for Windows at the computeRelativePath layer in my opinion.

riccardoferretti:
Thanks for the thorough response, it's helpful (see below).

> will you be release that API v1 soon? 

The fix will be there as soon as #537 is merged, as soon as this weekend, before the release of Foam API v1.

> what's the harm in releasing something that isn't optimal [...]

Possible side effects that might become harder to trace back if we start monkey patching the code instead of addressing the issue at its core.

> To answer your question about drive letters. I hate to answer, but I have to: it's stupid [...]

Why am I not even surprised? 😄 


To summarize:
Given the rework of URI in #537 I think we should fix the issue there.
The test case has already been added, and from what you have told me:
- we can't assume upper case or lower case for drive letter, so built-in normalization is not an option
- we should no longer encounter the issue of mixing case, as that was caused by VS Code
- we can assume that a drive letter is always the same across a workspace, so using e.g. the lowercase when computing a relative path, or when doing object equality, should be fine
- I would also be comfortable to assume that we will not encounter (or support) paths that are mixed in case, so when computing the common part of two paths we can lowercase the whole string to find it

Based on the above and on your PR I will implement a fix in #537 and tag you there. 
In the meantime feel free to take a look at the PR and leave any comment that might help improve it.

kenotron:
I'll close this PR in favor of #537. Is that PR in a kind of form I can test? Also, I don't know if it's feasible, but is this (even if it's just some consuming code that uses this API along side some real file fixture) something that can be tested with an end-to-end test by the build machines in your Windows build?

PR #514, opened 2021-03-04
Backlinks Panel

riccardoferretti:
Bringing the backlink in house.
Why?
- it's fully based on the Foam model, so it provides a consistent representation of notes in the workspace
- it provides the same preview functionality for note tree items as the other Foam panels
- it also provides a preview for the full row for each reference, to get better context of the link

In the process we needed to make some changes in the model, specifically:
- all links now have a reference to their position (before just wikilinks, not direct links)
- improved support for a note having more than one link to another

Will leave a couple of comments in the code to highlight some interesting points

image

### DIFF
--- a/packages/foam-core/src/model/workspace.ts
+++ b/packages/foam-core/src/model/workspace.ts
@@ -2,7 +2,7 @@ import { diff } from 'fast-array-diff';
 import { isEqual } from 'lodash';
 import * as path from 'path';
 import { URI } from '../common/uri';
-import { Resource, NoteLink, Note } from '../model/note';
+import { Resource, NoteLink, Note, Point, Position } from '../model/note';
 import {
   computeRelativeURI,
   isSome,
@@ -10,6 +10,7 @@ import {
   parseUri,
   placeholderUri,
   isPlaceholder,
+  isSameUri,
 } from '../utils';
 import { Emitter } from '../common/event';
 import { IDisposable } from '../index';
@@ -17,6 +18,7 @@ import { IDisposable } from '../index';
 export type Connection = {
   source: URI;
   target: URI;
+  link: NoteLink;
 };
 
 export function getReferenceType(
@@ -221,12 +223,12 @@ export class FoamWorkspace implements IDisposable {
     ];
   }
 
-  public static getLinks(workspace: FoamWorkspace, uri: URI): URI[] {
-    return workspace.links[uri.path]?.map(c => c.target) ?? [];
+  public static getLinks(workspace: FoamWorkspace, uri: URI): Connection[] {
+    return workspace.links[uri.path] ?? [];
   }
 
-  public static getBacklinks(workspace: FoamWorkspace, uri: URI): URI[] {
-    return workspace.backlinks[uri.path]?.map(c => c.source) ?? [];
+  public static getBacklinks(workspace: FoamWorkspace, uri: URI): Connection[] {
+    return workspace.backlinks[uri.path] ?? [];
   }
 
   public static set(
@@ -350,7 +352,7 @@ export class FoamWorkspace implements IDisposable {
       // prettier-ignore
       resource.links.forEach(link => {
         const targetUri = FoamWorkspace.resolveLink(workspace, resource, link);
-        workspace = FoamWorkspace.connect(workspace, resource.uri, targetUri);
+        workspace = FoamWorkspace.connect(workspace, resource.uri, targetUri, link);
       });
     }
     return workspace;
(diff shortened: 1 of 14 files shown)


packages/foam-core/src/model/workspace.ts
```diff
@@ -221,12 +223,12 @@ export class FoamWorkspace implements IDisposable {
     ];
   }
 
-  public static getLinks(workspace: FoamWorkspace, uri: URI): URI[] {
-    return workspace.links[uri.path]?.map(c => c.target) ?? [];
+  public static getLinks(workspace: FoamWorkspace, uri: URI): Connection[] {
```
riccardoferretti, line 226: to support multiple links across notes, `getLinks` and `getBacklinks` no longer return the `URI`, but the full `Connection` object.
An alternative would have been for these to just return the unique URIs, and leave it to `getAllConnections` to return the fully detailed list of connections. 
I am open to POVs here

packages/foam-core/src/model/workspace.ts
```diff
@@ -465,3 +487,19 @@ export class FoamWorkspace implements IDisposable {
     return workspace;
   }
 }
+
+// TODO move these utility fns to appropriate places
```
riccardoferretti, line 491: This is not really meant to be addressed here, because I don't want to make the PR too big.
But basically we need to consolidate these utility functions, and export them in `foam-core`

packages/foam-vscode/src/features/backlinks.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    .resolveLinks(true);
+
+  const provider = new BacklinksTreeDataProvider(ws, dataStore);
+
+  beforeEach(async () => {
+    await closeEditors();
+    provider.target = undefined;
+  });
+
+  // Skipping these as still figuring out how to interact with the provider
```
riccardoferretti, line 56: still figuring out how to do this. I kept the tests because this is part of the spec, but of course they don't work now because the `provider` instance I am talking to is not the one used by VS Code (or at least I believe this is why the "targets active editor" test fails)

packages/foam-vscode/src/features/backlinks.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+  it('navigates to the document if clicking on note', async () => {
+    provider.target = noteA.uri;
+    const notes = (await provider.getChildren()) as ResourceTreeItem[];
+    expect(notes[0].command).toMatchObject({
+      command: 'vscode.open',
+      arguments: expect.arrayContaining([noteB.uri]),
+    });
+  });
+  it('navigates to document with link selection if clicking on backlink', async () => {
```
riccardoferretti, line 101: note that these are not properly e2e tests, as I am not checking the editor on the VS Code instance.
It's more checking that the data structures to generate the appropriate commands is there.

packages/foam-vscode/src/features/backlinks.ts
```diff
@@ -0,0 +1,174 @@
...
+  },
+};
+export default feature;
+
+const isBefore = (a: Position, b: Position) =>
+  a.start.column - b.start.column || a.start.line - b.start.line;
+
+export class BacklinksTreeDataProvider
+  implements vscode.TreeDataProvider<BacklinkPanelTreeItem> {
+  public target?: URI = undefined;
```
riccardoferretti, line 50: the reason we use `target` instead of checking the active editor inside here is that it makes it more easily testable, even though that means that we can't just refresh on active editor change, but also assign it to `target` (see line 35)

packages/foam-vscode/src/utils/grouped-resources-tree-data-provider.ts
```diff
@@ -248,7 +249,6 @@ class ResourceTreeItem extends vscode.TreeItem {
         arguments: [resource.uri],
       };
     }
-    this.contextValue = itemLabel;
```
riccardoferretti, line 251: I think this got here by mistake so I removed it - pinging @leonhfr and @joeltjames for confirmation that we don't need it

jmg-duarte:
I think this PR should be broken into smaller ones and converted to a tracking Issue.
This would allow you to add all the small details while keeping an overview of what addresses what

riccardoferretti:
I understand the point @jmg-duarte, although most of the changes are just function calls with different arguments, I am sure it can be handled. 
I am not planning to break this PR down, I have left comments to help navigate it, and if there are parts that seem particularly unclear I am happy to clarify (just leave a question)

packages/foam-vscode/src/utils/grouped-resources-tree-data-provider.ts
leonhfr: confirmed it's not needed

packages/foam-core/test/core.test.ts
```diff
@@ -33,7 +33,9 @@ export const createTestNote = (params: {
   definitions?: NoteLinkDefinition[];
   links?: Array<{ slug: string } | { to: string }>;
   text?: string;
+  root?: URI;
 }): Note => {
+  const root = params.root ?? URI.file('/');
```
ingalless, line 38: Be good to clear this warning!

packages/foam-vscode/src/features/backlinks.spec.ts
ingalless: In case this jogs a brainwave (probably completely wrong!) - could this be solved in vscode-jest-environment (or closely similar name)?

ingalless (approved):
Really clear @riccardoferretti. I'm not familiar with much of the graphing and core model but found the code easy to read through! Just made a couple comments - one is a nitpick and the other is potentially of use!

Approved either way 🥳

PR #524, opened 2021-03-11
Link Navigation

riccardoferretti:
Brings in Foam the resolution, decoration, and navigation of links in the editor.

Notes:
- the link is not slugified when creating the file  (not visible from the demo)
- we use snippets for a better UX

link-navigation

### DIFF
--- a/packages/foam-core/src/index.ts
+++ b/packages/foam-core/src/index.ts
@@ -20,6 +20,7 @@ import { FoamWorkspace } from './model/workspace';
 export { IDataStore, FileDataStore };
 export { ILogger };
 export { LogLevel, LogLevelThreshold, Logger, BaseLogger } from './utils/log';
+export * as uris from './utils/uri'
 export { Event, Emitter } from './common/event';
 export { FoamConfig };
 export { isSameUri, parseUri } from './utils/uri';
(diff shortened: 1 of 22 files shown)


packages/foam-core/src/index.ts
```diff
@@ -9,16 +9,19 @@ import {
...
   getTitle,
+  NoteParser,
 } from './model/note';
 import { URI } from './common/uri';
 import { FoamConfig } from './config';
 import { IDataStore, FileDataStore } from './services/datastore';
 import { ILogger } from './utils/log';
 import { IDisposable, isDisposable } from './common/lifecycle';
 import { FoamWorkspace } from './model/workspace';
+import * as uris from './utils/uri';
```
ingalless, line 20: I find it difficult to parse the difference between this and URI on line 14. The change in casing perhaps is contributing to this.

A better name doesn't come to mind - maybe `uriHelper` based on the usage later? Unless the two can be merged? Curious to hear your thoughts. I'm not strongly worried about this one, but thought it worth mentioning.

ingalless (approved):
Looks good to me! I've left one comment but happy to approve without it being discussed. Nice work @riccardoferretti

packages/foam-core/src/index.ts
riccardoferretti: the goal of `uris` is to provide a namespace for all functions/types/.. related to uris.
`URI` is not part of that because it "existed" before (just like a couple of other uri related functions)..
but yeah, I should probably just make `URI` part of `uris`, I need to do a couple of refactoring PRs

does that address the question, or was there a wider point I am missing?

ingalless: Fine by me 👍🏻  Thank you for the explanation

PR #532, opened 2021-03-15
(refactor) Use Position and Range in Foam model

riccardoferretti:
Currently the Foam model uses a 1-index position and point object, inspired by the `unified` parsing library.

This is suboptimal for the following reasons:
- we work in VS Code, which uses 0-index objects for position, it's better to do the translation once at parsing time and rely on the model later on
- 1-index is counter-intuitive in most scenarios

This PR updates the Foam Model to use `Position` and `Range` objects.


packages/foam-core/src/model/position.ts
```diff
@@ -0,0 +1,87 @@
+export interface Position {
+  line: number;
+  character: number;
+}
+
+export const create = (line: number, character: number): Position => ({
```
jevakallio, line 6: Instead of having a separate named export re-exported from index, we can do:
```ts
export default interface Position {
  line: number;
  character: number;
}

export default abstract class Position implements Position {
  static create(line: number, character: number): Position {
    return {
      line,
      character
    };
  }
}
```

Issue #523, opened 2021-03-11
Template the filepath in `openDailyNote`


movermeyer:
After working on #482 , I realize there are further steps that could be done to improve this feature.

## Limitations of the current solution

In the current setup, there are two settings that govern the filepath used in the `openDailyNote` command:

* `foam.openDailyNote.directory`: Which directory to use
* `foam.openDailyNote.filenameFormat`: Which "format" to use when making the filename. It can accept any format understood by [`dateformat`](https://www.npmjs.com/package/dateformat)
* `foam.openDailyNote.fileExtension`: Which extension to append to the filename.

The full filepath is calculated by appending all of these together. Effectively:

`${directory}/${dateFormat(filenameFormat)}/${fileExtension}`

However, under this system, you cannot do useful things like:

* Include the date in the directory path. You might want to store your daily notes in a subdirectory that contains the date. For example, I'm considering using a browser extension (like [Markdown Clipper](https://github.com/deathau/markdownload)) to download the text of the sites I visit, then store them in daily directories. Alongside those files, you may want to store your daily notes.
* Add other characters to your filename besides the timestamp. For example, perhaps you want to add a prefix to each filename `daily_note_2021-03-11`. There is no way to specify that in the current setup. (You **can** add extra characters to `filenameFormat`, but only if they don't collide with the characters already used by dateFormat (now, or anytime in the future))

## A potential new solution

I posit that there is no reason to treat these as separate segments of the filepath and indeed that the feature can be simpler, while also more powerful by allowing a single format string for the whole path.

Perhaps instead of `directory`, `filenameFormat`, and `fileExtension`, you could have a pair of settings.

1. `foam.openDailyNote.filepathFormat` would be a format string that can be an absolute or relative filepath to use. Relative paths would be relative to the current workspace (as it is today) and absolute paths would be absolute (as they are in #482). The path would be templated though, allowing a `${date}` parameter to be inserted to indicate where a data should go in the path. It would default to `${date}.md` (to maintain the same default behaviour as today)
2. `foam.openDailyNote.dateFormat` would be the date format to use when replacing the `${date}` parameter. It would default to `isoDate` (as `filenameFormat` does today)

```jsonc
// Daily note is found at ${workspace}/my/subdir/daily_note_2021-03-11.md
"foam.openDailyNote.filepathFormat": "my/subdir/daily_note_${date}.md",

// Daily note is found at /absolute/path/2021-03-11.md
[...]

PR #534, opened 2021-03-16
[Templates v2] Add a proposal for how templates should work

movermeyer:
Here's my proposal for how templates should work within Foam (as requested [here](https://github.com/foambubble/foam/issues/523#issuecomment-800351797)).

You're going to want to view the [rendered version of this file](https://github.com/movermeyer/foam/blob/mo.template_v2_proposal/docs/proposals/templates-v2.md).

@riccardoferretti, did the file have to be called `docs/specs/templates-v2.md`? I went with `docs/proposals/templates-v2.md` instead.

----

cc @ingalless @riccardoferretti

### DIFF
--- a/docs/proposals/templates-v2.md
+++ b/docs/proposals/templates-v2.md
@@ -1,0 +1,326 @@
+# Templates v2 Proposal <!-- omit in TOC -->
+
+The current capabilities of templates is limited in some important ways. This document aims to propose a design that addresses these shortcomings.
+
+**IMPORTANT: This design is merely a proposal of a design that could be implemented. It DOES NOT represent a commitment by `Foam` developers to implement the features outlined in this document. This document is merely a mechanism to facilitate discussion of a possible future direction for `Foam`.**
+
+- [Introduction](#introduction)
+- [Limitations of current templating](#limitations-of-current-templating)
+  - [Too much friction to create a new note](#too-much-friction-to-create-a-new-note)
+    - [Manual note creation (Mouse + Keyboard)](#manual-note-creation-mouse--keyboard)
+    - [Manual note creation (Keyboard)](#manual-note-creation-keyboard)
+    - [Foam missing note creation](#foam-missing-note-creation)
+    - [`Markdown Notes: New Note` (Keyboard)](#markdown-notes-new-note-keyboard)
+    - [Foam template note creation (Keyboard)](#foam-template-note-creation-keyboard)
+  - [Templating of daily notes](#templating-of-daily-notes)
+  - [Templating of filepaths](#templating-of-filepaths)
+- [Goal / Philosophy](#goal--philosophy)
+- [Proposal](#proposal)
+  - [Summary](#summary)
+  - [Add a `${title}` and `${titleSlug}` template variables](#add-a-title-and-titleslug-template-variables)
+  - [Add a `Foam: Create New Note` command and hotkey](#add-a-foam-create-new-note-command-and-hotkey)
+    - [Case 1: `.foam/templates/new-note.md` doesn't exist](#case-1-foamtemplatesnew-notemd-doesnt-exist)
+    - [Case 2: `.foam/templates/new-note.md` exists](#case-2-foamtemplatesnew-notemd-exists)
+  - [Change missing wikilinks to use the default template](#change-missing-wikilinks-to-use-the-default-template)
+  - [Add a metadata section to templates](#add-a-metadata-section-to-templates)
+    - [Example](#example)
+  - [Add a replacement for `dateFormat`](#add-a-replacement-for-dateformat)
+  - [Add support for daily note templates](#add-support-for-daily-note-templates)
+  - [Eliminate all `foam.openDailyNote` settings](#eliminate-all-foamopendailynote-settings)
+- [Summary: resulting behaviour](#summary-resulting-behaviour)
+  - [`Foam: Create New Note`](#foam-create-new-note)
+  - [`Foam: Open Daily Note`](#foam-open-daily-note)
+  - [Navigating to missing wikilinks](#navigating-to-missing-wikilinks)
+  - [`Foam: Create Note From Template`](#foam-create-note-from-template)
+- [Extensions](#extensions)
+  - [More variables in templates](#more-variables-in-templates)
+  - [`defaultFilepath`](#defaultfilepath)
+  - [Arbitrary hotkey -> template mappings?](#arbitrary-hotkey---template-mappings)
+
+## Introduction
+
+Creating of new notes in Foam is too cumbersome and slow. Despite their power, Foam templates can currently only be used in very limited scenarios.
+
+This proposal aims to address these issues by streamlining note creation and by allowing templates to be used everywhere.
+## Limitations of current templating
+
+### Too much friction to create a new note
+
+Creating new notes should an incredibly streamlined operation. There should be no friction to creating new notes.
+
+Unfortunately, all of the current methods for creating notes are cumbersome.
+
+#### Manual note creation (Mouse + Keyboard)
+
+1. Navigate to the directory where you want the note
+2. Click the new file button
+3. Provide a filename
+4. Manually enter the template contents you want
+
+#### Manual note creation (Keyboard)
+
+1. Navigate to the directory where you want the note
+2. `⌘N` to create a new file
+3. `⌘S` to save the file and give it a filename
+4. Manually enter the template contents you want
+
+#### Foam missing note creation
+
+1. Open an existing note in the directory where you want the note
+2. Use the wikilinks syntax to create a link to the title of the note you want to have
+3. Use `Ctrl+Click`/`F12` to create the new file
+4. Manually enter the template contents you want
+
+#### `Markdown Notes: New Note` (Keyboard)
+
+1. Navigate to the directory where you want the note
+2. `Shift+⌘P` to open the command pallette
+3. Type `New Note` until it appears in the list. Press `Enter/Return` to select it.
+4. Enter a title for the note
+5. Manually enter the template contents you want
+
+#### Foam template note creation (Keyboard)
+
+1. `Shift+⌘P` to open the command pallette
+2. Type `Create New Note From Template` until it appears in the list. Press `Enter/Return` to select it.
+3. Use the arrow keys (or type the template name) to select the template. Press `Enter/Return` to select it.
+4. Modify the filepath to match the desired directory + filename. Press `Enter/Return` to select it.
+
+All of these steps are far too cumbersome. And only the last one allows the use of templates.
+
+### Templating of daily notes
+
+Currently `Open Daily Note` opens an otherwise empty note, with a title defined by the `foam.openDailyNote.titleFormat` setting.
+Daily notes should be able to be fully templated as well.
+
+### Templating of filepaths
+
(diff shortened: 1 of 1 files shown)


docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+A workable syntax is still to be determined.
+While this syntax probably doesn't work as a solution, for this example I will demonstrate the concept using a second frontmatter block:
+
+```markdown
+<!-- The below front-matter block is for foam-specific template settings -->
+---
+<!-- The default filepath to use when using this template -->
+<!-- Relative paths are relative to the workspace, absolute paths are absolute -->
+<!-- Note that you can include VSCode snippet variables to template the path -->
+filepath: `journal/${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}_#{title}.md`
```
riccardoferretti, line 202: I can see two related but properties that result in different behavior:
- `filepath` will not ask the user for the file path, simply use the value provided
- `defaultFilepath` will ask the user for the file path, defaulting to the provided value

The first allows "one-click" note creation, the second more customization.

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+
+As we'll see, when it comes to allowing daily notes to be templated, we don't even need to use `${title}` in our template, in which case we don't we don't even need to prompt for a title.
+
+In the `Create New Note From Template` workflow, during the step where we allow the user to customize the filepath, it will already templated according to the `filepath` in the template's metadata. This means that the user has to make fewer changes to the path, especially in cases where they want to include things like datetimes in the filenames. This makes it faster (eg. don't have to remember what day it is, and don't have to type it) and less error-prone (eg. when they accidentally type the wrong date).
+
+### Add support for daily note templates
+
+With the above features implemented, making daily notes use templates is simple.
+
+We create a new `foam.defaultDailyNoteTemplate` setting which is the template that `Foam: Open Daily Note` will always use (defaults to `null`)
```
riccardoferretti, line 227: see my comment above about convention over configuration

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+
+Both `Foam: Open Daily Note` and `Foam: Create New Note` can share all of the implementation code, with the only differences being the hotkeys used and the default template setting used (`foam.defaultDailyNoteTemplate` vs. `foam.defaultNoteTemplate`).
```
riccardoferretti, line 229: I like the idea of hotkeys, we might want to pull this out in its own section?

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+A new command optimized for speedy creation of new notes. This will become the default way to create new notes. In its fastest form, it simply opens the new note with no further user interaction.
+
+### `Foam: Open Daily Note`
+
+Simplified since it no longer has its custom settings, and re-uses all the same implementation code as `Foam: Create New Note`.
+Templates can now be used with daily notes.
+
+### Navigating to missing wikilinks
+
+Now creates the new notes using the default template. Re-uses all the same implementation code as `Foam: Create New Note`
```
riccardoferretti, line 268: and uses the wikilink as value for `${title}`

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+    "template": "my_fancy_note_template.md"
+  }
+}
+```
+
+## Limitations of this proposal
+
+### No `dateFormat`
+
+`foam.openDailyNote.filenameFormat` uses `dateFormat()` to put the current timestamp into the daily notes filename. This is much more flexible than what is available in VSCode Snippet variables. Before daily notes are switched over to use templates, we might have to come up with another mechanism/syntax to allow for calls to `dateFormat()` within template files.
```
riccardoferretti, line 303: I agree we'd need this

docs/proposals/templates-v2.md
ingalless: Hotkeys are great - but perhaps as an addition to the `foam-template` workspace-level settings rather than within the extension.

Current users of Foam might be frustrated if they're having to _unbind_ a hotkey vs binding one to "subscribe" to new functionality"

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+
+#### More variables in templates
+
+`${title}` is necessary in this case to replace the functionality of `Markdown Notes: New Note`.
+However, one could imagine that this pattern of "Ask the user for a value for missing variable values" could be useful in other situations too.
+Perhaps users could even define their own (namespaced) template variables, and Foam would ask them for values to use for each when creating a note using a template that used those variables.
+
+#### Schedule-based templating
+
+As suggested by @ingalless [here](https://github.com/foambubble/foam/issues/523#issuecomment-799231399), it might be handy if different templates could be used depending on the day of the week.
```
ingalless, line 333: Ah - sorry! I think my example drew unnecessary confusion. You can remove this here, or summarise the below if you like.

Let's say I've configured the `foam.defaultDailyNoteTemplate` configuration. I'd probably set this to something like `work-day.md` because 5 days a week are working. For the weekend though I'd probably want a bail out - something like "Create daily note from template" or "Create blank daily note" so that I don't generate my default template.

In fact, your proposal already nicely covers this - it's just a couple extra commands around daily notes.

ingalless:
Thank you for this write up @movermeyer!

docs/proposals/templates-v2.md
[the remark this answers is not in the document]
riccardoferretti: not sure whether we should prefix them to make it very explicit or can just "reserve" some property names

docs/proposals/templates-v2.md
riccardoferretti: yeah I agree - hotkeys are great, but they need to be done in the same paradigm that VS Code uses for all others keybindings, I don't think we should have a competing system there.
(i.e. we shouldn't have a `hotkey` key in the template or template configuration, the binding should live where all VS Code bindings live)

docs/proposals/templates-v2.md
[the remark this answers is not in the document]
movermeyer: All names can be changed to whatever makes the most sense.

----

If I had to justify it, I'd say that VSCode system-defined variables use ALL CAPS and it could make sense to distinguish Foam/application-defined variables from those. 🤷 Happy to do whatever makes the most sense.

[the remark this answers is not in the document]
riccardoferretti: the question was more about `$` vs `#`, but yes the casing as well should just be consistent. If we want we could namespace the Foam variables (e.g. `${FOAM_TITLE}`)?

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,334 @@
...
+
+**Example:**
+
+Given the name `Living in a dream world` to `Markdown Notes: New Note`, the filename is `living-in-a-dream-world.md` and the file contents are:
+
+```markdown
+# Living in a dream world
+```
+
+When creating a note from a template in Foam, you should be able to use a `${title}` variable. If the template uses the `${title}` variable, the user will be prompted for a title when they create a note from a template.
```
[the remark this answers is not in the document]
movermeyer: In an earlier draft, I had defined a `${titleSlugified}` in `filepath`, since any title would have to be sanitized.
Then my thought was that the contents of the `filepath` will always need to be sanitized/normalized, so we don't need `${titleSlugified}` and the case of adding `${title}` in `filepath` would already be covered.

That said, having `${titleSlugified}` would allow us to separate the logic that does the slugify-ing from the logic that does the sanitizing, so it likely makes sense to have it. 

I'll add it back in.

(And people could use `${titleSlugified}` anywhere they liked, if there were other uses for it)

docs/proposals/templates-v2.md
movermeyer: Added a mention of this.

docs/proposals/templates-v2.md
movermeyer: > In fact, your proposal already nicely covers this

Glad to hear it. I've removed that section.

docs/proposals/templates-v2.md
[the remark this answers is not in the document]
movermeyer: > the question was more about $ vs #

Whoops. That's my Ruby leaking in... Fixed.

movermeyer: I think this could be a good extension. I've added it to the "Extensions" section.

docs/proposals/templates-v2.md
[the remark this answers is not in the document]
movermeyer: I like this idea. I've added the idea of `${titleSlug}` to the Extensions section.

docs/proposals/templates-v2.md
movermeyer: Moved this section from `Limitations` into its own implementation step.

docs/proposals/templates-v2.md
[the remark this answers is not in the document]
movermeyer: > It might also be worth considering whether we will clean up the frontmatter in the generated template.

I've added mentions throughout the doc that make it explicit that this frontmatter section wouldn't be present in any note made from this template.

> not sure whether we should prefix them to make it very explicit or can just "reserve" some property names

I think we'll likely namespace them. Instead of `${title}`, I expect they will be `${FOAM_TITLE}`.
Aside: VSCode doesn't mind variables that it doesn't recognize.

docs/proposals/templates-v2.md
movermeyer: Sounds like this is something we'll refactor/figure out as part of the implementation.

In the meantime, I've moved the section about user defined hotkeys to the "Extensions" section.

movermeyer:
@riccardoferretti @ingalless 

I'm not sure what you want the end result of this PR to be. Do we merge this? Close this?
I've started [implementation of `${title}` here](https://github.com/foambubble/foam/pull/549)

ingalless:
@riccardoferretti I was thinking it might be nice to get some tasks up on the project board based on the proposal?

cc @movermeyer

docs/proposals/templates-v2.md
```diff
@@ -0,0 +1,325 @@
...
+
+If `.foam/templates/new-note.md` doesn't exist, it behaves the same as `Markdown Notes: New Note`:
+  * it would ask for a note name and create the note in the current directory. It would open a note with the note name as the title.
+
+**Note:** this would use an implicit default template, making use of the `${title}` variable.
+
+#### Case 2: `.foam/templates/new-note.md` exists
+
+If `.foam/templates/new-note.md` exists:
+  * it asks for the note name and creates the note in the current directory
```
riccardoferretti, line 172: what do you mean here by `name`: `title` or `filename`?

riccardoferretti (approved):
This proposal looks good to me.
Let's just be clear where we are asking for titles vs filenames, this can be confusing - I left a comment in that regard

ingalless:
@riccardoferretti Good idea! @movermeyer I imagine you have access to the project board? Do you think you can split the proposal out into PR-able tasks once the proposal is fully ironed out?

docs/proposals/templates-v2.md
movermeyer: Good point! I've gone through the document and removed all uses of `name`, changing them to either `title` or `filepath` as appropriate.

movermeyer:
@riccardoferretti @ingalless I've created new issues (#560-#566) to capture the work related to this proposal.

PR #535, opened 2021-03-17
Extract foam-cli to own repo

riccardoferretti:
https://github.com/foambubble/foam-cli


jevakallio (commented):
Ding dong the witch is dead! 🥳

PR #541, opened 2021-03-19
Rename recipes.md to index.md

zomars:
Many foam-based project are linking to this page using https://foambubble.github.io/foam/recipes/. This will prevent a 404 on those sites. Right now the current URL is "https://foambubble.github.io/foam/recipes/recipes" which doesn't feel quite right IMHO.


riccardoferretti:
I prefer not to rename the file at this point. an alternative option would be to include the redirect plugin in jekyll

```
plugins:
  - jekyll-redirect-from
```

and set a redirect in the recipe:

```
redirect_from:
  - /recipes
``` 

closing this PR, feel free to open another one with the approach above

PR #550, opened 2021-03-28
[TRIVIAL] Add a placeholder for the template file quick pick menu

movermeyer:
Slightly nicer UX when using `Create New Note From Template`.

The first step of  `Create New Note From Template` is selecting a template file to use.

**Before:**

image

**After:**

image


riccardoferretti (approved):
really good idea

Issue #561, opened 2021-04-02
[Templates v2] Add a ${FOAM_TITLE_SLUG} template parameter


movermeyer:
As described in #534, add a `${FOAM_TITLE_SLUG}` parameter that can be used in templates.

When the template uses the `${FOAM_TITLE_SLUG}` parameter, the value of [`${FOAM_TITLE}`](https://github.com/foambubble/foam/issues/560) is "slugified". If the value of `${FOAM_TITLE}` has not been provided, the user is prompted to provide a title during the `Foam: Create New Note From Template` command.

`${FOAM_TITLE_SLUG}` is meant to be a filesystem compatible version of `${FOAM_TITLE}`, for use in filepaths.
The exact algorithm is up for debate, but likely needs to follow the algorithm used when creating filepaths from Wikilinks.

### Example

1. A user starts the `Foam: Create New Note From Template` and chooses a template that references the `${FOAM_TITLE_SLUG}` parameter.
2. Since the value of `${FOAM_TITLE_SLUG}` is derived from the value of `${FOAM_TITLE}` and a title has not been supplied, the user is prompted to provide a title
3. The user provides a title (eg. `Living in a dream world`) which is then slugified into `living-in-a-dream-world` and assigned to `${FOAM_TITLE_SLUG}`
4. `${FOAM_TITLE_SLUG}` within the template is replaced by the value.

movermeyer:
As discussed in https://github.com/foambubble/foam/pull/569#discussion_r611091308, this is not the direction we want to go.

We don't want to use "Markdown Note"-style filenames, we're going to use Obsidian (ie. no slugification) filenames

PR #569, opened 2021-04-05
Add FOAM_TITLE_SLUG template variable

movermeyer:
Implements #561

I have refactored the code in #549. **This should not be merged until that one is merged, AND this one is rebased.**

The top level command cna pass in known values for each of the variables, which will be useful in a future where clicking a Wikilink provides the value of `FOAM_TITLE`.

Then I added `FOAM_TITLE_SLUG`, which required additional refactoring. This is because it depends on the value of `FOAM_TITLE`, which may or may not be present in the template, and may or may not be already resolved.

# ~Open~ question

**Update: I'm going to leave the existing behaviour for now: we won't ask for the title unless the template needs it.**

Should we always resolve `FOAM_TITLE_SLUG` (regardless of whether the template asks for it)?

**Benefit:** We always have it available to populate the default filepath
**Detriment:** We always ask the user for it when using `Foam: Create New Note From Template` (ie. 1 more step). If they use either `FOAM_TITLE_SLUG` or `FOAM_TITLE`, there won't be any additional steps.

### DIFF
--- a/packages/foam-vscode/src/features/create-from-template.ts
+++ b/packages/foam-vscode/src/features/create-from-template.ts
@@ -6,6 +6,7 @@ import {
   SnippetString,
   Uri,
 } from 'vscode';
+import GithubSlugger from 'github-slugger';
 import * as path from 'path';
 import { FoamFeature } from '../types';
 import { TextEncoder } from 'util';
@@ -63,7 +64,7 @@ function findFoamVariables(templateText: string): string[] {
   return uniqVariables;
 }
 
-function getFoamTitle() {
+function resolveFoamTitle(givenValues: Map<string, string>) {
   return window.showInputBox({
     prompt: `Enter a title for the new note`,
     value: 'Title of my New Note',
@@ -72,33 +73,55 @@ function getFoamTitle() {
   });
 }
 
-function resolveFoamVariable(
-  variable: string,
+async function resolveFoamTitleSlug(
+  resolver: Resolver,
   givenValues: Map<string, string>
 ) {
-  if (variable === 'FOAM_TITLE') {
-    return getFoamTitle();
-  } else {
-    return Promise.resolve(variable);
+  const slugger = new GithubSlugger();
+  const foamTitle = await resolver.resolve('FOAM_TITLE', givenValues);
+  return Promise.resolve(slugger.slug(foamTitle));
+}
+
+class Resolver {
+  promises = new Map<string, Thenable<string>>();
+
+  resolve(name: string, givenValues: Map<string, string>): Thenable<string> {
+    if (givenValues.has(name)) {
+      this.promises.set(name, Promise.resolve(givenValues.get(name)));
+    } else if (!this.promises.has(name)) {
+      switch (name) {
+        case 'FOAM_TITLE':
+          this.promises.set(name, resolveFoamTitle(givenValues));
+          break;
+        case 'FOAM_TITLE_SLUG':
+          this.promises.set(name, resolveFoamTitleSlug(this, givenValues));
+          break;
+        default:
+          this.promises.set(name, Promise.resolve(name));
+          break;
+      }
+    }
+    const result = this.promises.get(name);
+    return result;
   }
 }
 
-async function resolveFoamVariables(
+export async function resolveFoamVariables(
   variables: string[],
   givenValues: Map<string, string>
 ) {
+  const resolver = new Resolver();
   const promises = variables.map(async variable =>
-    Promise.resolve([
-      variable,
-      await resolveFoamVariable(variable, givenValues),
-    ])
+    Promise.resolve([variable, await resolver.resolve(variable, givenValues)])
   );
 
   const results = await Promise.all(promises);
+
   const valueByName = new Map<string, string>();
-  results.forEach(result => {
-    valueByName.set(result[0], result[1]);
+  results.forEach(([variable, value]) => {
+    valueByName.set(variable, value);
   });
+
   return valueByName;
 }
 
@@ -107,8 +130,11 @@ export function substituteFoamVariables(
   givenValues: Map<string, string>
 ) {
   givenValues.forEach((value, variable) => {
-    const regex = new RegExp(`\\\${${variable}}|\\$${variable}`, 'g');
-    templateText = templateText.replace(regex, value);
+    const regex = new RegExp(
+      `\\\${${variable}}|\\$${variable}([^_a-zA-Z0-9]|$)`,
+      'g'
+    );
+    templateText = templateText.replace(regex, `${value}$1`);
   });
 
   return templateText;
@@ -137,14 +163,25 @@ async function createNoteFromTemplate(): Promise<void> {
 
   const givenValues = new Map<string, string>();
   const variables = findFoamVariables(templateText.toString());
-  const results = await resolveFoamVariables(variables, givenValues);
-  const subbedText = await substituteFoamVariables(
+
+  // Always resolve FOAM_TITLE_SLUG if we have FOAM_TITLE
+  // We can use it to pre-populate the filepath.
+  if (
+    (variables.includes('FOAM_TITLE') || givenValues.has('FOAM_TITLE')) &&
+    !variables.includes('FOAM_TITLE_SLUG')
+  ) {
+    variables.push('FOAM_TITLE_SLUG');
+  }
+
+  const resolvedValues = await resolveFoamVariables(variables, givenValues);
+  const subbedText = substituteFoamVariables(
     templateText.toString(),
-    results
+    resolvedValues
   );
   const snippet = new SnippetString(subbedText);
 
-  const defaultFileName = 'new-note.md';
+  const defaultSlug = resolvedValues.get('FOAM_TITLE_SLUG') || 'New Note';
+  const defaultFileName = `${defaultSlug}.md`;
   const defaultDir = Uri.joinPath(currentDir, defaultFileName);
   const filename = await window.showInputBox({
     prompt: `Enter the filename for the new note`,
--- a/packages/foam-vscode/src/features/create-from-template.ts
+++ b/packages/foam-vscode/src/features/create-from-template.ts
@@ -72,7 +72,10 @@ function getFoamTitle() {
   });
 }
 
-function resolveFoamVariable(variable: string) {
+function resolveFoamVariable(
+  variable: string,
+  givenValues: Map<string, string>
+) {
   if (variable === 'FOAM_TITLE') {
     return getFoamTitle();
   } else {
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -52,6 +52,105 @@ async function offerToCreateTemplate(): Promise<void> {
...
+      prompt: `Enter a title for the new note`,
+      value: 'Title of my New Note',
+      validateInput: value =>
+        value.trim().length === 0 ? 'Please enter a title' : undefined,
+    });
+  }
+}
+
+function slugify(input: string) {
+  // TODO: Replace this with the actual algorithm.
```
riccardoferretti, line 80: you can use `github-slugger` (we already have the dependency)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -52,6 +52,105 @@ async function offerToCreateTemplate(): Promise<void> {
...
+  return input.toLowerCase().replace(/[^\p{L}\p{N}]/gu, '-');
+}
+
+async function getFoamTitleSlug(
+  resolver: Resolver,
+  givenValues: Map<string, string>
+) {
+  if (givenValues.has('FOAM_TITLE_SLUG')) {
```
riccardoferretti, line 88: mmmm... why not always derive it from the `FOAM_TITLE`?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -67,7 +166,30 @@ async function createNoteFromTemplate(): Promise<void> {
...
 
-  const defaultFileName = 'new-note.md';
+  const templateText = await workspace.fs.readFile(
+    URI.joinPath(templatesDir, selectedTemplate)
+  );
+
+  const givenValues = new Map<string, string>();
+  const variables = findFoamVariables(templateText.toString());
+
+  // TODO: Always ask for FOAM_TITLE_SLUG?
```
riccardoferretti, line 176: I take this is related to the open question in the PR.
if the template uses `FOAM_TITLE` or `FOAM_TITLE_SLUG`, we ask for the title of the note, and use that as base for both variables (that is, we wouldn't prompt for the slug, and we wouldn't prompt for the title if neither it nor its slug is required).
If no default path property is provided via the template, we default the path to current file dir + `New Note.md`, but in that case we always use that only as default (that is, we still prompt the user to confirm the final filepath).

Keep in mind that in Foam we don't resolve the title to its slugified filename, so our default behavior needs to match that so that wikilinks don't need to be slugified.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -52,6 +52,105 @@ async function offerToCreateTemplate(): Promise<void> {
...
+
+  resolve(name: string, givenValues: Map<string, string>): Thenable<string> {
+    if (!this.promises.has(name)) {
+      switch (name) {
+        case 'FOAM_TITLE': {
+          this.promises.set(name, getFoamTitle(givenValues));
+          break;
+        }
+        case 'FOAM_TITLE_SLUG': {
+          this.promises.set(name, getFoamTitleSlug(this, givenValues));
```
riccardoferretti, line 106: have you considered explicitly making this `getFoamTitle` + slugification?

riccardoferretti (commented):
I left some initial thoughts, let me know what you think

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Is the question whether to have a `getFoamTitleSlug` or inline the method here?

I like the symmetry/encapusulation of each foam variable having its implementation in its own method. 🤷 

---

(Aside: Part of me doesn't like the aesthetic of this switch statement at all and wants to remove it, but it hasn't made sense yet. Perhaps one day we'll be able to do something more dynamic/eval-y to make the code shorter/nicer: `get${variable}.call(this, givenValues)` or whatever the syntax is in JS.)

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: For symmetry of implementation between all the variable resolver methods.

In fact, now that my code is cleaner than what I previously had, I had planned to pull this check up into the `resolve` method. Something like:

```javascript
if (givenValues.has(name)) {
  return Promise.resolve(givenValues.get(name))
} else {
  switch(name) {
     ...
  }
}
```

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Wonderful! I've switched it to use that. 🎉

packages/foam-vscode/src/features/create-from-template.ts
riccardoferretti: I see as semantically different "getting" a variable from "derive a variable from another one", maybe that's why, although i agree there can be symmetry of implementation, I am not sure there is a symmetry of semantics.
E.g. imagine if when we create a new note from a wikilink we are automatically providing the `FOAM_TITLE` variable, but not the `FOAM_TITLE_SLUG` (we shouldn't and if we did it shouldn't be used); separating the two semantics would make this distinction very clear

In the end this is partially styling, I don't feel strongly about it, and I think the current approach is good, so I am happy to move forward with it.

Regarding your point about it being more dynamic, you could provide  have a map:

```
const resolvers = {
  FOAM_TITLE: getFoamTitle
  FOAM_TITLE_SLUG: getFoamTitleSlug
}
```

then check if the variable exists in the resolver table. Of course all functions would need to have the same signature but that sounds reasonable enough.
Connecting this to my previous point about symmetry, the map makes sense to me from an implementation POV, but to highlight the difference in semantics of the various methods I might name them differently, e.g. `getFoamTitle` and `computeFoamTitleSlug` 

Food for thoughts

movermeyer: It might be pedantic, but:

* the code as it stands doesn't just "derive a variable from another one". It "resolves the value of `FOAM_TITLE`, **then** resolves a variable from it"
* IMO, the caller shouldn't care how the value is resolved, just that it is resolved. 

I've kept the naming generic, but changed `get` to `resolve` (ie. `resolveFoamTitle` and `resolveFoamTitleSlug`).

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Done.

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: I've left the behaviour as is, for the case where `FOAM_TITLE` is not provided nor requested by the template.

---

> Keep in mind that in Foam we don't resolve the title to its slugified filename, so our default behavior needs to match that so that wikilinks don't need to be slugified.

I had some difficulty parsing this sentence.

Is the concern that we won't be able to derive `FOAM_TITLE_SLUG` from `FOAM_TITLE` in a way that matches the way that the current wikilinks implementation does? In a future where clicking a wikilink passes in the value of `FOAM_TITLE`, we won't end up producing the same filename that would be produced today?

riccardoferretti: What I mean here is that the wikilink `[[This is my title]]` looks for a file called `This is my title.md` (like Obsidian), not `this-is-my-title.md` (like Markdown Notes).

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -52,6 +53,96 @@ async function offerToCreateTemplate(): Promise<void> {
...
+
+class Resolver {
+  promises = new Map<string, Thenable<string>>();
+
+  resolve(name: string, givenValues: Map<string, string>): Thenable<string> {
+    if (givenValues.has(name)) {
+      this.promises.set(name, Promise.resolve(givenValues.get(name)));
+    } else if (!this.promises.has(name)) {
+      switch (name) {
+        case 'FOAM_TITLE': {
```
riccardoferretti, line 93: minor: throughout the codebase our style for `switch` is to only have `{` at the `switch` level, and not for each individual `case` (TIL that's possible).
I am surprised the linter didn't pick this up 🤔

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -69,7 +160,31 @@ async function createNoteFromTemplate(): Promise<void> {
...
+  }
+
+  const resolvedValues = await resolveFoamVariables(variables, givenValues);
+  const subbedText = substituteFoamVariables(
+    templateText.toString(),
+    resolvedValues
+  );
+  const snippet = new SnippetString(subbedText);
+
+  const defaultSlug = resolvedValues.get('FOAM_TITLE_SLUG') || 'new-note';
```
riccardoferretti, line 186: this should use the title, not the slug

```
const defaultFilename = resolvedValues.get('FOAM_TITLE') || 'New Note'
```

riccardoferretti: This is what I was referring to here: https://github.com/foambubble/foam/pull/569#discussion_r611087416

riccardoferretti (commented):
there are some conflicts with master, but the code looks good otherwise

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Some thoughts....

What is the preferred slug format then? `Obsidian` or `Markdown Notes`?

1. It sound like the `Obsidian` format is preferable, since otherwise it would be a breaking change when we change `createNoteFromPlacehoder` to use `createNoteFromTemplate` (and pass in the title).
    * Alternatively, if the `Markdown Notes` format is preferable, `Foam: Open Resource` could just be changed to look in both locations, which maintains compatibility.  
2. The existing code was already defaulting to `new-note`. Was that a mistake?
3. If there isn't any slugifying work to be done (`Obsidian`), is there still value in having `FOAM_TITLE_SLUG` at all?

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Good catch. Fixed.

riccardoferretti (approved):
LGTM

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: @riccardoferretti I think we need to answer these before we can merge this? ☝️

riccardoferretti: ooops, totally missed this

The answer is Obsidian.

1. Obsidian
2. the code was created before the migration to the Obsidian format (for historic reasons before we were using the markdown notes semantics)
3. It could still be used, because someone might want to create `A new note` that results in `a-new-note.md` (if the slug is used for the file path), aware that links to this file will look like `[[a-new-note]]`. Possibly an edge case, but that's basically the semantics of Markdown Notes (except for the link part) - thoughts?

movermeyer: > Possibly an edge case... - thoughts?

@riccardoferretti If we want Foam to move towards using Obsidian style links everywhere, let's not add edge case support for this other format.

I'll rework this PR a bit to remove any reference to `FOAM_TITLE_SLUG`. I still like the refactor of the code to use a `Resolver`, and I've based my next work on it.

movermeyer: https://github.com/foambubble/foam/pull/586

movermeyer:
Closing in favour of https://github.com/foambubble/foam/pull/586, after discussion [here](https://github.com/foambubble/foam/pull/569#discussion_r611091308)

Issue #562, opened 2021-04-02
[Templates v2] Create a `Foam: Create New Note` command


movermeyer:
As described in #534, create a `Foam: Create New Note` command which is a streamlined version of the `Foam: Create New Note From Template` command.

After this first implementation, it will reduce the number of steps required to create a new note (relative to `Foam: Create New Note From Template`) by defaulting to use the template found at `.foam/templates/new-note.md` (or an implicit template if `.foam/templates/new-note.md` does not exist).

### Future steps

As described in #534, as the rest of `Templates v2` gets implemented, `Foam: Create New Note` will become more and more streamlined, earning its position as the default way to create new notes in Foam.

PR #586, opened 2021-04-17
Template variable refactor

movermeyer:
Taking the good bits from the abandoned [`FOAM_TITLE_SLUG`](https://github.com/foambubble/foam/pull/569) work.

I'll be using this as the basis of my work on #562.

Ref: https://github.com/foambubble/foam/pull/569#discussion_r615299054

### User impact

The only user-facing changes are related to the default filename used in the filepath confirmation step of `Foam: Create New Note From Template`:

* The default filename is changed from `new-note.md` (Markdown Notes format) to `New Note.md` (Obsidian format)
*  If available, it will use `FOAM_TITLE` as the default filename

cc @riccardoferretti

### DIFF
--- a/packages/foam-vscode/src/features/create-from-template.ts
+++ b/packages/foam-vscode/src/features/create-from-template.ts
@@ -127,7 +127,7 @@ async function createNoteFromTemplate(): Promise<void> {
   const subbedText = await substituteFoamVariables(templateText.toString());
   const snippet = new SnippetString(subbedText);
 
-  const defaultFileName = 'new-note.md';
+  const defaultFileName = 'New Note.md';
   const defaultDir = Uri.joinPath(currentDir, defaultFileName);
   const filename = await window.showInputBox({
     prompt: `Enter the filename for the new note`,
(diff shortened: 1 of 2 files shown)


packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -63,42 +63,64 @@ function findFoamVariables(templateText: string): string[] {
...
+  return valueByName;
+}
+
+export function substituteFoamVariables(
+  templateText: string,
+  givenValues: Map<string, string>
+) {
+  givenValues.forEach((value, variable) => {
+    const regex = new RegExp(
+      `\\\${${variable}}|\\$${variable}([^_a-zA-Z0-9]|$)`,
```
riccardoferretti, line 120: minor: I think I understand what you are doing here, I wonder if there is a better way to encode it

I would at least leave a comment to improve comprehension 

cc @jmg-duarte as he is a regex lover (and I hate them ;) )

riccardoferretti (approved):
LGTM, I left a minor comment, will merge the PR once a comment is added

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Sure thing. I've simplified the character class I copied from the [snippet syntax](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_snippet-syntax), since it exactly matched `\W`, and added a comment that hopefully helps explain it.

riccardoferretti (approved):
looking great!

Issue #591, opened 2021-04-19
Running the contributed command: 'foam-vscode.open-daily-note' failed.


emanueledf:
- Foam version: 0.13.0
- Platform: Windows 
- Issue occurs on the [foam template](https://github.com/foambubble/foam-template) repo: Yes 

**Summary**
open daily note command fails

**Steps to reproduce**
1. execute command open daily note

**Additional information**
I restored previous version 0.12.1 and it works correctly

faydanube:
![image](https://user-images.githubusercontent.com/8817046/115330428-975d4200-a1c6-11eb-897e-b1ba58fce57e.png)
I'm having the same problem

eduardado:
Can confirm. Also thanks @emanueledf, going back to the previous version 0.12.1, solved it.
Just in case somebody doesn't know how to go back to a previous version on a VScode extension:
![imagen](https://user-images.githubusercontent.com/41479143/115341491-075ccf80-a1a9-11eb-806d-57e7c37f8ed6.png)

RobinKing:
Something strange happened:
When I use Foam under Windows, the same problem happened in v0.13.0.
And when I tried to locate the problem in my development environment (Ubuntu 20.04 in WSL2), everything works fine. . .

moorebi:
I experienced the same issue on my install:

- Version:  0.13.0
- OS: Windows 10

**Additional Issue Noted**

- One additional point is that my `auto create daily note` did not work tried couple of times by closing and re-opening the app and didn't auto create.

RobinKing:
I think I have located the problem. This problem should only happen on OS: Windows.

The specific questions are as follows:

1. when using the'foam-vscode.open-daily-note' command to open an existing daily note, everything works fine;
2. but when using the'foam-vscode.open-daily-note' command to create a new daily note, the problem will be caused by an error in the mkdir command.

The specific log is as follows:

```[error] Error: UNKNOWN: unknown error, mkdir'\\e:\my_lib_path\tests_with_template\journal' foam-vscode.open-daily-note```

I think the problem is the `\\` before the path.

I guess some code in `packages/foam-vscode/src/dated-notes.ts` or 'packages/foam-core/src/model/uri.ts' may related to this bug.

Since I don't have a windows development environment, it may take some time to solve and verify this problem by me...

riccardoferretti:
I confirm I can't replicate on a Mac.

@RobinKing where did you find that log statement without a Win environment?

[...]

PR #594, opened 2021-04-21
fix:(#591) 'foam-vscode.open-daily-note' error on Windows

RobinKing:
* when calling URI.file more than two time on windows
* a extra slash('/') at path's beginning may cause some problems
* so add a condition to solve it


riccardoferretti (approved):

riccardoferretti:
thx @RobinKing!

PR #595, opened 2021-04-23
API v1 - Resource refactoring

riccardoferretti:
Note: base branch is the refactored `FoamWorkspace` / `FoamGraph` API

Goals:
- simplify the `Resource` interface
- introduce `ResourceProvider` for extensibility 

Tasks:
- [x] consolidate `Resource` interface, remove `Note` `Attachment` `Placeholder` types
- [x] move into `Resource` abstract class the related utility functions
- [x] implement `MarkdownProvider` as a `ResourceProvider`
- [x] move into `ResourceProvider` resource specific implementation details (in `foam-core`)
- [x] move into `FoamResourceProvider` resource specific implementation details (in `foam-vscode`)
- [ ] chore: rename variables/methods `note` -> `resource`


packages/foam-core/src/model/provider.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Resource } from './note';
+import { FoamWorkspace } from './workspace';
+
+export interface ResourceProvider {
+  match: (uri: URI) => boolean;
+  fetch: (uri: URI) => Promise<Resource>;
+  resolveLink: (
+    workspace: FoamWorkspace,
+    resource: Resource,
+    link: NoteLink
```
jevakallio, line 11: Should this be ResourceLink?

packages/foam-vscode/src/features/janitor.ts
```diff
@@ -71,7 +70,7 @@ async function janitor(foam: Foam) {
 }
 
 async function runJanitor(foam: Foam) {
-  const notes: Note[] = foam.workspace.list().filter(isNote);
+  const notes: Resource[] = foam.workspace.list();
```
jevakallio, line 73: If we already have attachment resources in the workspace, I bet the janitor doesn't know what to do with them

packages/foam-vscode/src/services/provider.ts
```diff
@@ -0,0 +1,6 @@
+import { Resource, ResourceProvider } from 'foam-core';
+
+export interface FoamResourceProvider extends ResourceProvider {
+  getNoteTooltip(resource: Resource): string;
```
jevakallio, line 4: What's the thinking behind having these display-specific methods on the resource provider as opposed to just utility functions? 

Is the FoamResourceProvider going to have multiple different implementations that need different implementations of these methods?

packages/foam-core/src/model/provider.ts
riccardoferretti: yes

packages/foam-vscode/src/features/janitor.ts
riccardoferretti: what's the janitor?

... sorry I couldn't resist hehe

but yes, here we would probably need to filter by type (it's a bit of a coupling, but md is still the "native" resource in Foam, so I think it's fine - I don't think it's worth changing the provider api to make the janitor happy)

PR #601, opened 2021-04-25
Add streamline "Create new note" command

movermeyer:
Closes #562

This PR creates a new `Foam: Create New Note` command that serves as a streamlined version of `Foam: Create New Note From Template`, as described in #534.

### Open questions

I have used the term "Default template" to refer to this special `new-note.md` template.
I'm not 100% satisfied with this name, but didn't come up with a better name. Perhaps you can?

### Next steps

As we implement the rest of the steps of #534, this command will become more and more streamlined, in many cases, it won't have to ask for any user input at all.

PR #622, opened 2021-05-09
Ignore unknown Foam variables

movermeyer:
When testing #601, I changed my mind on how we should be handling unknown Foam variables.

Generally, if you don't know what something is, you shouldn't mess with it.

Previously it would replace `$FOAM_FOO` with `FOAM_FOO` before passing it VSCode.
Now it just leaves the unknown variables alone and passes them to VSCode unmodified.

### DIFF
--- a/packages/foam-vscode/CHANGELOG.md
+++ b/packages/foam-vscode/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to the "foam-vscode" extension will be documented in this fi
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
+## Unreleased
+
+- Unknown Foam template variables are now ignored by Foam.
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to the "foam-vscode" extension will be documented in this fi
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
+## Unreleased
+
+- Unknown Foam template variables are now ignored by Foam.
```
riccardoferretti, line 9: I like this, I wonder whether it might become source of conflicts when rebasing/merging feature branches, but I like the approach otherwise

riccardoferretti (approved):
LGTM

PR #623, opened 2021-05-09
Handle users cancelling "Create New Note" commands

movermeyer:
As part of the refactoring done in #601, I broke the cancellation flow.

### 2 cases

When resolving `$FOAM_TITLE`, if the user cancels (i.e. hits `Esc` key), then the command should cancel.
Instead it was setting `FOAM_TITLE` to `undefined` and continuing onto the next steps, ultimately producing a new note with `undefined` in it. (The problem is not that the note has `undefined` in it, the problem is that cancelling the command didn't stop the command from continuing) 

image

Surely, there has to be a better (built-in?) pattern for a user cancelling an operation while deep in a stack.
Is there some built-in Exception I should be using instead? Some other related pattern?

### DIFF
--- a/packages/foam-vscode/CHANGELOG.md
+++ b/packages/foam-vscode/CHANGELOG.md
@@ -8,7 +8,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 - Unknown Foam template variables are now ignored by Foam.
   - e.g. before `$FOAM_FOO` would become `FOAM_FOO`, now it is passed to VSCode which turns it into an empty string
-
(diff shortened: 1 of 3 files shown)


riccardoferretti:
Thanks @movermeyer 
Can you please add a test case that would have failed before and works now?
I thought there was already a test of the interrupted flow, I am unsure as to why it didn't fail, and what we need to do to improve it

movermeyer:
>Can you please add a test case that would have failed before and works now?

Done.

> I thought there was already a test of the interrupted flow, I am unsure as to why it didn't fail, and what we need to do to improve it

There is a test for an interrupted `Foam: New Template`, but not one for `Foam: Create New Note` nor `Foam: Create New Note From Template`

PR #625, opened 2021-05-10
Add related projects section to readme

qpwo:
Not sure if you'll want this but could be of service to potential users. Or maybe could be in a separate file with comparisons to foam?


riccardoferretti:
Thanks for proposing adding the list of similar projects, but I would rather not include it here, mostly because it's another list that would need to be maintained and I don't feel it's a good use of maintainers' time. Plus what belongs to the list is blurry, and I can quickly see it getting out of hand or needing to have conversations as to why this project belongs and this other doesn't.
Finally, there are plenty of resources that would be more up to date and in greater detail (e.g. feature comparison) out there.

PR #642, opened 2021-05-18
Upgrade force-graph to 1.40.5

### LINKED ISSUE #347: ["BUG"] Foam Graph uses inappropriate amount of CPU
**Describe the bug**

As soon as I open the Foam Graph two electron processes starts hogging a questionable high amount of CPU performance. On a workspace with around a dozen notes(i.e. nodes in the graph) they are both between 10% and 15% while on a workspace with ~330 notes one is at roughly 65-75% and the other at a quite stable ~24-26%. Numbers according to `htop` and on an AMD Ryzen 5 3600

This is without any interaction with the graph and even if VS Code is on a workspace that isn't even being displayed

**To Reproduce**
1. Check a process manager like `htop` for the current list of processes requiring the most CPU performance
2. Open VS Code with your Foam Workspace (which should not yet show a CPU load spike)
3. Open the Foam Graph and probably observe two new processes hogging roughly an entire CPU core between them

**Expected behavior**

Not hog an absurdly high amount of CPU performance. The Markdown Links graph requires a similar amount of CPU when opening it and occasionally during panning, but this returns to baseline shortly after again.

**Additional Notes**

I assume this is something with the interactivity and live update feature. Potentially this could be deactivated via an option until the exact cause of this issue is found.

jmg-duarte:
According to https://github.com/vasturiano/force-graph/issues/169, this should fix #347


jmg-duarte:
We can further apply the changes described in the force-graph issue to try and reduce performance issues further

riccardoferretti (approved):
thanks for spotting the library upgrade! LGTM

Issue #614, opened 2021-05-05
Add support for nested tags


huy-vuong:
### Context

Other note-taking setups like with [Bear](https://bear.app/) and [Notable](https://notable.app/) have the ability to nest tags such as `#ideas/foam` to indicate "ideas specifically for Foam".

The nesting is denoted by the `/` character, like a directory, so this should be able to work in either `#hashtag` form or in the YAML frontmatter.

This way, it'd be easier to group similar ideas together and instead of just searching for "ideas", you could search for "ideas specifically for Foam" for instance.

The **Tag Explorer** might look something like this:

- #️⃣ ideas
    - #️⃣ foam
        - 📝 Add support for nested tags

### Current Behaviour

- In the **Tag Explorer**:
    - `#ideas/foam` gets treated as just having the tag `#ideas`.
    - Adding the tag `ideas/foam` to the frontmatter gets treated as just having the tag `#ideas`.

### Desired Behaviour

- In the **Tag Explorer**:
    - `#ideas/foam` gets treated as having the tag `#ideas`.
    - `#ideas/foam` also gets treated as having the tag `#ideas/foam`, which is nested inside `#ideas`.
    - Adding the tag `ideas/foam` to the frontmatter gets treated as just having the tag `#ideas`.
    - Adding the tag `ideas/foam` to the frontmatter also gets treated as just having the tag `#ideas/foam`, which is nested inside `#ideas`.

riccardoferretti:
I like the idea, happy to provide guidance if someone wants to take it on

pderaaij:
Part of release 0.13.4. Please try it out and leave feedback, if any.

riccardoferretti:
this is now in 0.13.4, so I am closing the ticket. If anything is amiss, feel free to further comment

PR #643, opened 2021-05-19
Add support for nested tags

pderaaij:
This is an attempt to implement nested tags like described in #614.

My coding skills are a bit rusty, so feel free to have a thorough look at it.

The changes ensure that forward slashes are accepted in tags. The tag tree view explorer is adjusted to display parent/child relations in tags, and respect a logical sorting algorithm that puts childs on top, and references after those.
Screenshot 2021-05-19 at 13 32 07

I've also made a minor to change to the call to the search dialog. It now starts as a RegEx search to make the Pound signal optional. This ensures we also find the tags jotted down in the Frontmatter.

Screenshot 2021-05-19 at 13 32 29

### DIFF
--- a/packages/foam-core/src/utils/hashtags.ts
+++ b/packages/foam-core/src/utils/hashtags.ts
@@ -1,6 +1,6 @@
 import { isSome } from './core';
-const HASHTAG_REGEX = /(^|\s)#([0-9]*[\p{L}_-][\p{L}\p{N}_-]*)/gmu;
-const WORD_REGEX = /(^|\s)([0-9]*[\p{L}_-][\p{L}\p{N}_-]*)/gmu;
+const HASHTAG_REGEX = /(^|\s)#([0-9]*[\p{L}\/_-][\p{L}\p{N}\/_-]*)/gmu;
+const WORD_REGEX = /(^|\s)([0-9]*[\p{L}\/_-][\p{L}\p{N}\/_-]*)/gmu;
 
 export const extractHashtags = (text: string): Set<string> => {
   return isSome(text)
--- a/packages/foam-core/src/utils/hashtags.ts
+++ b/packages/foam-core/src/utils/hashtags.ts
@@ -1,6 +1,6 @@
 import { isSome } from './core';
-const HASHTAG_REGEX = /(^|\s)#([0-9]*[\p{L}\/_-][\p{L}\p{N}\/_-]*)/gmu;
-const WORD_REGEX = /(^|\s)([0-9]*[\p{L}\/_-][\p{L}\p{N}\/_-]*)/gmu;
+const HASHTAG_REGEX = /(^|\s)#([0-9]*[\p{L}/_-][\p{L}\p{N}/_-]*)/gmu;
+const WORD_REGEX = /(^|\s)([0-9]*[\p{L}/_-][\p{L}\p{N}/_-]*)/gmu;
 
 export const extractHashtags = (text: string): Set<string> => {
   return isSome(text)
--- a/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
+++ b/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
@@ -36,7 +36,7 @@ describe('Tags tree panel', () => {
   });
 
   afterAll(async () => {
-    _foam.workspace.dispose();
+    _foam.dispose();
     await cleanWorkspace();
   });
 
@@ -66,18 +66,15 @@ describe('Tags tree panel', () => {
     provider.refresh();
 
     const parentTreeItems = (await provider.getChildren()) as Tag[];
-    const parentTagItem = parentTreeItems.filter(item => item instanceof Tag);
-    expect(parentTagItem[0].title).toEqual('parent');
-    expect(parentTagItem[0].title).not.toEqual('child');
+    const parentTagItem = parentTreeItems.pop();
+    expect(parentTagItem.title).toEqual('parent');
+    expect(parentTagItem.title).not.toEqual('child');
 
-    const childTreeItems = (await provider.getChildren(
-      parentTagItem[0]
-    )) as Tag[];
+    const childTreeItems = (await provider.getChildren(parentTagItem)) as Tag[];
 
     childTreeItems.map(child => {
       if (child instanceof Tag) {
         expect(child.title).toEqual('child');
-        expect(child.title).not.toEqual('parent');
       }
     });
   });
@@ -101,7 +98,6 @@ describe('Tags tree panel', () => {
     )[0];
 
     expect(parentTagItem.title).toEqual('parent');
-    expect(parentTagItem.title).not.toEqual('child');
 
     const childTreeItems = (await provider.getChildren(parentTagItem)) as Tag[];
 
--- a/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
+++ b/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
@@ -1,0 +1,115 @@
+import {
+  cleanWorkspace,
+  closeEditors,
+  createTestNote,
+} from '../../test/test-utils';
+
+import { Tag, TagsProvider } from '.';
+
+import {
+  bootstrap,
+  createConfigFromFolders,
+  Foam,
+  FileDataStore,
+  FoamConfig,
+  MarkdownResourceProvider,
+  Matcher,
+} from 'foam-core';
+
+describe('Tags tree panel', () => {
+  let _foam: Foam;
+  let provider: TagsProvider;
+
+  beforeAll(async () => {
+    await cleanWorkspace();
+
+    const config: FoamConfig = createConfigFromFolders([]);
+    const mdProvider = new MarkdownResourceProvider(
+      new Matcher(
+        config.workspaceFolders,
+        config.includeGlobs,
+        config.ignoreGlobs
+      )
+    );
+    _foam = await bootstrap(config, new FileDataStore(), [mdProvider]);
+    provider = new TagsProvider(_foam, _foam.workspace);
+  });
+
+  afterAll(async () => {
+    _foam.workspace.dispose();
+    await cleanWorkspace();
+  });
+
+  beforeEach(async () => {
+    await closeEditors();
+  });
+
+  it('correctly provides a tag from a set of notes', async () => {
+    const noteA = createTestNote({
+      tags: new Set(['test']),
+      uri: './note-a.md',
+    });
+    _foam.workspace.set(noteA);
+    provider.refresh();
+
+    const treeItems = (await provider.getChildren()) as Tag[];
+
+    treeItems.map(item => expect(item.tag).toContain('test'));
+  });
+
+  it('correctly handles a parent and child tag', async () => {
+    const noteA = createTestNote({
+      tags: new Set(['parent/child']),
+      uri: './note-a.md',
+    });
+    _foam.workspace.set(noteA);
+    provider.refresh();
+
+    const parentTreeItems = (await provider.getChildren()) as Tag[];
+    const parentTagItem = parentTreeItems.filter(item => item instanceof Tag);
+    expect(parentTagItem[0].title).toEqual('parent');
+    expect(parentTagItem[0].title).not.toEqual('child');
+
+    const childTreeItems = (await provider.getChildren(
+      parentTagItem[0]
+    )) as Tag[];
+
+    childTreeItems.map(child => {
+      if (child instanceof Tag) {
+        expect(child.title).toEqual('child');
+        expect(child.title).not.toEqual('parent');
+      }
+    });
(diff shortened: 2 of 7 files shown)


packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -62,23 +63,51 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
...
+    if (element.title.indexOf(TAG_SEPARATOR) > -1) {
+      return;
+    }
     return element;
   }
 
   getChildren(element?: Tag): Thenable<TagTreeItem[]> {
     if (element) {
-      const references: TagReference[] = element.notes
+      const tagChilds: TagTreeItem[] = this.tags
```
riccardoferretti, line 74: minor: the plural of child is children 😅 
I would suggest naming this something like `nestedTagItems`

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -62,23 +63,51 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
...
+        .map(
+          ({ tag, notes }) =>
+            new Tag(tag, tag.substring(tag.indexOf(TAG_SEPARATOR) + 1), notes)
+        );
+
+      const references: TagTreeItem[] = element.notes
         .map(({ uri }) => this.foam.workspace.get(uri))
-        .map(note => new TagReference(element.tag, note));
+        .map(note => new TagReference(element.tag, note) as TagTreeItem)
+        .concat(tagChilds);
```
riccardoferretti, line 84: if you references separate from nested tags it will simplify your sorting function below as the first separation will already come from the two independent arrays

riccardoferretti (commented):
thanks for looking into this, I have left a few comments, let me know your thoughts!

packages/foam-vscode/src/features/tags-tree-view/index.ts
pderaaij: Not sleeping enough with kids is not helping my English ;-)

packages/foam-vscode/src/features/tags-tree-view/index.ts
pderaaij: I had to do all this stuff as I had issues with the promises and this was the way to get it working properly. I'll update the PR with a nicer solution I found based on your comments. Thanks!

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -102,23 +131,27 @@ type TagMetadata = { title: string; uri: URI };
...
   iconPath = new vscode.ThemeIcon('symbol-number');
   contextValue = 'tag';
 }
 
 export class TagSearch extends vscode.TreeItem {
+  public readonly title: string;
   constructor(public readonly tag: string) {
     super(`Search #${tag}`, vscode.TreeItemCollapsibleState.None);
-    const searchString = `#${tag}`;
+    const searchString = `#?${tag}`;
```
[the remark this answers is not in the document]
pderaaij: Good point, missed that one. Reverted it to not break current behaviour. It does mean that one actually searches on the use of the tag in the text. Which makes sense in a way. If you want to know which bubbles are tagged with that tag you already have the tag explorer..

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -63,52 +63,37 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
...
-            if (
-              (a instanceof Tag && b instanceof Tag) ||
-              (a instanceof TagReference && b instanceof TagReference)
-            ) {
-              a.title.localeCompare(b.title);
-            }
-
-            return 0;
-          }),
+        ...nestedTagItems.concat(references),
```
riccardoferretti, line 90: very minor: to be in line with the general Foam (unformalized) code style, could you change this to

```suggestion
        ...nestedTagItems,
        ...references,
```

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -67,19 +68,32 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
...
-        ...references.sort((a, b) => a.title.localeCompare(b.title)),
+        ...nestedTagItems.concat(references),
       ]);
     }
     if (!element) {
-      const tags: Tag[] = this.tags.map(
-        ({ tag, notes }) => new Tag(tag, notes)
-      );
+      const tags: Tag[] = this.tags
+        .filter(item => item.tag.indexOf(TAG_SEPARATOR) < 0)
```
riccardoferretti, line 95: does this mean that tags that have children won't show up at top level?
if a workspace has the following tags: `#hello/world` and `#lalaland` wouldn't the first one be filtered here?
haven't tried it on my machine, so i might be missing something

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -102,23 +116,27 @@ type TagMetadata = { title: string; uri: URI };
...
+    public readonly title: string,
     public readonly notes: TagMetadata[]
   ) {
-    super(tag, vscode.TreeItemCollapsibleState.Collapsed);
+    super(title, vscode.TreeItemCollapsibleState.Collapsed);
     this.description = `${this.notes.length} reference${
       this.notes.length !== 1 ? 's' : ''
     }`;
     this.tooltip = getContainsTooltip(this.notes.map(n => n.title));
+    this.tag = tag;
```
riccardoferretti, line 127: i don't believe you need this, as the `public readonly` in the argument `tag` is implicitly doing this assignment

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -102,23 +116,27 @@ type TagMetadata = { title: string; uri: URI };
...
   iconPath = new vscode.ThemeIcon('symbol-number');
   contextValue = 'tag';
 }
 
 export class TagSearch extends vscode.TreeItem {
+  public readonly title: string;
   constructor(public readonly tag: string) {
     super(`Search #${tag}`, vscode.TreeItemCollapsibleState.None);
     const searchString = `#${tag}`;
+    this.title = tag.replace(TAG_SEPARATOR, '-');
```
riccardoferretti, line 139: out of curiosity: why replace the separator?

riccardoferretti (commented):
this is looking pretty good, i have left a few more comments!

packages/foam-vscode/src/features/tags-tree-view/index.ts
pderaaij: Interesting remark. I assumed a behaviour that would incline the parent tag always would be explicitly mentioned. Of course this is not ideal, so I changed the code to match the behaviour you described. Fixed in update of the PR.

packages/foam-vscode/src/features/tags-tree-view/index.ts
pderaaij: Turns out it's not necessary anymore due to the other changes. I needed as vscode wouldn't display the item with a forward slash in the title.

packages/foam-vscode/src/features/tags-tree-view/index.ts
riccardoferretti: great, can you add an explicit test to check this behavior (in `foam-vscode`).
Also, can you add please add a test in the `markdown-provider.test.ts` to check that parent/children tags are picked up also from the front matter?

pderaaij: @riccardoferretti I took a stab at tests. Needed to combine a bunch of things, so hopefully this is in line with the style being used.

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        config.includeGlobs,
+        config.ignoreGlobs
+      )
+    );
+    _foam = await bootstrap(config, new FileDataStore(), [mdProvider]);
+    provider = new TagsProvider(_foam, _foam.workspace);
+  });
+
+  afterAll(async () => {
+    _foam.workspace.dispose();
```
riccardoferretti, line 39:
```suggestion
    _foam.dispose();
```

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+  it('correctly handles a parent and child tag', async () => {
+    const noteA = createTestNote({
+      tags: new Set(['parent/child']),
+      uri: './note-a.md',
+    });
+    _foam.workspace.set(noteA);
+    provider.refresh();
+
+    const parentTreeItems = (await provider.getChildren()) as Tag[];
+    const parentTagItem = parentTreeItems.filter(item => item instanceof Tag);
```
riccardoferretti, line 69: what else could these items be?

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    expect(parentTagItem[0].title).toEqual('parent');
+    expect(parentTagItem[0].title).not.toEqual('child');
```
riccardoferretti, line 71: minor: I think you don't need this check as it's implied by the previous one?

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const childTreeItems = (await provider.getChildren(
+      parentTagItem[0]
+    )) as Tag[];
+
+    childTreeItems.map(child => {
+      if (child instanceof Tag) {
+        expect(child.title).toEqual('child');
+        expect(child.title).not.toEqual('parent');
```
riccardoferretti, line 80: same as above

riccardoferretti (commented):
i left a few comments but this is looking really good, great job on the tests

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
pderaaij: On the root level it can only be Tag indeed.

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    const parentTreeItems = (await provider.getChildren()) as Tag[];
+    const parentTagItem = parentTreeItems.filter(
+      item => item instanceof Tag
+    )[0];
+
+    expect(parentTagItem.title).toEqual('parent');
+
+    const childTreeItems = (await provider.getChildren(parentTagItem)) as Tag[];
+
+    childTreeItems.map(child => {
```
riccardoferretti, line 104: to make linter happy. as we are not returning a value, `forEach` is more appropriate than `map`

```suggestion
    childTreeItems.forEach(child => {
```

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -1,8 +1,9 @@
 import * as vscode from 'vscode';
-import { Foam, Resource, URI, FoamWorkspace } from 'foam-core';
+import { Foam, Resource, URI, FoamWorkspace, Logger } from 'foam-core';
```
riccardoferretti, line 2:
```suggestion
import { Foam, Resource, URI, FoamWorkspace } from 'foam-core';
```

packages/foam-core/src/utils/hashtags.ts
```diff
@@ -1,6 +1,6 @@
 import { isSome } from './core';
-const HASHTAG_REGEX = /(^|\s)#([0-9]*[\p{L}_-][\p{L}\p{N}_-]*)/gmu;
-const WORD_REGEX = /(^|\s)([0-9]*[\p{L}_-][\p{L}\p{N}_-]*)/gmu;
+const HASHTAG_REGEX = /(^|\s)#([0-9]*[\p{L}\/_-][\p{L}\p{N}\/_-]*)/gmu;
```
riccardoferretti, line 2: can you please add a linter ignore directive in this and the following line?
we want to keep linter warnings to zero

riccardoferretti (approved):
this all looks good. I only left a couple of comments aimed at making the linter happy, then this is ready to be merged

packages/foam-core/src/utils/hashtags.ts
pderaaij: Turns out the escaping isn't necessary.

pderaaij:
Linter is happy! Sorry for all the back and forth.

riccardoferretti:
thanks @pderaaij!

Issue #640, opened 2021-05-17
Create New Note Skips Asking Filename


smithalexk:
- Foam version: 0.13.3
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**
When I use the Foam: Create New Note command with a .foam/templates/new-note.md template present, I am not given an option for a name. Instead, it creates a note in the current directory called "New Note.md" using the new-note.md template.

This seems to be a bug when reading: https://github.com/movermeyer/foam/blob/mo.template_v2_proposal/docs/proposals/templates-v2.md#case-2-foamtemplatesnew-notemd-exists

**Steps to reproduce**
1. Generate new foam workspace
2. Create a new-note.md in .foam/templates (can be done by renaming the "your-first-template.md" file)
3. Run the Foam: Create New Note command

riccardoferretti:
cc @movermeyer as he is the best person to address the comment

movermeyer:
### Summary

Thanks @smithalexk for raising this issue. It is certainly not a great user experience. I've [opened a PR](https://github.com/foambubble/foam/pull/645) to improve it.

### The technical details

This is happening since the `new-note.md` template does not contain a reference to [`$FOAM_TITLE`](https://github.com/foambubble/foam/blob/master/docs/features/note-templates.md#variables).

This behaviour of not asking for the title unless the template asks for it was intentional, as in the future we want things like `Foam: Open Daily Note` to use this same code to create a new note from template without prompting the user.

This decision lead to this confusing UX.

#### "Working as expected...incorrectly"

It's important to note that our imagined future of `Foam: Open Daily Note` uses the proposed [metadata syntax](https://github.com/foambubble/foam/issues/563) to create a sensible filepath without requiring the user to enter a title:

(Disclaimer: proposed syntax below. Not official/formalized. May still change)

```markdown
---
filepath: '${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}.md'
---

# ${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}
```

In this case, the template provides everything you need in order to create the note (filepath + file contents) without referencing `$FOAM_TITLE`, so there is no need to prompt the user.

In the code, this was implemented as "if the template doesn't mention `$FOAM_TITLE`, then there's nothing to prompt the user for".
[...]

Issue #563, opened 2021-04-02
[Templates v2] Add a metadata syntax to templates


movermeyer:
As described in #534, we want to be able to template the default filepath of a template. Further, this templated filepath metadata should live within the template files themselves.

This requires specifying a syntax for use in template files where this metadata can be written, without being included in the files that are created from the template.

Create this syntax, and use it to allow for the specification of a default filepath for a template.
You should be able to template the filepath using variables just like any other content in the template.

Therefore, the process for creating a template file is:

1. Select the template to use
2. Resolve all the values for all the variables
3. Parse the contents of the metadata section to determine the default filepath
4. Strip the metadata section from the template
5. Copy the resulting template as the contents of the new note

### Example

Using a "second front-matter" mock syntax (Note: this may not be the syntax we want, but is used for demonstration):

```markdown
<!-- The below front-matter block is for foam-specific template settings -->
<!-- It is removed when the user creates a new note using this template -->
---
<!-- The default filepath to use when using this template -->
<!-- Relative paths are relative to the workspace, absolute paths are absolute -->
<!-- Note that you can include VSCode snippet variables to template the path -->
filepath: `journal/${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}_${FOAM_TITLE_SLUG}.md`
---

<!-- The actual contents of the template begin after the `---` thematic break immediately below this line-->
---
---
created: ${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}T${CURRENT_HOUR}:${CURRENT_MINUTE}:${CURRENT_SECOND}
tags: []
---

# ${FOAM_TITLE}
```

jevakallio:
Hey @movermeyer! @riccardoferretti just gave me a walkthrough of the spec, thank you so much for working on this!

The spec all makes sense to me. I was wondering, what is the purpose of the `---` thematic break separator between the "metadata frontmatted" and the "document frontmatter" blocks? If that is required for the markdown parser to read the document, that's fine. If it's not necessary, I'd perhaps opt out of it and simply have two frontmatter blocks run back-to-back?
```md
---
filepath: "..."
---

---
tags: []
---

# Heading
```

Great work on the specification, I like the direction and the way it unifies the handling of different features!

jevakallio:
Further, a suggestion: 

[...]

PR #645, opened 2021-05-20
Always ask for the note title when creating from template

movermeyer:

In order to create a note from a template you need two things:

1. the filepath
2. the file contents.

In cases where the template doesn't explicitly specify the path using the metadata syntax (i.e. all current templates since the metadata syntax isn't implemented yet), we'll add `$FOAM_TITLE` to the list of variables to resolve, and use it as the filepath.

This will prompt the user for a title as they (reasonably) [expect it to](https://github.com/foambubble/foam/issues/640).

Then in the future when the template metadata block contains an explicit override of the filepath, we will stop adding `$FOAM_TITLE`, and whether or not the user gets prompted will depend on the contents of the `filepath` metadata.

Fixes https://github.com/foambubble/foam/issues/640
Ref: https://github.com/foambubble/foam/issues/640#issuecomment-844826452

I've added `FOAM_TITLE` unconditionally to the list of variables to resolve, resulting in `Foam: Create New Note` always asking for a note's title.

In the future, we'll make this conditional on whether the template provides the filepath to use in its [metadata block](#563 ).

Are there cases you could imagine where someone would not want it to prompt for a title, besides the future scenario where the `filepath` metadata is provided in the template?

I'm writing this very early in the morning with little sleep. I may be forgetting some case???

### What doesn't change with this PR?

**Note:** This PR does not change the behaviour of `Foam: Create New Note From Template`.
In cases where the template does not reference `${FOAM_TITLE}`, the user will still not be prompted to enter a title.
This works, since in `Foam: Create New Note From Template` there is a separate step where they fill in the filepath they desire. If we were to always resolve `${FOAM_TITLE}`, it could make this command take longer than necessary, especially in cases where their desired filepath does not match the desired title:

**Example of where always asking for title in `Foam: Create New Note From Template` would be a worse UX**

1. User runs `Foam: Create New Note From Template`; selects an empty template.
1. User is prompted for a title; enters "On the Origin of Species"
1. Filepath prompt now has `<dir>/On the Origin of Species.md`
1. User deletes the existing filepath prompt and changes it to `<dir>/draft-1.md`

The user never wanted the title to be reflected in the filepath, so there was no need to prompt them.
It's fine to always prompt them in the `Foam: Create New Note` case, since that command is opinionated in order to have the most streamlined command for the most common use case. `Foam: Create New Note From Template` is slower, but less opinionated, and more flexible/powerful.

### DIFF
--- a/packages/foam-vscode/src/features/create-from-template.test.ts
+++ b/packages/foam-vscode/src/features/create-from-template.test.ts
@@ -58,15 +58,15 @@ describe('resolveFoamVariables', () => {
   });
 
   test('Resolves FOAM_TITLE', async () => {
-    const foam_title = 'My note title';
+    const foamTitle = 'My note title';
     const variables = ['FOAM_TITLE'];
 
     jest
       .spyOn(window, 'showInputBox')
-      .mockImplementationOnce(jest.fn(() => Promise.resolve(foam_title)));
+      .mockImplementationOnce(jest.fn(() => Promise.resolve(foamTitle)));
 
     const expected = new Map<string, string>();
-    expected.set('FOAM_TITLE', foam_title);
+    expected.set('FOAM_TITLE', foamTitle);
 
     const givenValues = new Map<string, string>();
     expect(await resolveFoamVariables(variables, givenValues)).toEqual(
@@ -75,14 +75,14 @@ describe('resolveFoamVariables', () => {
   });
 
   test('Resolves FOAM_TITLE without asking the user when it is provided', async () => {
-    const foam_title = 'My note title';
+    const foamTitle = 'My note title';
     const variables = ['FOAM_TITLE'];
 
     const expected = new Map<string, string>();
-    expected.set('FOAM_TITLE', foam_title);
+    expected.set('FOAM_TITLE', foamTitle);
 
     const givenValues = new Map<string, string>();
-    givenValues.set('FOAM_TITLE', foam_title);
+    givenValues.set('FOAM_TITLE', foamTitle);
     expect(await resolveFoamVariables(variables, givenValues)).toEqual(
       expected
     );
--- a/packages/foam-vscode/src/features/create-from-template.test.ts
+++ b/packages/foam-vscode/src/features/create-from-template.test.ts
@@ -120,4 +120,30 @@ describe('resolveFoamTemplateVariables', () => {
 
     expect(await resolveFoamTemplateVariables(input)).toEqual(expected);
   });
+
+  test('Allows extra variables to be provided; only resolves the unique set', async () => {
+    const foamTitle = 'My note title';
+
+    jest
+      .spyOn(window, 'showInputBox')
+      .mockImplementationOnce(jest.fn(() => Promise.resolve(foamTitle)));
+
+    const input = `
+      # $FOAM_TITLE
+    `;
+
+    const expectedOutput = `
+      # My note title
+    `;
+
+    const expectedMap = new Map<string, string>();
+    expectedMap.set('FOAM_TITLE', foamTitle);
+
+    const expectedSnippet = new SnippetString(expectedOutput);
+    const expected = [expectedMap, expectedSnippet];
+
+    expect(
+      await resolveFoamTemplateVariables(input, new Set(['FOAM_TITLE']))
+    ).toEqual(expected);
+  });
 });
--- a/packages/foam-vscode/src/features/create-from-template.ts
+++ b/packages/foam-vscode/src/features/create-from-template.ts
@@ -184,12 +184,16 @@ async function askUserForFilepathConfirmation(
 }
 
 export async function resolveFoamTemplateVariables(
-  templateText: string
+  templateText: string,
+  extraVariablesToResolve: Set<string> = new Set()
 ): Promise<[Map<string, string>, SnippetString]> {
   const givenValues = new Map<string, string>();
-  const variables = findFoamVariables(templateText.toString());
+  const variables = findFoamVariables(templateText.toString()).concat(
+    ...extraVariablesToResolve
+  );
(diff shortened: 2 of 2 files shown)


packages/foam-vscode/src/features/create-from-template.test.ts
```diff
@@ -120,4 +120,30 @@ describe('resolveFoamTemplateVariables', () => {
 
     expect(await resolveFoamTemplateVariables(input)).toEqual(expected);
   });
+
+  test('Allows extra variables to be provided; only resolves the unique set', async () => {
+    const foam_title = 'My note title';
```
riccardoferretti, line 125: minor: camel case? 😅

movermeyer:
> What do you think?

@riccardoferretti (I hope I've interpreted it correctly 🤞)

I've tried to map out the behaviour more clearly so that we can be sure that we are talking about the same things.

There are 5 cases, which I have labeled **A-E**. I believe that the only difference between the behaviours is in **Case D**.

### Current behaviour on `master` branch

|                                   | `filepath` doesn't exist              | `filepath` exists, but doesn't contain `$FOAM_TITLE` | `filepath` exists, and contains `$FOAM_TITLE`                       |
| --------------------------------- | ------------------------------------- | ---------------------------------------------------- | ------------------------------------------------------------------- |
| `$FOAM_TITLE` in the template     | **A:** 1 step: Ask for title          | **B:** 1 step: Ask for title                         | **C:** 1 step: Ask for title                                        |
| `$FOAM_TITLE` not in the template | **D:** 0 steps: `New Note.md` created | **E:** 0 steps: note created according to `filepath` | **N/A**, if `$FOAM_TITLE` is in the filepath, it is in the template |

### Current behaviour on this PR

This PR changes the behaviour of **Case D** to ask for the title. This is preferred over the existing `master`-branch behaviour since always having to rename `New Note.md` to the desired name (outside of the note creation flow) is tiresome.

|                                   | `filepath` doesn't exist     | `filepath` exists, but doesn't contain `$FOAM_TITLE` | `filepath` exists, and contains `$FOAM_TITLE`                       |
| --------------------------------- | ---------------------------- | ---------------------------------------------------- | ------------------------------------------------------------------- |
| `$FOAM_TITLE` in the template     | **A:** 1 step: Ask for title | **B:** 1 step: Ask for title                         | **C:** 1 step: Ask for title                                        |
| `$FOAM_TITLE` not in the template | **D:** 1 step: Ask for title | **E:** 0 steps: note created according to `filepath` | **N/A**, if `$FOAM_TITLE` is in the filepath, it is in the template |

### Behaviour described in comment proposal

If I'm parsing it correctly, **Case D** is once again the only thing that changes:

|                                   | `filepath` doesn't exist                                          | `filepath` exists, but doesn't contain `$FOAM_TITLE` | `filepath` exists, and contains `$FOAM_TITLE`                       |
| --------------------------------- | ----------------------------------------------------------------- | ---------------------------------------------------- | ------------------------------------------------------------------- |
| `$FOAM_TITLE` in the template     | **A:** 1 step: Ask for title                                      | **B:** 1 step: Ask for title                         | **C:** 1 step: Ask for title                                        |
| `$FOAM_TITLE` not in the template | **D:** 1 step: Ask for path (Showing `<current_dir>/New Note.md`) | **E:** 0 steps: note created according to `filepath` | **N/A**, if `$FOAM_TITLE` is in the filepath, it is in the template |

> we skip one step by only asking for the filepath

Both this PR and the comment proposal have the same number of steps to create a note. No steps are skipped?

> What do you think?

I think the only difference between these is what to ask the user for when a template does not include `$FOAM_TITLE`?:

* In this PR, it asks the user for the ***title*** of the note
* In the comment proposal, it asks the user for the ***path*** of the note

I prefer asking the user for the title over the path, for two reasons:

* **Consistency:** In this PR, `Foam: Create New Note` only ever asks for a single thing in all cases: the title
  * In the comment proposal, **Case D** will ask for a path, unlike **Cases A-C** which ask for a title
* **Aesthetic:** In general, I'd rather that I never have to look at paths. Titles look cleaner / nicer.
  * If you need control over the path, you can specify a `filepath` (once implemented) or fall back to use `Foam: Create New Note From Template`

riccardoferretti:
hi @movermeyer, in your example above:

> 1. User runs Foam: Create New Note From Template; selects an empty template.
> 2. User is prompted for a title; enters "On the Origin of Species"
> 3. Filepath prompt now has <dir>/On the Origin of Species.md
> 4. User deletes the existing filepath prompt and changes it to <dir>/draft-1.md

There are two inputs required from the user, wasn't this the flow you were thinking of?

movermeyer:
> There are two inputs required from the user, wasn't this the flow you were thinking of?

That example is for `Foam: Create New Note From Template`, to demonstrate why I'm not also changing the behaviour of that flow in this PR. The code paths look very similar at first glance, so a reviewer might have asked "Why aren't we also making this change for `Foam: Create New Note From Template`?". I've updated the PR description to hopefully be a little more clear.

I think your comment describes the current behaviour of `Foam: Create New Note From Template`: We only ask for the title if `$FOAM_TITLE` is needed.

Current behaviour of `Foam: Create New Note From Template`:

1. User selects the template
2. If `$FOAM_TITLE` is needed, ask for the title
3. Confirm the filepath (using `$FOAM_TITLE` if it was resolved, else `New Note` as the default)

So currently, there are at most 3 steps to `Foam: Create New Note From Template`, but one of them is conditional on `$FOAM_TITLE` use. Making the same change to `Foam: Create New Note From Template` that this PR makes to `Foam: Create New Note`would cause it to always be 3 steps, which is not what we want.

riccardoferretti:
oops, my wires crossed on that one, you are right

riccardoferretti (approved):
LGTM!

PR #647, opened 2021-05-20
Allow for title property missing in frontmatter

pderaaij:
During development of another feature I saw the foam log was overflowing with undefined errors. As I don't have any title properties in Frontmatter `props.title` is undefined. As a result, the toString() leads to problems. This change allows for optional presence for the title property.

### DIFF
--- a/packages/foam-core/src/markdown-provider.ts
+++ b/packages/foam-core/src/markdown-provider.ts
@@ -177,7 +177,7 @@ const titlePlugin: ParserPlugin = {
   },
   onDidFindProperties: (props, note) => {
     // Give precendence to the title from the frontmatter if it exists
-    note.title = props.title.toString() ?? note.title;
+    note.title = props.title?.toString() ?? note.title;
   },
   onDidVisitTree: (tree, note) => {
     if (note.title === '') {


riccardoferretti:
thanks for spotting this, I am actually really surprised because I thought there was a test that was checking just that.
why do you think it's not working? can you add a test case that is fixed by this PR?

pderaaij:
So, this is a bit weird. Looking at the tests I would expect the problem to arise in this test as well:

```typescript
it('should parse empty frontmatter', () => {
    const note = createNoteFromMarkdown(
      '/page-f.md',
      `
---
---

# Empty Frontmatter
`
    );

    expect(note.properties).toEqual({});
  });
```
However, things are fine there. If I open the foam log I do so the errors. Could it be that Jest is hiding the errors in a way?

Screenshot 2021-05-21 at 09 37 23

riccardoferretti:
I think the error is being caught by the parser logic.
your fix is correct though, let's move forward with it.

riccardoferretti (approved):

riccardoferretti:
@allcontributors add @pderaaij for code

Issue #565, opened 2021-04-02
[Templates v2] Make `Foam: Open Daily Note` use a template


movermeyer:
As described in #534, change `Foam: Open Daily Note` to no longer use the existing settings, but instead use a template found at `.foam/templates/daily-note.md` (or an implicit template if `.foam/templates/daily-note.md` does not exist).

Existing settings:

- `foam.openDailyNote.directory`
- `foam.openDailyNote.filenameFormat`
- `foam.openDailyNote.fileExtension`

jonasdiemer:
Hi @movermeyer ,

struggling with my daily notes using a "long" date format (# Monday, June 21, 2021) rather than the preferred isolate (# 2021-06-21). Looks like "foam.openDailyNote.titleFormat" doesn't work anymore - is this a know issue due to your mentioned refactoring? Would be grateful for a tip on how to fix it...

Thanks!

movermeyer:
@jonasdiemer I haven't done any work in this area yet, so its not related to this ticket. 

I just customized `foam.openDailyNote.titleFormat` (`"dddd, mmmm dS, yyyy"`) and it seems to have worked fine.

Could you please [open a new issue](https://github.com/foambubble/foam/issues/new/choose) and include the steps needed to cause the issue? Thanks!

jonasdiemer:
Thanks. Done: https://github.com/foambubble/foam/issues/687

janaka:
Hi, just wondering about status of this. Is it in-progress? Is there an Eta?

I'm really feeling it will give me a productivity boost.

PR #655, opened 2021-05-24
Allow for YAML metadata in templates

movermeyer:

As part of #563, we are adding a mechanism to provide metadata within a Frontmatter block within a Foam note template.

As noted in [this comment](https://github.com/foambubble/foam/issues/563#issuecomment-841688114), it's nicer to be able to add the metadata to an existing YAML frontmatter block.
However, it is also very difficult to programmatically remove an entry from a YAML mapping without disrupting the rest of the YAML.

So in this PR, I allow for both. If you have an existing YAML frontmatter block, and you're willing to abide by fairly strict formatting, then you can add the metadata to the existing frontmatter block.

If for some reason, you cannot use the restricted syntax, you are still able to provide a second, Foam-specific YAML frontmatter block at the start of the template instead.

#### Not quite there

This PR does not succeed at fully implementing the desired behaviour.
I had dreamed of being able to use any VS Code snippet variable in the `filepath`.

```yaml
---
foam_template:
  filepath: `journal/${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}.md`
---
```

This would allow `Foam: Open Daily Note` to make use of templates easily.
However, this is not currently possible. VS Code does not currently offer an interface for resolving the snippet variables without opening the snippet in an editor for the user.

In the future, we might vendor the relevant bits snippet resolution code, or submit a patch to VS Code to expose their variable resolution functionality as part of the API.

The PR as it stands still offers value, since it allows you to define relative or absolute directories to store the files into, instead of the default `Foam: Create New Note` behaviour of opening the file in the same directory as the active file.

**Example of using metadata to specify a directory relative to the current workspace**

```yaml
---
# This will create the note in the "journal" subdirectory of the current workspace, 
# regardless of which file is the active file. 
foam_template:
  filepath: `journal/$FOAM_TITLE.md` 
---
```

### Next steps

* Figure out a way to allow for VS Code snippet variables within the metadata
* [Make `Foam: Open Daily Note` use templates](https://github.com/foambubble/foam/issues/565)


riccardoferretti:
The general approach looks good to me, a few comments:
- I didn't realize there was such limitation on Snippet strings, that's unfortunate, we should look into getting an API to the text expansion
- I am happy to not support JSON frontmatter
- I am happy to only support a single YAML block if that simplifies your life, I don't mind supporting 2 YAML blocks if you prefer that way and it doesn't make the code awfully more complicated

movermeyer:
@riccardoferretti, I think this is ready for review.

Since you last looked, I've implemented the `name` and `description` template metadata attributes that [`@jevakallio` suggested](https://github.com/foambubble/foam/issues/563#issuecomment-817979917). 

IMO, it looks pretty slick:

foam-template-picker

---

Replying to your previous comment.

> I didn't realize there was such limitation on Snippet strings, that's unfortunate, we should look into getting an API to the text expansion

:100:, that's the next step after this PR.

> I am happy to not support JSON frontmatter

👍

> I am happy to only support a single YAML block if that simplifies your life, I don't mind supporting 2 YAML blocks if you prefer that way and it doesn't make the code awfully more complicated

It wasn't very complicated to support both (with the caveat that users cannot use the full flexibility of the YAML syntax if they want to have the metadata within an existing Frontmatter bock)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -51,9 +60,21 @@ For a full list of features see [the VS Code snippets page](https://code.visuals
 2. create a note from this template by running the \`Foam: Create New Note From Template\` command
 `;
 
-async function getTemplates(): Promise<string[]> {
+async function templateMetadata(templateUri: Uri): Promise<object> {
+  const contents = await workspace.fs
+    .readFile(templateUri)
+    .then(bytes => bytes.toString());
+  let templateMetadata: object, _templateWithFoamFrontmatterRemoved: string;
```
riccardoferretti, line 67: minor: why are you declaring these independently? couldn't you just do the following?

```
const [ templateMetadata ] = extract...
```

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -153,12 +174,73 @@ export function substituteFoamVariables(
...
+  return templatesMetadata.sort((t1, t2) => {
+    if (t1['name'] === undefined && t2['name'] !== undefined) {
+      return 1;
+    }
+
+    if (t1['name'] !== undefined && t2['name'] === undefined) {
+      return -1;
+    }
+
+    if (t1['name'] > t2['name']) {
```
riccardoferretti, line 188: minor: I wonder if this whole section could be replaced by

```
return t1['name'].localeCompare(t2['name']) 
  || t1['templatePath'].localeCompare(t2['templatePath'])
```

?

(haven't tested it, just a hunch)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -153,12 +174,73 @@ export function substituteFoamVariables(
...
+}
+
 async function askUserForTemplate() {
   const templates = await getTemplates();
   if (templates.length === 0) {
     return offerToCreateTemplate();
   }
-  return await window.showQuickPick(templates, {
+
+  const templatesMetadata = sortTemplatesMetadata(
```
riccardoferretti, line 214: somewhat minor, mostly style: in general in the code we try to provide sorting functions instead of functions that sort (I hope it's clear what I mean).

basically here it would translate into something like:

```
(await Promise.all( .... )).sort(sortMetadataByNameAndPath)
```

thoughts?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -218,15 +299,41 @@ function currentDirectoryFilepath(filename: string) {
   return Uri.joinPath(currentDir, filename);
 }
 
+export function determineDefaultFilepath(
+  resolvedValues: Map<string, string>,
+  templateMetadata
```
riccardoferretti, line 304: is this an implicit `any`? in that case I am surprised it didn't raise a warning from the linter.

I also noticed in other places you are using `object` for metadata.. given that we have a specific shape for the template metadata, could it be helpful to map it onto a type/interface?

riccardoferretti (commented):
the screenshots look pretty cool, thanks for this!
I have added a few comments, mostly minor, let me know your thoughts

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: Makes total sense. Fixed.

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: > is this an implicit any? in that case I am surprised it didn't raise a warning from the linter.

It is. I guess I just missed the warning: 

image

---

I've switched it to `Map<string, string>`... if that helps?

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: 🎉 That allows it to shrink a bit. We still have to handle the cases where `name` is `undefined`.

I've made the changes. I'm happy to be shown even more compact solutions.

packages/foam-vscode/src/features/create-from-template.ts
movermeyer: I seem to recall the compiler complaining at me about that. 🤔 
But more likely, my limited TypeScript knowledge could have messed me up and I could have been doing something else wrong that caused the compiler to yell at me.

---

At any rate, it works as you suggested so I've made the change. Thanks. 👍

packages/foam-vscode/src/features/create-from-template.ts
riccardoferretti: yeah I was thinking of some defaults, but I couldn't find something that wouldn't make things more obscure, so I am happy to keep it the way it is

packages/foam-vscode/src/features/create-from-template.ts
riccardoferretti: I am surprised it didn't show up as a warning in the PR, I thought the integration would do that (kinda like when there is an unused import/variable)... mm, to be looked into at some point.

Given the metadata has a specific type, I would just defined that, as it also provides a good in-code documentation for what info can be found (and is allowed) as template metadata, wdyt?

movermeyer: > Given the metadata has a specific type, I would just defined that, as it also provides a good in-code documentation for what info can be found (and is allowed) as template metadata, wdyt?

@riccardoferretti Sounds good to me. I won't have time to make this change until next week. If you're eager, we can ship this now, and I'll do that refactor as part of my next PR.

riccardoferretti (approved):
happy to merge this in the meantime, will do a release later so the feature will already be available. thanks!

chrisUsick:
Just came across an issue that I couldn't find in the docs, started reading the code and found `foam_template.filepath` on master but couldn't figure out why it appeared template metadata wasn't working, then noticed that this if 5h old!

PR #656, opened 2021-05-25
Revive support for basic link alias

pderaaij:
## What are you trying to accomplish?
Implementing support of piped links to handle basic aliases of links. 

Screenshot 2021-05-25 at 20 51 07

## What approach did you choose and why?
I've revived the implementation as in #223. It resembles the same functionality, but in the refactored code and with additional tests.

## Next steps

- There is an issues with piped links displayed as a table in Github Pages. This PR will not solve that problem. We should look into that specific issue.

### DIFF
--- a/packages/foam-core/src/markdown-provider.ts
+++ b/packages/foam-core/src/markdown-provider.ts
@@ -14,6 +14,7 @@ import {
   ResourceLink,
   WikiLink,
   ResourceParser,
+  NoteSource,
 } from './model/note';
 import { Position } from './model/position';
 import { Range } from './model/range';
@@ -31,9 +32,11 @@ import { ResourceProvider } from 'model/provider';
 import { IDataStore, FileDataStore, IMatcher } from './services/datastore';
 import { IDisposable } from 'common/lifecycle';
 
+const ALIAS_DIVIDER_CHAR = '|';
+
 export interface ParserPlugin {
   name?: string;
-  visit?: (node: Node, note: Resource) => void;
+  visit?: (node: Node, note: Resource, noteSource: NoteSource) => void;
   onDidInitializeParser?: (parser: unified.Processor) => void;
   onWillParseMarkdown?: (markdown: string) => string;
   onWillVisitTree?: (tree: Node, note: Resource) => void;
@@ -121,7 +124,7 @@ export class MarkdownResourceProvider implements ResourceProvider {
     switch (link.type) {
       case 'wikilink':
         const definitionUri = resource.definitions.find(
-          def => def.label === link.slug
+          def => def.label === link.target
         )?.url;
         if (isSome(definitionUri)) {
           const definedUri = URI.resolve(definitionUri, resource.uri);
@@ -130,8 +133,8 @@ export class MarkdownResourceProvider implements ResourceProvider {
             URI.placeholder(definedUri.path);
         } else {
           targetUri =
-            workspace.find(link.slug, resource.uri)?.uri ??
-            URI.placeholder(link.slug);
+            workspace.find(link.target, resource.uri)?.uri ??
+            URI.placeholder(link.target);
         }
         break;
 
@@ -197,12 +200,28 @@ const titlePlugin: ParserPlugin = {
 
 const wikilinkPlugin: ParserPlugin = {
   name: 'wikilink',
-  visit: (node, note) => {
+  visit: (node, note, noteSource) => {
     if (node.type === 'wikiLink') {
+      const text = node.value as string;
+      const alias = node.data?.alias as string;
+      const literalContent = noteSource.text.substring(
+        node.position!.start.offset!,
+        node.position!.end.offset!
+      );
+
+      const hasAlias =
+        literalContent !== text && literalContent.includes(ALIAS_DIVIDER_CHAR);
       note.links.push({
         type: 'wikilink',
-        slug: node.value as string,
-        target: node.value as string,
+        rawText: literalContent,
+        label: hasAlias
+          ? alias.trim()
+          : literalContent.substring(2, literalContent.length - 2),
+        target: hasAlias
+          ? literalContent
+              .substring(2, literalContent.indexOf(ALIAS_DIVIDER_CHAR))
+              .trim()
+          : text.trim(),
         range: astPositionToFoamRange(node.position!),
       });
     }
@@ -259,7 +278,7 @@ export function createMarkdownParser(
   const parser = unified()
     .use(markdownParse, { gfm: true })
     .use(frontmatterPlugin, ['yaml'])
-    .use(wikiLinkPlugin);
+    .use(wikiLinkPlugin, { aliasDivider: ALIAS_DIVIDER_CHAR });
 
   const plugins = [
     titlePlugin,
@@ -307,6 +326,13 @@ export function createMarkdownParser(
         },
       };
 
+      var noteSource: NoteSource = {
+        text: markdown,
+        contentStart: astPointToFoamPosition(tree.position!.start),
+        end: astPointToFoamPosition(tree.position!.end),
+        eol: eol,
+      };
+
       plugins.forEach(plugin => {
         try {
           plugin.onWillVisitTree?.(tree, note);
@@ -342,7 +368,7 @@ export function createMarkdownParser(
 
         for (let i = 0, len = plugins.length; i < len; i++) {
           try {
-            plugins[i].visit?.(node, note);
+            plugins[i].visit?.(node, note, noteSource);
           } catch (e) {
             handleError(plugins[i], 'visit', uri, e);
           }
@@ -431,7 +457,14 @@ export function createMarkdownReferences(
         : dropExtension(relativePath);
 
       // [wiki-link-text]: path/to/file.md "Page title"
-      return { label: link.slug, url: pathToNote, title: target.title };
+      return {
+        label:
+          link.rawText.indexOf('[[') > -1
+            ? link.rawText.substring(2, link.rawText.length - 2)
+            : link.rawText || link.label,
+        url: pathToNote,
+        title: target.title,
+      };
     })
     .filter(isSome)
     .sort();
--- a/packages/foam-core/src/markdown-provider.ts
+++ b/packages/foam-core/src/markdown-provider.ts
@@ -14,7 +14,6 @@ import {
   ResourceLink,
   WikiLink,
   ResourceParser,
-  NoteSource,
 } from './model/note';
 import { Position } from './model/position';
 import { Range } from './model/range';
@@ -36,7 +35,7 @@ const ALIAS_DIVIDER_CHAR = '|';
 
 export interface ParserPlugin {
   name?: string;
-  visit?: (node: Node, note: Resource, noteSource: NoteSource) => void;
+  visit?: (node: Node, note: Resource, noteSource: string) => void;
   onDidInitializeParser?: (parser: unified.Processor) => void;
   onWillParseMarkdown?: (markdown: string) => string;
   onWillVisitTree?: (tree: Node, note: Resource) => void;
@@ -204,7 +203,7 @@ const wikilinkPlugin: ParserPlugin = {
     if (node.type === 'wikiLink') {
       const text = node.value as string;
       const alias = node.data?.alias as string;
-      const literalContent = noteSource.text.substring(
+      const literalContent = noteSource.substring(
         node.position!.start.offset!,
         node.position!.end.offset!
       );
@@ -326,13 +325,6 @@ export function createMarkdownParser(
         },
       };
 
-      var noteSource: NoteSource = {
-        text: markdown,
-        contentStart: astPointToFoamPosition(tree.position!.start),
-        end: astPointToFoamPosition(tree.position!.end),
-        eol: eol,
-      };
-
       plugins.forEach(plugin => {
         try {
           plugin.onWillVisitTree?.(tree, note);
@@ -368,7 +360,7 @@ export function createMarkdownParser(
 
         for (let i = 0, len = plugins.length; i < len; i++) {
           try {
-            plugins[i].visit?.(node, note, noteSource);
+            plugins[i].visit?.(node, note, markdown);
           } catch (e) {
             handleError(plugins[i], 'visit', uri, e);
           }
--- a/packages/foam-vscode/src/features/preview-navigation.ts
+++ b/packages/foam-vscode/src/features/preview-navigation.ts
@@ -1,7 +1,10 @@
-import * as vscode from 'vscode';
-import markdownItRegex from 'markdown-it-regex';
 import { Foam, FoamWorkspace, Logger, URI } from 'foam-core';
+import markdownItRegex from 'markdown-it-regex';
+import * as vscode from 'vscode';
 import { FoamFeature } from '../types';
+import { isNone } from '../utils';
+
+const ALIAS_DIVIDER_CHAR = '|';
 
 const feature: FoamFeature = {
   activate: async (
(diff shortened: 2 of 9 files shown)


pderaaij:
@riccardoferretti The change seems to work as it should. I am a bit sceptical about the tests. I needed to make sure the newly added test in `documents-link-provider.spec.ts` was the last one. Otherwise, I could run into flaky tests (see: https://github.com/foambubble/foam/runs/2668379484?check_suite_focus=true#step:7:174). Feels not good, but I can't imagine why the workspace is randomly disposed. Perhaps you have some ideas?

riccardoferretti (commented):
This looks really good, I would like to suggest another change, which is a big deeper but will make it much easier for us to (in a different PR) merge the model of `WikiLink` and `DirectLink`.

Basically in the future I would like to get to a point where I have this API:

```
interface Link {
  type: 'wikilink' | 'link'
  range: Range  // the range of the link (including brackets, to be checked whether now that's the behavior)
  target: string  // the target of the link (slug if wikilink, path if link)
  label: string // the label of the link (alias if wikilink with divider, slug if wikilink without divider, title if link)
  text: string // the original text from the document (what you call literalContent), including brackets (so it can support both links and wikilinks)
}
```

note that the fields will always exist (no `| null`), and have specific semantics.

if this feels a bit out of scope (which I can see why), then I'd say:
- rename `literalContent` to `text`, and have it include brackets
- rename `alias` to `label`, make it mandatory, and default to `slug` when no divider is present

I am also open to other names for `text`. Maybe `sourceText`, `rawText`, or let me know if you have other ideas now that I have shared with you some of the directional thinking for this model. (`literalContent` is not bad either, I wonder though if it can be "simpler")

pderaaij:
I took a first attempt at the refactor. If I understood it well, this should roughly resemble what you see for the future?

All functionalities still work, but an interesting thing became present. When I am not making use of the auto-generated markdown references the alias is properly shown in the markdown preview.

Screenshot 2021-05-27 at 20 40 04

However, as soon as the references are generated the alias won't be shown in the preview any more. 
Screenshot 2021-05-27 at 20 40 31

I need to look into the reason why this occurs. Perhaps good not to merge this PR before I've finished the investigation.

packages/foam-core/src/markdown-provider.ts
```diff
@@ -190,10 +191,22 @@ const wikilinkPlugin: ParserPlugin = {
...
+        literalContent !== text && literalContent.includes(ALIAS_DIVIDER_CHAR);
       note.links.push({
         type: 'wikilink',
-        slug: node.value as string,
-        target: node.value as string,
+        rawText: literalContent,
+        label: hasAlias
+          ? alias.trim()
+          : literalContent.substring(2, literalContent.length - 2),
+        target: (node.value as string).trim(),
```
riccardoferretti, line 209: should we use `data.permalink` here instead?

packages/foam-core/src/markdown-provider.ts
```diff
@@ -422,7 +436,14 @@ export function createMarkdownReferences(
         : dropExtension(relativePath);
 
       // [wiki-link-text]: path/to/file.md "Page title"
-      return { label: link.slug, url: pathToNote, title: target.title };
+      return {
+        label:
+          link.rawText.indexOf('[[') > -1
```
riccardoferretti, line 441: labels should only be created for wikilinks, this test should always be true, right?

packages/foam-core/src/markdown-provider.ts
```diff
@@ -422,7 +436,14 @@ export function createMarkdownReferences(
...
+            ? link.rawText.substring(2, link.rawText.length - 2)
```
riccardoferretti, line 442: you raised a good point in your PR comment, I don't know what text goes here exactly, e.g. whether just the target or the raw text, we should look closely at the specs

riccardoferretti (commented):
the direction is definitely right.
I have left a few comments, let me know your thoughts.
thanks!

pderaaij:
I've been looking into the link aliasing with the link references. CommonMark does not allow for link aliases. The behaviour of CommonMark and GitHub flavour as such is to display the link with the pipe and the alias, e.g. `[link|alias]`. See this Gist as an example: https://gist.github.com/pderaaij/bd67bb6c0c0074389e70ef1249a8ba1e#file-test-alias-md.

As this is the behaviour of CommonMark the preview window of CommonMark displays the link fully as well. The target of the link is correct however, it's just not pretty.

The proper way to work with aliases in CommonMark is to use: `[integration][202105072028_Integration-in-organisation]`, see: https://gist.github.com/pderaaij/bd67bb6c0c0074389e70ef1249a8ba1e#file-test-only-alias-md

So, I see a couple of options moving forward:

1. implement link aliases for all situations, but knowing it would not work properly when using link references.
2. Only make link aliases work when the setting `foam.edit.linkReferenceDefinitions` is set to off. 
3. (in case of above) Create something for publishing to rewrite aliased links to CommonMark allowed format. But, don't know how to handle the preview screen in that case.

Or something else, but open for ideas here :)

packages/foam-core/src/markdown-provider.ts
pderaaij: The format we use right now is the correct one if I follow the CommonMark specs, so suggest we stick to this.

riccardoferretti:
thanks for the investigation.
I would say we go for n. 3.
It will be possible to handle the preview by manually adjusting the links, see `preview-navigation.ts` - wanna take a stab at it?

pderaaij:
I'll have a look at it!

packages/foam-core/src/markdown-provider.ts
```diff
@@ -190,10 +191,22 @@ const wikilinkPlugin: ParserPlugin = {
   name: 'wikilink',
   visit: (node, note) => {
     if (node.type === 'wikiLink') {
+      const text = node.value as string;
+      const alias = node.data?.alias as string;
+      const literalContent = note.source.text.substring(
```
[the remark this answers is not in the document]
pderaaij: I am doing some tests with using `data.permalink` but that seems to break compatibility. The permalink fields will contained lowercased and hyphens (in place of spaces). If we moved forward to use this field I am assuming it breaks current setups that have spaces in their filenames. But, perhaps I am missing something here?

pderaaij:
I've been looking into it, but getting a bit stuck. So the code of `preview-navigation` is simple, but it seems not to be used. Even with all extensions disabled it doesn't use the code for the markdown preview. Probably I might be missing something here?

With regards to rewrite the alias I am bit searching what a right approach would be. As I see it right now foam has nothing present to manipulate files in publishing command or something like that? So, would this be something we'd like to add to the foam cli?

riccardoferretti:
> So the code of preview-navigation is simple, but it seems not to be used.

Not sure I understand, `preview-navigation.ts` is exposed in `features/index.ts` and then initialized with all the other features in `extension.ts` - is this what you are referring to?

> With regards to rewrite the alias I am bit searching what a right approach would be

if you look at `markdownItWithFoamLinks` in `preview-navigation.ts` you will see that we search for `[[wikilinks]]` and wrap them into a `<a href=...>wikilink</a>` for the preview html.
In this same context we could support the alias: `[[wikilink|alias]]` -> `<a href=...>alias</a>`

Is that helping?

packages/foam-core/src/markdown-provider.ts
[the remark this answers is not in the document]
riccardoferretti: oh good find, I wasn't aware of it. in this case I agree that we can't use the `permalink` property, but instead need to use the text from the wikilink (and strip the alias if present).

pderaaij:
Yeah, I've been manipulating that piece of code. But even when I put in some dummy text it is not showing up in the preview. The debugger also doesn't reach that parrt of the code.. so, I am quite sure I am at the right part of the code. However, the code is not being used in my current setup and I don't know why.

riccardoferretti:
happy to look into this together, I think the work is really important and 98% there, we should push this through.

A couple of things that might help:
- I reviewed your code, you are based off an older version of master, that's why you don't see the decorating code being executed. Rebase from master to fix that
- I'd suggest you disable the wikilink definition generation while working on this, as it can add a level of complexity

Let me know if you want to pair up, or have questions, thanks again!

pderaaij:
The issue has to do with the link references. If they are present `markdown-it` will use these definitions to alter the state of the links. Rightfully so looking at the CommonMark spec. As such, I am manipulating the `markdown-it`plugin to disable the `reference` rule. That makes the link work, but leaves it up to me to remove the link definition block in the preview. Right now, I can only manipulate the AST tree, so need to find a better way for that. Hope to have some time to finish the last bit tomorrow.

Without the link definitions the preview is working properly.

pderaaij:
So, I woke up with a solution for this. It basically means we disable the default link rule of `markdown-it` and replace it with one that does not make use of references. Works for our use case, bust most likely interferes with other extensions. In any case, this does make the preview work accordingly. We could make it a configurable thing, so the choice is with the user?

pderaaij:
Okay, this is a bit weird. All the tests work on my [fork](https://github.com/pderaaij/foam/commit/e920999011f91c498184f65f58cf4ca663010c29), but not here as a PR. Any clue?

riccardoferretti:
it looks like that `link.target` in `markdown-provider.ts:132` is `undefined` or `null`, how could that happen?

pderaaij:
Okay, this is a bit weird. All the tests work on my [fork](https://github.com/pderaaij/foam/commit/e920999011f91c498184f65f58cf4ca663010c29), but not here as a PR. Any clue?

> it looks like that `link.target` in `markdown-provider.ts:132` is `undefined` or `null`, how could that happen?

Hmmz, weird that it only happens in a Windows environment in that case, and not in my personal fork? That feels off. I don't have the time to look into it tonight. Try to have a look tomorrow

packages/foam-core/src/plugins/index.ts
```diff
@@ -0,0 +1,86 @@
+import * as fs from 'fs';
```
riccardoferretti, line 1: this file had been delete on master

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -11,11 +12,14 @@ const feature: FoamFeature = {
     const foam = await foamPromise;
 
     return {
-      extendMarkdownIt: (md: markdownit) =>
-        [markdownItWithFoamTags, markdownItWithFoamLinks].reduce(
-          (acc, extension) => extension(acc, foam.workspace),
-          md
-        ),
+      extendMarkdownIt: (md: markdownit) => {
+        // md.disable('reference');
```
riccardoferretti, line 16: remove

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -29,13 +33,24 @@ export const markdownItWithFoamLinks = (
...
     replace: (wikilink: string) => {
       try {
-        const resource = workspace.find(wikilink);
-        if (resource == null) {
-          return getPlaceholderLink(wikilink);
+        let linkHasAlias = wikilink.includes('|');
+        let resourceLink = wikilink;
+        if (linkHasAlias) {
+          resourceLink = wikilink.substring(0, wikilink.indexOf('|'));
+        }
```
riccardoferretti, line 40: minor: 

```suggestion
        const resourceLink = linkHasAlias  
          ? wikilink.substring(0, wikilink.indexOf('|'))
          : wikilink;
```

for two reasons:
1. let's use `const` whenever possible
2. this is easier to read

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -29,13 +33,24 @@ export const markdownItWithFoamLinks = (
...
+        const resource = workspace.find(resourceLink);
+        if (resource === null) {
+          return getPlaceholderLink(resourceLink);
         }
+
+        let linkLabel = wikilink;
+        if (linkHasAlias) {
+          linkLabel = linkLabel.substr(wikilink.indexOf('|') + 1);
+        }
```
riccardoferretti, line 49:
```suggestion
        const linkLabel = linkHasAlias
          ? linkLabel.substr(wikilink.indexOf('|') + 1)
          : wikilink;
```

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -60,7 +75,7 @@ export const markdownItWithFoamTags = (
     replace: (tag: string) => {
       try {
         const resource = workspace.find(tag);
-        if (resource == null) {
+        if (resource === null) {
```
riccardoferretti, line 78: the reason why we use `==` instead of `===` is because it also catches `undefined`, as opposed to just strictly `null`.

But the right approach here would be to use the utility function `isNone`

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -77,4 +92,120 @@ export const markdownItWithFoamTags = (
...
   `<span class='foam-tag'>${content}</span>`;
 
+export const markdownItWithRemoveLinkReferences = (
+  md: markdownit,
+  workspace: FoamWorkspace
+) => {
+  return md.disable('link').use(linkRuleProcessingWithoutReferences());
+};
+
+function linkRuleProcessingWithoutReferences() {
```
riccardoferretti, line 102: I fear this approach is a bit heavy and disconnects us from future improvements in markdown-it.

Have you considered instead filtering the files to remove the section generated by Foam before it's sent to markdown-it? (not sure whether with the available API this is possible, have you already looked at that option?)

riccardoferretti (commented):
this is looking very good. I have left a couple of minor comments and one about the link definitions, let me know what you think

packages/foam-vscode/src/features/preview-navigation.ts
pderaaij: I've tried to do that, but it seems manipulation can only be done on tokens. The link reference definition is not passed on as a token and as such manipulation is hard, The only thing I could find is to change the rule set of the inline link rules like I did above.

What still can be an option is that we don't display the alias when link references are present in the file. It then just shows `link|alias` in the preview. Which is actually the same as in GitHub without any manipulation of the Markdown files.

riccardoferretti: mmm... I would need to look a bit more deeply into the API to properly help here.

I came across this: https://www.npmjs.com/package/markdown-it-replace-link - could it help?

pderaaij: I would need to look at. The order in which all changes are executed seem to be of importance here. As soon as you save a document, the presentation of the link reference gets priority above the foam link regex. 

Fresh eyes probably help here ;-)

pderaaij: For what I can see [vscode just reads the document ](https://github.com/microsoft/vscode/blob/77ff6eb03bcfb11d3bad26346f0b5ed593cb2a99/extensions/markdown-language-features/src/markdownEngine.ts#L132) and puts it in the parser of MarkdownIt. Then the tokenisation and rules start to apply. In both projects I don't see an possibility to intervene in this behaviour. For example, to filter out parts of the document before the tokenisation starts.

pderaaij:
Well, it has been quite a journey. I do believe an acceptable solution has been found. In this case, we add a custom rule to the link parser and iterate over the tokens to get the right behaviour. It was a bit of a journey, but it feels like the right approach. 

I fixed the latest remarks and squashed the commits. Please have a look at it @riccardoferretti and see what you think of it.

lautjy:
I understand that my comment is a bit late in the game, but do you have any thoughts on the order of the alias/wikilink?
Like in Dendron the order is [[alias | wikilink_338]], but here you are choosing reversed. Is there a reason.
Dendron docs on the matter: https://wiki.dendron.so/notes/3472226a-ff3c-432d-bf5d-10926f39f6c2.html

pderaaij:
@lautjy the WikiLink format is alias last. I actually came across an issue at Dendron to support the WikiLink format with alias last. So decided to take this order as it seemed the most common format.

riccardoferretti (approved):
LGTM!

StoltHD:
I have a problem with wikilinks with aliases in tables, those need to be escaped with a single backslash to work, but in Foam links with aliases that are escaped with backslash point to "notename\".
And if you add a link with alias without the "escape" the link do not work.

PR #661, opened 2021-05-27
Filter out tag references that are part of a nested tag

pderaaij:
While transforming my repo to use nested tags I saw an issue in the tag tree view explorer. Due to the way we nest tags it wrongfully displayed references that are part of a nested tag. 

Screenshot 2021-05-27 at 14 18 23

The two references displayed here are part of the `change` nested tag. This minor fix ensures proper displaying of references and nested tags, see:

Screenshot 2021-05-27 at 14 46 36

### DIFF
--- a/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
+++ b/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
@@ -4,7 +4,7 @@ import {
   createTestNote,
 } from '../../test/test-utils';
 
-import { Tag, TagReference, TagsProvider } from '.';
+import { Tag, TagsProvider } from '.';
 
 import {
   bootstrap,
--- a/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
+++ b/packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
@@ -4,7 +4,7 @@ import {
   createTestNote,
 } from '../../test/test-utils';
 
-import { Tag, TagsProvider } from '.';
+import { Tag, TagReference, TagsProvider } from '.';
 
 import {
   bootstrap,
@@ -68,7 +68,6 @@ describe('Tags tree panel', () => {
     const parentTreeItems = (await provider.getChildren()) as Tag[];
     const parentTagItem = parentTreeItems.pop();
     expect(parentTagItem.title).toEqual('parent');
-    expect(parentTagItem.title).not.toEqual('child');
 
     const childTreeItems = (await provider.getChildren(parentTagItem)) as Tag[];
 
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -82,6 +82,16 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
 
       const references: TagTreeItem[] = element.notes
         .map(({ uri }) => this.foam.workspace.get(uri))
+        .filter(note =>
+          // foreach nested tag item check if the exact tag is present as a tag on the note. If so,
+          // it belongs to a nested tag item and should not be displayed as a reference underneath the current element
+          nestedTagItems.reduce(
```
riccardoferretti, line 88: IIUC what you are trying to achieve, could you use `nestedTagItems.find`?
Then depending on what you need, you can use `isSome` or `isNone`

pderaaij: Ah, that might be a nice solution. Was working with my limited knowledge and didn't know about that. Will have a look.

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -82,6 +87,11 @@ export class TagsProvider implements vscode.TreeDataProvider<TagTreeItem> {
 
       const references: TagTreeItem[] = element.notes
         .map(({ uri }) => this.foam.workspace.get(uri))
+        .filter(note =>
+          // foreach nested tag item check if the exact tag is present as a tag on the note. If so,
+          // it belongs to a nested tag item and should not be displayed as a reference underneath the current element
+          isNone(nestedTagItems.find(item => note.tags.has(item.tag)))
```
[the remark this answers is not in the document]
pderaaij: I've been looking why the problem exists anyway. Turns out I've added all the complexity because I didn't assign the right tag to the parent item. With the right tag on that element, it is easier to filter for matches. Should be more readable now and function as it should.

riccardoferretti:
Sounds good, I agree it looks better now. 
Can you still add the use case I mentioned in the previous comment? It's an interesting edge case to monitor

pderaaij:
Well, that is an interesting one. Do you expect a that takes a combination of the markdown parser and the tag tree view. Or just the tags on a note that properly display in the tag explorer?

riccardoferretti:
> Well, that is an interesting one. Do you expect a that takes a combination of the markdown parser and the tag tree view. Or just the tags on a note that properly display in the tag explorer?

I am not sure I understand what you mean, but if a file has both `#parent` and `#parent/child` I would expect the note to show up as a reference under both the `parent` item and the `child` item (which itself belongs to the `parent` item).
Something like, assuming the file is called `my-note.md`:

```
parent
- child
  - my-note.md
- my-note.md
```

makes sense?

packages/foam-vscode/src/features/tags-tree-view/index.spec.ts
```diff
@@ -68,7 +68,6 @@ describe('Tags tree panel', () => {
     const parentTreeItems = (await provider.getChildren()) as Tag[];
     const parentTagItem = parentTreeItems.pop();
     expect(parentTagItem.title).toEqual('parent');
-    expect(parentTagItem.title).not.toEqual('child');
```
riccardoferretti, line 71: a comment about line 34 of this file, where you initialize `_foam`. 
I feel this pattern (which IIRC you picked up from`backlinks.spec.ts` might not apply very well here.
The reason is that in that file, `foam` was used for read only.
In these tests you are changing the state of the `_foam` object and this makes things harder to track.
I would suggest one of these approaches:

1. move the foam bootstrap (and disposal) into every test
2. move the foam bootstrap into `beforeEach` and disposal into `afterEach`

Of course happy to hear other approach if you prefer a different take.

This will allow in each test to check exactly the state you would expect, without risk of extra objects polluting your test workspace.

Thoughts?

riccardoferretti (commented):
I have left a comment to make the tests more reliable.
the new code to handle parent/child looks really good to me.
Once the tests are fixed, I am happy to merge the PR!

riccardoferretti (approved):
great work, thanks!

PR #666, opened 2021-06-04
WIP: Move selected text into a new note and auto create a backlink

### LINKED ISSUE #653: A way to move selected text into a newly created note
# Feature request

I find myself writing some text in a note and then deciding that that text should live in a new note (a .md file with a random string as name). Currently I perform the following steps to do this. I:

1. generate a short random string
2. run Foam: create new note with the random string as file name and title
3. create a backlink to the note that I want to move the text from
4. navigate the backlink to the note that I want to move the text from
5. highlight and cut the text
6. navigate back to the new note
7. insert the cut text in the new note

I would like to have a shortcut that will perform steps 1.-7. on a highlighted piece of text.

pderaaij:
This pr resolves #653 , the suggestion adds the functionality to create a new note for a piece of selected text. It will use the functionalities as introduced by @movermeyer to create a new note from the default template or select a template. I've added a configuration to configure foam to always select a note template if desired. Setting it false means it will use the default template.

To keep behaviour similar to the other two new note command I am reusing that code as much as possible. It simply adds some additional text and view behaviour to the existing methods. This should keep the user interaction consistent and will reduce maintenance work. I need to add tests and do some more manual testing, but first want to validate the approach with @movermeyer and @riccardoferretti. You are welcome to leave your suggestions and comments.

note-selection


packages/foam-vscode/package.json
```diff
@@ -342,6 +347,11 @@
           "type": "object",
           "description": "Custom graph styling settings. An example is present in the documentation.",
           "default": {}
+        },
+        "foam.selection.newNote.askForNoteTemplate": {
+          "description": "When creating a note from a text selection ask for template to be used",
+          "type": "boolean",
```
riccardoferretti, line 353: generally speaking I am not a big fan of boolean options as they allow for little future flexibility (unless the option is clearly bound).
What do you think of a property with the following options? 

```
"foam.createNoteFromSelection.templateToUse" : {
          "type": "string",
          "default": "use new-note template",
          "enum": ['ask', 'use new-note template', 'use new-note-from-selection template'],
          ...
}
```

I am not feeling very strongly about the above, I am happy to explore the approaches here

riccardoferretti: an interesting thing here is that we could use the same pattern with "create new notes from (wiki)links"

packages/foam-vscode/src/utils.ts
```diff
@@ -169,9 +170,16 @@ export function isNone<T>(
   return value == null; // eslint-disable-line
 }
 
-export async function focusNote(notePath: URI, moveCursorToEnd: boolean) {
+export async function focusNote(
+  notePath: URI,
+  moveCursorToEnd: boolean,
+  nextToCurrentEditor?: boolean
```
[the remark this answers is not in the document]
riccardoferretti: (this comment applies to all the places where you have used `nextToCurrentEditor`)

riccardoferretti (commented):
this looks pretty good @pderaaij! I have left some food for thought, let me know what you think

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -347,6 +355,10 @@ async function createNoteFromDefaultTemplate(): Promise<void> {
   ] = extractFoamTemplateFrontmatterMetadata(templateWithResolvedVariables);
   const templateSnippet = new SnippetString(templateWithFoamFrontmatterRemoved);
 
+  if (additionalNoteText) {
+    templateSnippet.appendText(additionalNoteText);
```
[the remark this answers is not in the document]
pderaaij: I think I could implement it as suggested, but please have a look at it :) I need to read into the VSCode API to create some solid tests. Could not find how to manipulate the selection and document part of an editor.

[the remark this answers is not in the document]
riccardoferretti: cc @movermeyer as he has been championing the resolution and I want to be sure he is in the loop

movermeyer (commented):
Generally looking good! It's exciting to see how other features can build on top of the new templates patterns. 🎉 

---

I have some questions about error handling patterns. Generally when something is unable to complete the task that the user asked it to, I'd expect some type of flash message to appear in the bottom right of VSCode. I'd expect that we'd want something similar in the cases I've noted in comments. cc @riccardoferretti 

---

When you select a template that doesn't have `$FOAM_CONTENT`, your selection contents are quietly deleted. 🙀 

Perhaps users should only see templates in the picker that use `$FOAM_CONTENT` (showing a prompt in the case of an empty picker)?
And perhaps the command should fail if `ask-for-template` is `false` and `new-note.md` does not use `$FOAM_CONTENT`?

IDK. I suspect some more thought may be needed around the UX in these cases. cc @riccardoferretti 

---

We also need to document `$FOAM_CONTENT` in `docs/features/note-templates.md`.

pderaaij:
> I have some questions about error handling patterns. Generally when something is unable to complete the task that the user asked it to, I'd expect some type of flash message to appear in the bottom right of VSCode. I'd expect that we'd want something similar in the cases I've noted in comments. cc @riccardoferretti
> 
> When you select a template that doesn't have `$FOAM_CONTENT`, your selection contents are quietly deleted. 🙀
> 
> Perhaps users should only see templates in the picker that use `$FOAM_CONTENT` (showing a prompt in the case of an empty picker)?
> And perhaps the command should fail if `ask-for-template` is `false` and `new-note.md` does not use `$FOAM_CONTENT`?
> 
> IDK. I suspect some more thought may be needed around the UX in these cases. cc @riccardoferretti
> 
> We also need to document `$FOAM_CONTENT` in `docs/features/note-templates.md`.

Thanks for the feedback! Definitely needs some more testing and exploration, but good to the general line is okay. I'll have a look at things this week.

Starting this weekend, I need to focus for a couple of weeks on a study assignment next to my day job, so time for this will be limited. Feel free, to co-create on the PR and tune things. Happy to see this being finished in a collaborative approach.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -317,7 +326,10 @@ export function determineDefaultFilepath(
   return defaultFilepath;
 }
 
-async function createNoteFromDefaultTemplate(): Promise<void> {
+async function createNoteFromDefaultTemplate(
+  additionalNoteText?: string,
```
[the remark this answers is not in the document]
riccardoferretti: probably it should just match the name of the variable, in our case `foamContent`

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -450,6 +469,50 @@ async function createNewTemplate(): Promise<void> {
...
+  if (!editor) {
+    return;
+  }
+
+  const document = editor.document;
+  const selection = editor.selection;
+
+  if (!document || selection.isEmpty) {
+    return;
+  }
```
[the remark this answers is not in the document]
riccardoferretti: that's a very good point actually

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -450,6 +469,50 @@ async function createNewTemplate(): Promise<void> {
   await focusNote(filenameURI, false);
 }
 
+async function createNoteFromSelection(): Promise<void> {
+  const editor = window.activeTextEditor;
+
+  if (!editor) {
+    return;
+  }
```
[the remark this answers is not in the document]
riccardoferretti: same as above

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -450,6 +469,50 @@ async function createNewTemplate(): Promise<void> {
...
+
+  const newNoteFile = window.activeTextEditor.document.fileName;
+
+  if (newNoteFile === document.fileName) {
+    return;
+  }
+
+  const newNoteTitle = newNoteFile
+    .substr(newNoteFile.lastIndexOf('/') + 1)
+    .replace('.md', '');
```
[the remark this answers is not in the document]
riccardoferretti: yeah I think that's a valid comment, I feel there needs to be a bit of a refactoring around the encoding/decoding of note title to file path. this is a bigger topic, but something like that would:
- consolidate the relationship between note title and file name (and path) and possibly wikilink
- make it easier to use `.mdx` as extension (we currently hardcode `.md` in a bunch of places)
- possibly (to be _heavily_ tested) even allow for custom translation functions (for the hardcore supporters of slugification, or even to fix some of the issues around case sensitivity for links/paths)

I feel this is a bigger project, but the specific point about OS specific separator applies

pderaaij:
I've made a start on processing the comments. I am a bit wondering what to do with the behaviour of the missing FOAM_CONTENT variable. We can check it when we try to resolve it, and when missing abort and throw an error message. 

Or we could prevent it from showing the templates at all, but that might give more confusion. What do we prefer here?

riccardoferretti:
Let me explicitly address the previous comments:

> I have some questions about error handling patterns. Generally when something is unable to complete the task that the user asked it to, I'd expect some type of flash message to appear in the bottom right of VSCode. I'd expect that we'd want something similar in the cases I've noted in comments. cc @riccardoferretti

I agree with this pattern. I think we don't follow it in several places (there has been "no pattern" around this behavior so far). Let's start with this PR (only for the flow directly affected by the code changes of course)

> When you select a template that doesn't have $FOAM_CONTENT, your selection contents are quietly deleted. 🙀
> 
> Perhaps users should only see templates in the picker that use $FOAM_CONTENT (showing a prompt in the case of an empty picker)?
> And perhaps the command should fail if ask-for-template is false and new-note.md does not use $FOAM_CONTENT?

My take here is that in this case the $FOAM_CONTENT should be considered to be at the end of the created note.
My reasoning is the following:

1. if FOAM_CONTENT is set, it's assumed that the template is being called with some "additional text" (to use the name from the code), or in general some text that is expected to live in the body of the note
2. if FOAM_CONTENT is used by the template but the variable is not set, it will default to an empty string (and therefore the variable will simply be replaced with nothing)
3. this would also make the new note template automatically compatible with the "New note from selection" command

The case not supported here is someone explicitly wanting to avoid the FOAM_CONTENT when using a template called with it.
This feels like it "shouldn't happen" because FOAM_CONTENT is only set by the extension itself in those use cases when we want some text in the template.
Thoughts?

pderaaij:
That basically means we assume FOAM_CONTENT is present in the template? If not, then we add the content at the end of the note. Only if there is any content to be set of course.

Or am I missing something?

movermeyer:
Oops. I missed [this comment](https://github.com/foambubble/foam/pull/666#discussion_r645877812) from Riccardo where they already had suggested how to handle templates that didn't use `$FOAM_CONTENT`. 👍 

I agree. Adding an implicit `$FOAM_CONTENT` at the end of the template in the case where a value for `$FOAM_CONTENT` is provided seems like a very good solution.

|                              | `$FOAM_CONTENT` in template                                                                                         | `$FOAM_CONTENT` NOT in template                                                                                                                                                                                                     |
| ---------------------------- | ------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `$FOAM_CONTENT` provided in `givenValues`     | Each instance of `$FOAM_CONTENT` in the template is replaced with the value (Handled by `substituteFoamVariables` and `givenValues`). | `$FOAM_CONTENT` is appended at the end of the template (perhaps after a blank line). i.e. Templates that don't mention `$FOAM_CONTENT` get an implicit `\n$FOAM_CONTENT\n` at the end, when a value for `$FOAM_CONTENT` is provided |
| `$FOAM_CONTENT` NOT provided in `givenValues` | `$FOAM_CONTENT` is resolved to empty string (possibly with `async function resolveFoamContent() {return "";}` called from `Resolver.resolve`)      | Nothing special happens                                                                                                                                                                                                             |

---

Personally, I [don't like the name](https://github.com/foambubble/foam/pull/666#discussion_r647360891) `$FOAM_CONTENT`. IMO, it's too generic. That said, I don't have a better name... `$FOAM_SELECTION_CONTENT`?

Which then raises the question.. What if creating a new note always used your current selection as `$FOAM_SELECTION_CONTENT`, (and replaced it with a backlink)? 

Would we actually want/need this behaviour to be a separate command at all? I'd be interested in trying that out. 

There is an maintenance/cognitive overhead to having more commands built into Foam (User asks "When should I use this command vs the other? What's the difference?") 

Perhaps this selection functionality could just be the standard behaviour of `Foam: Create New Note` and `Create New Note From Template`?

---

There is already a [`$TM_SELECTED_TEXT` VSCode snippet variable](https://code.visualstudio.com/docs/editor/userdefinedsnippets):

> `TM_SELECTED_TEXT` The currently selected text or the empty string

Perhaps we shouldn't be defining our own at all, but instead making sure that `$TM_SELECTED_TEXT` is working properly with `window.activeTextEditor.insertSnippet`, and then creating the backlink.

riccardoferretti:
> Which then raises the question.. What if creating a new note always used your current selection as `$FOAM_SELECTION_CONTENT`, (and replaced it with a backlink)?
> 
> Would we actually want/need this behaviour to be a separate command at all? I'd be interested in trying that out.

This is a very interesting idea, and at first glance I like it a lot, with one caveat: we should only have one command.
Basically this pattern could "swallow" the `Create new note from selection`, because creating a new note when there is a selection will already do it - always.
Thoughts? 
 
> There is an maintenance/cognitive overhead to having more commands built into Foam (User asks "When should I use this command vs the other? What's the difference?")

Well, that's why the caveat, I feel this actually reduces the overhead and it follows a simple rule.
Only possible drawback is users finding this pattern counter-intuitive (after having discovered it of course)
 
> Perhaps this selection functionality could just be the standard behaviour of `Foam: Create New Note` and `Create New Note From Template`?

I think this mean we are saying the same thing? Or did you mean something different?
 
> Perhaps we shouldn't be defining our own at all, but instead making sure that `$TM_SELECTED_TEXT` is working properly with `window.activeTextEditor.insertSnippet`, and then creating the backlink.

I agree 100%, let's use `$TM_SELECTED_TEXT` (which, tbh, I dislike as a name, but the pattern of using VS Code snippet variables when available matters more).

movermeyer:
> I think this mean we are saying the same thing?

@riccardoferretti I think we are saying the same thing. The behaviour of the two existing note creation commands (`Foam: Create New Note and`Foam: Create New Note From Template`) could be extended to:

1. Produce the backlink to the new note in the original note 
2. and provide the selection text to the template.

If we get `$TM_SELECTED_TEXT` working, then we won't even need step 2. At any rate, I suspect we wouldn't need a separate `Foam: Create New Note From Selection`.

riccardoferretti:
ok, this sounds like a good direction to me. @pderaaij thoughts?

pderaaij:
Could be a promising option, but a short test on my machine failed. When I put `${TM_SELECTED_TEXT}` in a template, select some text and then create a new note nothing happens. So probably more is required to make it work. As said, I lack time for the upcoming weeks due to a time-consuming study assignment next to my day-job. Happy if someone else has a look, otherwise it will have to wait a couple of weeks.

bronson:
TM_CURRENT_LINE and TM_CURRENT_WORD are also empty when creating a new note from a template. All the other VS Code variables appear to be filled in properly.

riccardoferretti:
thanks for reporting this @bronson, this is pretty odd, I wonder why that is. Could you please create an issue with VS Code to understand if it's a bug or there is something we are missing?

pderaaij:
My unvalidated hypothesis is that these variables might only work in the same editor/document. I simply have no time to validate it. But, I can imagine that these work perfectly as snippet template in the file you are editing, but not an other document.

One way to test it, I think, is to create a custom snippet with one of the variables. Test it by inserting the snippet in the same file and in the new file and see what happens.

pderaaij:
Did a quick test with a global snippet defined as:

```
"test selected text": {
		"body": "Test: ${TM_SELECTED_TEXT}",
		"description": "Test is selected text works in a new file"
	}
```

Works when I do it in the same file, but not when I try to insert it in a new file.

Looking at the source code of vscode it shows it passes (line 476) on a [text model which is based on the active editor](https://github.com/microsoft/vscode/blob/94c9ea46838a9a619aeafb7e8afd1170c967bb55/src/vs/editor/contrib/snippet/snippetSession.ts#L416). Feels for me the `${TM_SELECTED_TEXT}` will not fly for this feature. But, I might be missing something here.

@movermeyer @riccardoferretti any thoughts?

riccardoferretti:
My temptation would be to still populate `TM_SELECTED_TEXT` so that users can use the same variable name, seems fairly intuitive to me; I am concerned about side effects where this would clash with the original variable, but I can't see how, given they would match even if we would run into this issue. 
If that is confusing or problematic, we could use `FOAM_SELECTED_TEXT`

pderaaij:
I would prefer using `FOAM_SELECTED_TEXT` to prevent interference and being explicit about the behaviour. If we would follow this route the implementation has to change as follows:

- remove the new note via selection command
- allow for selection content in the current two commands
- rename current var to `FOAM_SELECTED_TEXT`
- Update docs

movermeyer:
I agree: we **should not** use `TM_SELECTED_TEXT`, for the reasons outlined.

riccardoferretti:
sounds good to me, we have a plan 🚀

pderaaij:
@riccardoferretti @movermeyer This should be it if I see it well. Please have a look and add your thoughts.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -295,6 +311,39 @@ function currentDirectoryFilepath(filename: string) {
   return Uri.joinPath(currentDir, filename);
 }
 
+function findSelectionContent(): FoamSelectionContent {
```
riccardoferretti, line 314: is this signature correct? shouldn't it include `| undefined`?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -383,12 +452,21 @@ async function createNoteFromTemplate(
...
     [
       resolvedValues,
       templateWithResolvedVariables,
-    ] = await resolveFoamTemplateVariables(templateText);
+    ] = await resolveFoamTemplateVariables(
+      templateText,
+      new Set(),
+      new Map().set(
+        'FOAM_SELECTION_CONTENT',
+        selectedContent !== undefined ? selectedContent.content : ''
```
riccardoferretti, line 467: minor: here we could use `selectedContent?.content ?? ''` which is the short version of what you wrote

riccardoferretti (approved):
This is looking good.
I have left a couple of minor comments, but looks great in general.
I will test it out then merge it.

bronson:
Just tried it out. It works great, including undo. Impressive work @pderaaij.

One thing: if the template doesn't include FOAM_SELECTION_CONTENT, the user's selected text just disappears. Earlier in this issue's discussion it was suggested that, if the template doesn't include FOAM_SELECTION_CONTENT, that the selection be appended to the template.

Also, this feature is not very discoverable... The default template includes `${FOAM_SELECTION_CONTENT}`, but people will have to update all their other templates by hand.

To address this, here are some thoughts on helping users make this change...

* It merits a stout entry in the CHANGELOG, including the changes that foam suggests users make to their template files.
* Should `${FOAM_SELECTION_CONTENT}` and a quick description of the feature be added to [your-first-template.md](https://github.com/foambubble/foam-template/blob/master/.foam/templates/your-first-template.md)?
* Maybe add a blurb on how it works to [foam-template / note-templates](https://github.com/foambubble/foam-template/blob/master/docs/features/note-templates.md) and [foam / docs / note-templates.md](https://github.com/foambubble/foam/blob/master/docs/features/note-templates.md)?

After this feature lands, pretty much every template should now include FOAM_SELECTION_CONTENT shouldn't it? Is there ever a reason not to?

I can munge together some PRs if people agree with any of these.

movermeyer:
> if the template doesn't include `FOAM_SELECTION_CONTENT`, the user's selected text just disappears

Yeah, we should fix that up, as discussed.

> After this feature lands, pretty much every template should now include FOAM_SELECTION_CONTENT shouldn't it?

I disagree. If every [template has an implicit](https://github.com/foambubble/foam/pull/666#issuecomment-859706092) `FOAM_SELECTION_CONTENT` at the end, then it will only be those who need to customize the position within the template that will ever need to use it.

> people will have to update all their other templates by hand.
> ...
> It merits a stout entry in the CHANGELOG, including the changes that foam suggests users make to their template files.

The implicit `FOAM_SELECTION_CONTENT` removes any demand on users to change their templates, and significantly reduces the criticality of getting the information out.

IMO, standard documentation would be fine. All your suggestions of places to document it seem fine. It's just not nearly as critical as it would be if we were expecting people to always remember to include it.

(IMO, mentioning it in [`foam-template/note-templates`](https://github.com/foambubble/foam-template/blob/master/docs/features/note-templates.md) would probably be overkill??)

pderaaij:
> if the template doesn't include FOAM_SELECTION_CONTENT, the user's selected text just disappears

Hmmz,  I missed something here after my latest refactorings. Not sure what happened, but thanks for catching it. Sorry about that. Will have a look at it somewhere today or tomorrow.

pderaaij:
I've added the implicit behaviour. If there is selected content and the variables is not resolved, the content will be appended at the end of the note.

riccardoferretti:
@pderaaij two things I just thought of from the last commit:

- I just noticed that the variable name is `FOAM_SELECTION_CONTENT`, but we went for `FOAM_SELECTED_TEXT` in the comments above. Given this becomes part of the template API and it will be hard to change it afterwards, so let's fix it here as `FOAM_SELECTED_TEXT`

- in your latest commit, I saw you implemented the "implicit behavior" in the top level functions. Have you considered doing that in `resolveFoamTemplateVariables`? I think it might have a couple of advantages:
  1. it becomes more easily testable (and hence "documented") in `create-from-template.test.ts`
  2. I feel it better describes the logic flow (in a way, we are simply adding the variable at the end of the template if it doesn't exist, and if the variable is not defined it will simply be resolved with empty text)
  3. minor: it consolidates the behavior (no need to write it twice)

pderaaij:
@riccardoferretti I've attempted to implement it as suggested. Please have a look at it if it matches expectations.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -283,6 +283,17 @@ export async function resolveFoamTemplateVariables(
   );
   const uniqVariables = [...new Set(variables)];
 
+  const missingFoamVariables = [...givenValues.keys()].filter(
+    x => uniqVariables.indexOf(x) < 0
+  );
+  if (missingFoamVariables.length > 0) {
+    missingFoamVariables.forEach(key => {
+      templateText = `${templateText}
+      \${${key}}`;
```
riccardoferretti, line 292: do I understand correctly that you are appending all un-referenced variables to the template?
This behavior should only occur for `FOAM_SELECTED_TEXT`, we treat this variable a "special case" in this regard

riccardoferretti (commented):
This looks good, the only change I think we need is appending only the `FOAM_SELECTED_TEXT` variable to the template, not all un-referenced variables.
Once that is done, this PR is good to be merged! Thanks!!

packages/foam-vscode/src/features/create-from-template.ts
pderaaij: Changed the behaviour. In hindsight it indeed makes more sense to only implicitly add `FOAM_SELECTED_TEXT`.

packages/foam-vscode/src/features/create-from-template.test.ts
```diff
@@ -149,6 +149,38 @@ describe('resolveFoamTemplateVariables', () => {
       await resolveFoamTemplateVariables(input, new Set(['FOAM_TITLE']))
     ).toEqual(expected);
   });
+
+  test('Correctly substitutes given variables that are missing in the template', async () => {
```
riccardoferretti, line 153:
```suggestion
  test('Adds FOAM_SELECTED_TEXT at end of template if it has variable has a value but is not referenced', async () => {
```

Also, we should have a test for when the `FOAM_SELECTED_TEXT` is used in the template but no text was selected (expected behavior: no content added. the variable should default to empty string).

Asking as by looking at the code I was unclear how that case would be handled

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -263,14 +275,28 @@ async function askUserForFilepathConfirmation(
...
   const variables = findFoamVariables(templateText.toString()).concat(
     ...extraVariablesToResolve
   );
   const uniqVariables = [...new Set(variables)];
 
+  const missingFoamVariables = [...givenValues.keys()].filter(
+    x => uniqVariables.indexOf(x) < 0
+  );
+  if (
+    missingFoamVariables.length > 0 &&
```
riccardoferretti, line 290: minor: isn't this implicit by the next test condition?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -263,14 +275,28 @@ async function askUserForFilepathConfirmation(
...
+    missingFoamVariables.includes('FOAM_SELECTED_TEXT')
```
riccardoferretti, line 291: here we are not checking that an actual value exists for `FOAM_SELECTED_TEXT`.
Normally in templates a non-existing variable is simply left there untouched, in our case we would want the text to be empty.
So, we either:
- default `FOAM_SELECTED_TEXT` to the empty string (my personal favorite option, but I am open to disagreements)
- only append `FOAM_SELECTED_TEXT` in the template if it's also defined in the provided variables

what do you think

riccardoferretti (commented):
I have left a couple more questions for you
I promise we are getting there ;)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -295,6 +325,39 @@ function currentDirectoryFilepath(filename: string) {
...
+    selection,
+    content: document.getText(selection),
+  };
+}
+
+async function replaceSelectionWithWikiLink(
+  document: TextDocument,
+  selection: Selection
+) {
+  const newNoteFile = window.activeTextEditor.document.fileName;
```
movermeyer, line 352: We already have this filepath computed (we pass it into `writeTemplate`).

Instead of relying on the behaviour of VS Code in this line, let's pass the filepath into `replaceSelectionWithWikiLink` and use that.

pderaaij:
@movermeyer @riccardoferretti Thanks for the feedback. I was a bit too eager to get this merged in whilst being busy with other stuff. So didn't put enough attention in. Processed the feedback and squashed the commits.

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -263,14 +276,29 @@ async function askUserForFilepathConfirmation(
...
   const variables = findFoamVariables(templateText.toString()).concat(
     ...extraVariablesToResolve
   );
   const uniqVariables = [...new Set(variables)];
 
+  const missingFoamVariables = [...givenValues.keys()].filter(
+    x => uniqVariables.indexOf(x) < 0
+  );
+  if (missingFoamVariables.includes('FOAM_SELECTED_TEXT')) {
+    missingFoamVariables.forEach(key => {
```
riccardoferretti, line 291: would this add `FOAM_SELECTED_TEXT` once per missing variable? So if there are 3 missing variables we would end up with the selection 3 times in the new note?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -263,14 +276,29 @@ async function askUserForFilepathConfirmation(
...
+      templateText = `${templateText}
+      \${FOAM_SELECTED_TEXT}`;
+      uniqVariables.push('FOAM_SELECTED_TEXT');
+
+      if (!givenValues.get('FOAM_SELECTED_TEXT')) {
+        givenValues.set('FOAM_SELECTED_TEXT', '');
+      }
```
riccardoferretti, line 298: minor: I wonder if for better readibility we should separate the "`FOAM_SELECTED_TEXT` defaults to ''" from the "if the template doesn't include `FOAM_SELECTED_TEXT` append it". 
In that perspective, maybe you could just add `FOAM_SELECTED_TEXT` to the `variables` (if it doesn't already exist) right before computing `uniqVariables`?

packages/foam-vscode/src/features/create-from-template.test.ts
riccardoferretti: > we should have a test for when the FOAM_SELECTED_TEXT is used in the template but no text was selected (expected behavior: no content added. the variable should default to empty string).

Was this test implemented? I can't find it

packages/foam-vscode/src/features/create-from-template.ts
pderaaij: Ow darn.. sorry 'bout that. Will change it!

packages/foam-vscode/src/features/create-from-template.ts
pderaaij: I feel this won't work. Adding it to variables would break the determination if it exists as a FOAM variable. I did extract it from the if-loop. What do you think?

riccardoferretti (approved):
LGTM! Thanks for the great work!

PR #674, opened 2021-06-11
doc: describe how to select launch configuration

bronson:
It took me a while to figure out why my extension host was quitting on its own soon after I launched it. It's because my launch config defaulted to "Test VSCode Extension". Hope this will save others the time.


riccardoferretti (approved):
thanks for updating the comments here, I think the launch configurations more in general need a bit of love, better naming, and a bit more structure as well - we'll get to that :)

Issue #670, opened 2021-06-06
`SAME_AS_ACTIVE_NOTE` and new note template issue


JakubGIT:
- Foam version:  0.13.7
- Markdown Notes version:  0.0.24
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**
- `"vscodeMarkdownNotes.newNoteDirectory": "SAME_AS_ACTIVE_NOTE"` doesn't work, the new notes are created in both places `WORKSPACE_ROOT` and `SAME_AS_ACTIVE_NOTE`
- `"vscodeMarkdownNotes.newNoteTemplate": "# ${noteName}"` doesn't work properly, the new note contains `# <file name>`

- with `foam.foam-vscode` 0.9.1 and `kortina.vscode-markdown-notes` 0.0.21 it works like a charm!! a few versions ago it stopped working so I don't update any more

**Steps to reproduce**
1. download, unpack and open the zipped master of foam
2. install the plugins if needed
3. create a new folder `temp`
4. create a new file `temp/temp.md`
5. in the new file type `- [[new-note]]`
6. click on the newly written link
7. there are two new files `new-note.md`, one in root, one in `temp`
8. none of them has the first line `# New Note'

**Additional information**
I don't have any custom settings, all left default

riccardoferretti:
The reason why you see two notes being created is that one is done by markdown notes, the other by Foam.

Most of the features of Markdown Notes have been moved into Foam itself (mostly from 0.12.0), so you don't need both extensions, and in this case they might actually conflict. If you want to keep Markdown Notes and Foam together you could use < 0.12.0

ashleemboyer:
@riccardoferretti is there a setting similar to `"vscodeMarkdownNotes.newNoteDirectory"` to tell Foam where we want the new file created?

riccardoferretti:
this is addressed by the template metadata (see https://github.com/foambubble/foam/issues/563) and Template V2 related issues.

Although I wonder whether a simpler setting might be useful - happy to consider it if good arguments can be made for it (and how it would interact with the template settings)

cc @movermeyer and @pderaaij in case they have comments on this

movermeyer:
@ashleemboyer

Yeah, the [`filepath` metadata](https://github.com/foambubble/foam/blob/master/docs/features/note-templates.md#metadata) should do what you want.

---

[...]

PR #679, opened 2021-06-13
WIP: Allow templates to be defined in the user's home directory

movermeyer:
### WIP Status

* [ ] Add tests
* [ ] Also manually test to confirm things work on Windows
* [ ] Update the code to match the outcome from [this proposal](https://github.com/foambubble/foam/pull/683)

As noted in https://github.com/foambubble/foam/issues/670#issuecomment-860092115, you cannot specify "global" templates, even though you should be able to.

I refactored the existing template commands to fallback to using the templates in the user's home directory.

* `Foam: Create New Note`:
    * Looks for `new-note.md` in the current workspace, then the user's home directory if not found in the current workspace.

* `Foam: Create New Note From Template`:
    * Now has a `Show templates from your home directory` option in the quick-pick that appears when the current workspace has templates

image

* `Foam: Create New Template`:
    * Asks the user where they would like to create the new template

image
 
There were a bunch of refactoring steps in this PR. I've broken them out by commit.
Reviewing this PR commit by commit might be easier.

### DIFF
--- a/packages/foam-vscode/src/features/create-from-template.ts
+++ b/packages/foam-vscode/src/features/create-from-template.ts
@@ -11,15 +11,10 @@ import * as path from 'path';
 import { FoamFeature } from '../types';
 import { TextEncoder } from 'util';
 import { focusNote } from '../utils';
-import { existsSync } from 'fs';
+import { homedir } from 'os';
+import { existsSync, readdirSync } from 'fs';
 import { isAbsolute } from 'path';
 import { extractFoamTemplateFrontmatterMetadata } from '../utils/template-frontmatter-parser';
-
-const templatesDir = Uri.joinPath(
-  workspace.workspaceFolders[0].uri,
-  '.foam',
-  'templates'
-);
 
 export class UserCancelledOperation extends Error {
   constructor(message?: string) {
@@ -32,6 +27,7 @@ export class UserCancelledOperation extends Error {
 
 const knownFoamVariables = new Set(['FOAM_TITLE']);
 
+const defaultTemplateFilename = 'new-note.md';
 const defaultTemplateDefaultText: string = `---
 foam_template:
   name: New Note
@@ -39,7 +35,6 @@ foam_template:
 ---
 # \${FOAM_TITLE}
 `;
-const defaultTemplateUri = Uri.joinPath(templatesDir, 'new-note.md');
 
 const templateContent = `# \${1:$TM_FILENAME_BASE}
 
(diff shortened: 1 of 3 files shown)


riccardoferretti:
I have a comment at the functionality level, have you considered merging the list of local/global snippets so that there is a single step for selection?
Maybe next to the file name we can add whether the template is `workspace` specific or `global`

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -186,46 +193,41 @@ function sortTemplatesMetadata(
...
-  const items: QuickPickItem[] = await Promise.all(
+  return await Promise.all(
     templatesMetadata.map(metadata => {
-      const label = metadata.get('name') || metadata.get('templatePath');
-      const description = metadata.get('name')
-        ? metadata.get('templatePath')
-        : null;
+      const basename = path.basename(metadata.get('templatePath'));
+      const label = metadata.get('name') || basename;
+      const description = metadata.get('name') ? basename : null;
```
riccardoferretti, line 224: in my idea, here we could add whether the template is local or global

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -285,30 +336,33 @@ async function writeTemplate(templateSnippet: SnippetString, filepath: Uri) {
...
+  const filepathFromMetadata = templateMetadata.get('filepath');
+  if (filepathFromMetadata) {
     if (isAbsolute(filepathFromMetadata)) {
       defaultFilepath = Uri.file(filepathFromMetadata);
     } else {
-      defaultFilepath = Uri.joinPath(
-        workspace.workspaceFolders[0].uri,
-        filepathFromMetadata
-      );
+      defaultFilepath = workspace.workspaceFolders
```
riccardoferretti, line 363: would the resolution here match the one of `foam.newNoteDirectory`?

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -407,29 +459,64 @@ async function createNoteFromTemplate(
...
   }
-  const filepathURI = Uri.file(filepath);
-  await writeTemplate(templateSnippet, filepathURI);
+  const filepathUri = Uri.file(filepath);
+  await writeTemplate(templateSnippet, filepathUri);
+}
+
+function askTemplateCreationLocation(): Thenable<QuickPickItem | undefined> {
+  const homeDirQuickPick = {
+    label: 'In your home directory',
```
riccardoferretti, line 477: I like that you show the user where to create the template. A couple of things:
- to make things as simple as possible for the user, should we nudge them in a direction? 
I would say workspace template as it feels like the more standard case, but open to other opinions.

- what about the terminology? I would be keen to focus on functionality (global vs local/workspace templates) vs implementation (that global templates live in the user's directory).  
Not sure whether we should say `workspace` vs `global` vs `device`. We want to convey that `workspace` templates are local, but also portable with the workspace. `global`/`device` templates can be reused on the same machine, but do live on the machine (well, actually on the user, so maybe we should use `user` and mimic VS Code terminology: `Workspace` and `User`)

packages/foam-vscode/src/features/create-from-template.ts
```diff
@@ -407,29 +459,64 @@ async function createNoteFromTemplate(
...
+  const templateLocationSelection = await askTemplateCreationLocation();
+  if (templateLocationSelection === undefined) {
+    return;
+  }
+
+  const defaultFilepath = Uri.joinPath(
+    Uri.file((templateLocationSelection as QuickPickItem).detail),
+    defaultFilename
+  ).fsPath;
+  const defaultFilepathNoExtension = defaultFilepath.replace(/\.[^.]+$/, '');
```
riccardoferretti, line 513: I have seen this in a couple of places, plus I think it could be useful in general.. how about moving it to `URI` (in `foam-core`?

riccardoferretti (commented):
The code looks good, I have added a couple of comments about UX, let me know what you think

Acumane:
What happened? Is this feature still planned?

riccardoferretti:
We decided to scrap the feature for the time being

PR #688, opened 2021-06-24
Handle different capitalizations of wikilinks and files

### LINKED ISSUE #641: Create Note deletes previous content if wikilink has different capitalization than Markdown file
- Foam version: v0.13.3
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**
Mismatching wikilink vs filename capitalization causes file contents to be cleared.

If [[WIKI-links]] has capitalized letters (like that one), then CTRL+click will **clear existing content** of file `wiki-links.md` if it exists. If filename is `WIKI-links.md` CTRL+click works ok - opens file.
Or if [[wiki-links]] has no caps, but the filename is for example `wiki-LINKs.md`, then CTRL+click will clear file contents.

I just lost all my TODO notes. I'm using [[TODO-project]] to link different TODO's in a project, but I don't like
caps in filenames. Hence the file was `todo-project.md`.

I don't think this has happened from the beginning (used this for 1 month'ish), because I'm sure CTRL+click just worked at some point.

Point of NOTE: I don't have Markdown All-in-one installed, because it has a bug where **it disables the Enter and Backspace keys in .md files**. I have never had an issue with that extension before, but.. for some reason it started misbehaving.

Markdown All-in-one issues:

* This issue: https://github.com/yzhang-gh/vscode-markdown/issues/335
* Issue with more discussion: https://github.com/yzhang-gh/vscode-markdown/issues/95

OSX: I have also OSX laptop where I have NOT encountered this issue. Also.. I have used Foam for month+ on this Win10 machine, and.. all of a sudden this bug appears. Maybe some update, maybe disabling an extension, maybe some setting changed?
_[ EDIT: Issue reproes on OSX with a slight twist. See my next comment below. ]_

NOTE 2: Funnily when CTRL+hovering over the [[WIKI-links]] it shows the actual contents from [[wiki-links.md]], but clicking will still override the contents.

**Steps to reproduce**
1. Clone foam-template repo
2. (have what I have on my machine now -  I might try to repro this from fresh install, but can't right now)
3. Edit `inbox.md` to have 2 wikilinks: [[DOG-cat]]
4. CTRL+click this wikilink
5. Foam creates file `dog-cat.md`
6. CTRL+click the wikilink again
7. Foam clears file contents, and.. actually keeps adding the Title again and again.

**Additional information**
* Attached DEBUG logs from Starting in foam-templates dir, and Clicking on "Smurfs" to override `smurfs.md` contents.
  - [foam_debug.log](https://github.com/foambubble/foam/files/6499335/foam_debug.log)

VSCode info:

* Version: 1.56.2 (system setup)
* Commit: 054a9295330880ed74ceaedda236253b4f39a335
* Date: 2021-05-12T17:13:13.157Z
* Electron: 12.0.4
* Chrome: 89.0.4389.114
* Node.js: 14.16.0
* V8: 8.9.255.24-electron.0
* OS: Windows_NT x64 10.0.21376

VSCode extensions:

```
code --list-extensions
ban.spellright (disabled due to it being annoying)
DavidAnson.vscode-markdownlint
eamodio.gitlens
evilz.vscode-reveal
foam.foam-vscode
goessner.mdmath
hediet.vscode-drawio
kortina.vscode-markdown-notes
ms-azuretools.vscode-docker
ms-python.python
ms-python.vscode-pylance
ms-toolsai.jupyter
ms-vscode-remote.remote-containers
ms-vscode-remote.remote-ssh
ms-vscode-remote.remote-ssh-edit
ms-vscode-remote.remote-wsl
ms-vscode-remote.vscode-remote-extensionpack
mushan.vscode-paste-image
PKief.markdown-checkbox
stuart.unique-window-colors
tomoki1207.pdf
```

pderaaij:
This should fix #641. The essence of the fix is to make sure we register and lookup resources by name in lowercase. This will ensure the key is always the same.

A risk I see is that there might be an issue with backward compatibility. If I now have a wikilink `[[Page]]` and `[[page]]` in different folders, this will clash now. What do we want to do here?

Given the spirit of wikilinks I would say that the situation of above is incorrect anyway. But, we might want to do something here? Add a warning box if the situation occurs, perhaps? Looking for thoughts here.

### DIFF
--- a/packages/foam-core/src/model/workspace.ts
+++ b/packages/foam-core/src/model/workspace.ts
@@ -26,7 +26,8 @@ const pathToResourceId = (pathValue: string) => {
 };
 const uriToResourceId = (uri: URI) => pathToResourceId(uri.path);
 
-const pathToResourceName = (pathValue: string) => path.parse(pathValue).name;
+const pathToResourceName = (pathValue: string) =>
+  path.parse(pathValue).name.toLowerCase();
 export const uriToResourceName = (uri: URI) => pathToResourceName(uri.path);
 
 export class FoamWorkspace implements IDisposable {
--- a/packages/foam-core/test/workspace.test.ts
+++ b/packages/foam-core/test/workspace.test.ts
@@ -368,6 +368,37 @@ describe('Wikilinks', () => {
 
     expect(graph.getLinks(noteA.uri).map(l => l.target)).toEqual([noteB1.uri]);
   });
+
+  it('Handles capatalization of files and wiki links correctly', () => {
+    const noteA = createTestNote({
+      uri: '/path/to/page-a.md',
+      links: [
+        // uppercased filename, lowercased slug
+        { slug: 'page-b' },
+        // lowercased filename, camelcased wikilink
+        { slug: 'Page-C' },
+        // lowercased filename, lowercased wikilink
+        { slug: 'page-d' },
+      ],
+    });
+    const ws = createTestWorkspace()
+      .set(noteA)
+      .set(createTestNote({ uri: '/somewhere/PAGE-B.md' }))
+      .set(createTestNote({ uri: '/path/another/page-c.md' }))
+      .set(createTestNote({ uri: '/path/another/page-d.md' }));
+    const graph = FoamGraph.fromWorkspace(ws);
+
+    expect(
+      graph
+        .getLinks(noteA.uri)
+        .map(link => link.target.path)
+        .sort()
+    ).toEqual([
+      '/path/another/page-c.md',
+      '/path/another/page-d.md',
+      '/somewhere/PAGE-B.md',
+    ]);
+  });
 });
 
 describe('markdown direct links', () => {


packages/foam-core/src/model/workspace.ts
```diff
@@ -56,7 +56,7 @@ export class FoamWorkspace implements IDisposable {
   set(resource: Resource) {
     const id = uriToResourceId(resource.uri);
     const old = this.find(resource.uri);
-    const name = uriToResourceName(resource.uri);
+    const name = uriToResourceName(resource.uri).toLowerCase();
```
riccardoferretti, line 59: this is connected to my comment in the previous PR about needing to iron out the mapping of resources and wikilinks. Specifically, that means that `pathToResourceId` and `pathToResourceName`.

In that spirit, have you considered having `toLowerCase` in those functions to "normalize" the paths?

It is true that there might be case conflicts (in file path), but that seems like such an edge case I am ok not optimizing for it.

Regarding your comment about `[[Page]]` and `[[page]]`, the expected behavior would need to be properly addressed (my guess is that in the future `/path/to/Page.md` and `/another/page.md` will be treated just like the case of `/path/to/page.md` and `/another/page.md`). This needs a proper spec to be written though, for which this PR and your previous one already provide two cases.

riccardoferretti: we also talked about this here: https://github.com/foambubble/foam/pull/666#discussion_r649152167

pderaaij: > In that spirit, have you considered having toLowerCase in those functions to "normalize" the paths?

I am still wrapping my head around how everything works, so went this route to minimise impact and unintended side effects. But can have a look if things will work as it should when I put in the utility functions.

riccardoferretti: ok, I will wait to hear about the investigation before moving forward with the PR. Let me know if you have any question or need support. If the rabbit hole goes too deep, we can re-evaluate whether to simply go with this interim solution

pderaaij: Seems to don't cause any side effect, so comfortable making the change.

riccardoferretti (approved):
LGTM! will test it out then merge

PR #691, opened 2021-06-25
fix: tags in code blocks or inlines should be ignored

### LINKED ISSUE #682: Ignore tags in code blocks
Notes with CSS id's in code blocks gets the id's parsed as tags. For example `div#myId {color: red}` or

```css
div#myId {color: red}
```

would extract the tag #myId. I think it would be ok to always ignore tags in code blocks. If tags in code blocks are considered useful, maybe the parsing could at least be made an option?

RobinKing:
Not yet completed.

This pr resolves #682. The tags in code blocks and inlines are ignored. 

The new code uses [markdown-it](https://github.com/markdown-it/markdown-it) as a parser.

- [x] The first version ignores tags in the code
- [x] Add other ignore items
- [ ] Beautify the code

### DIFF
--- a/packages/foam-core/test/utils.test.ts
+++ b/packages/foam-core/test/utils.test.ts
@@ -26,7 +26,7 @@ describe('hashtag extraction', () => {
   });
   it('ignores tags that only have numbers in text', () => {
     expect(
-      extractHashtags('this #123 tag should be ignore, but not #123four')
+      extractHashtags('this #123 tag should be ignored, but not for #123four')
     ).toEqual(new Set(['123four']));
   });
   it('supports unicode letters like Chinese charaters', () => {
@@ -38,6 +38,15 @@ describe('hashtag extraction', () => {
     ).toEqual(
       new Set(['tag_with_unicode_letters_汉字', '纯中文标签', '标签1', '123四'])
     );
+  });
+  it('ignores tags in code blocks', () => {
+    expect(
+      extractHashtags(`\`\`\`
+      this #tag_in_code_blocks should be ignored
+      \`\`\`
+      and \` #tag_of_inline_code \` should be ignored, but not for #tag_outside
+      `)
+    ).toEqual(new Set(['tag_outside']));
   });
 
   it('ignores hashes in plain text urls and links', () => {
--- a/packages/foam-core/test/utils.test.ts
+++ b/packages/foam-core/test/utils.test.ts
@@ -32,8 +32,8 @@ describe('hashtag extraction', () => {
   it('supports unicode letters like Chinese charaters', () => {
     expect(
       extractHashtags(`
-        this #tag_with_unicode_letters_汉字, pure Chinese tag like #纯中文标签 and 
-        other mixed tags like #标签1 #123四 should work
+this #tag_with_unicode_letters_汉字, pure Chinese tag like #纯中文标签 and 
+other mixed tags like #标签1 #123四 should work
       `)
     ).toEqual(
       new Set(['tag_with_unicode_letters_汉字', '纯中文标签', '标签1', '123四'])
@@ -41,10 +41,11 @@ describe('hashtag extraction', () => {
   });
   it('ignores tags in code blocks', () => {
     expect(
-      extractHashtags(`\`\`\`
-      this #tag_in_code_blocks should be ignored
-      \`\`\`
-      and \` #tag_of_inline_code \` should be ignored, but not for #tag_outside
+      extractHashtags(`
+\`\`\`
+this #tag_in_code_blocks should be ignored
+\`\`\`
+and \` #tag_of_inline_code \` should be ignored, but not for #tag_outside
       `)
     ).toEqual(new Set(['tag_outside']));
   });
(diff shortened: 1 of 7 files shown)


RobinKing:
Here is a brief flow of the procedure.

---

##  For input text try to find tags

split text  by `\n` and try to find tags from each line
diagA


## process each line

do different thing according current line is in code block or not




###  Current line is in code block 

If it is already in a code block, determine current line is an end line or not


### Current line is not in code block

If it is not in a code block, determine current line is a start line or not. 


---

After execution is complete, `arrayTags` stores all tags that are not in the code block.

packages/foam-core/test/utils.test.ts
```diff
@@ -53,7 +62,7 @@ describe('hashtag extraction', () => {
 
   it('ignores hashes in links to sections', () => {
     expect(
-      extractHashtags(`
+      extractHashtags(`first line not blank
```
riccardoferretti, line 65: why does the first line need not be blank?

riccardoferretti (commented):
Hi Robin, thanks for taking a look at this issue.
Generally speaking the approach makes sense.
I was originally thinking of having the "are we in a code block" question in the Markdown Provider, the reason being that this has also other implications. For example: wikilinks in codeblocks should be ignored.
Can you take a a look at `markdown-provider.ts` and see how you could have that state being part of the parsing context?
What do you think?

packages/foam-core/test/utils.test.ts
RobinKing: The basic Markdown syntax allows to create code blocks by indenting lines by more than four spaces or one tab. So an indented block after a blank line would be considered as a block of code.

Links could be found https://spec.commonmark.org/0.30/#indented-code-block  and
https://www.markdownguide.org/basic-syntax#code-blocks.

RobinKing:
> Hi Robin, thanks for taking a look at this issue.
> Generally speaking the approach makes sense.
> I was originally thinking of having the "are we in a code block" question in the Markdown Provider, the reason being that this has also other implications. For example: wikilinks in codeblocks should be ignored.
> Can you take a a look at `markdown-provider.ts` and see how you could have that state being part of the parsing context?
> What do you think?

I had this in mind at first.
As markdown is too liberal in the formatting of code blocks, it may be necessary to use the Positon parameter to determine if it is a code block. I was too lazy for this issue. So I solved the issue with a barely working and barely reusable piece of code like this.

Thanks to your reply, I have a little more idea.
1. Iterate through the notes and save the {type, Range} of all the code blocks in the current note
2. determine if the Positon (e.g. the link position or the tag position) is in all the above Range as needed.

Since I'm new to TypeScript and given my spare time, I guess I'll be able to change the code at the weekend, hopefully not too late.

Thanks again for your reply and review!

packages/foam-core/test/utils.test.ts
riccardoferretti: This is a really good point. 
Purely to make life easier for people reading the test in the future (who will probably wonder the same thing, and won't have access to your answer), I wonder if we should instead keep the first line blank, but remove the indentation from the following lines (except in the tests cases that will check the behavior in code blocks).
what do you think?

RobinKing: I think it would be more concise to remove the indentation rather than to add the first line.

riccardoferretti: sounds good, let's go with that!

riccardoferretti:
Hey Robin, I saw the latest approach, which is definitely an improvement over the previous one.
There are a couple of considerations though that have me thinking, iiuc this would require reparsing the document (every time the function is called), I wonder what the performance implications might be?
Have you considered, within `createMarkdownParser` keeping some state of whether we are in a code block (whether inline or not), and pass it along to the plugins? 
This way all plugins would just have access to the information and be able to do whatever they believe is best for their use case (e.g. `wikilinkPlugin` will stop searching for links, `tagsPlugin` will stop searching for tags, ...). I think as design this might be more "future proof", and crucially not require additional parsing.

I understand this would require rethinking the way the `tagsPlugin` works, because back in the days I took the shortcut of parsing the whole text at once, whereas in this approach we would have to do it for single node (after checking it's not in a code block).
Calling the regex multiple times could have itself some performance implications, so truth be told I am not even positive that would be the best approach (we should do a quick perf test).

There is another consideration that you might find helpful.
I would like `Resource.tags` in the future to not just be a `Set<string>`, but a `Tag[]`, where 

```
interface Tag {
  name: string
  range: Range
}
```

similar to what we have for `Wikilink`.
This means that in the future we would need to know where the tag comes from. 
In this model, using `ParserPlugin.visit` would work well. Changing the text would cause issues.
To be clear, I don't think we should make this change in this PR, but it's good to know also what I am thinking about while looking at the solution, sorry for not sharing this direction sooner, it might have also informed some of your thinking around the approach.

I know, long comment ;) thoughts?

RobinKing:
@riccardoferretti 

I feel like I'm always tinkering to solve a problem without thinking long term. Since getting Range is slightly harder, I gave up trying to get Range and determine if it is in Range.
Thank you for your comment. I think a few things are clearer:

1. tags need more than just the name string, we also need the Range property.
2. we need a state to determine if we are in block of code.
3. the current Tag parsing method and TagsPlugin may need further optimization.
4. it is better to get all the Resources by parsing the file once to improve performance.


I think it is possible to solve this problem by adding a code type Resource, like this:
```typescript
interface Code {
  type: string;
  value: string;
  range: Range
}
```
Also, I think it would be better to use a new parsing rule to extract tag type Resource instead of using regrex for the full text.


These are a challenge for me, I have to learn a bit more about typescript and javascript, and look at some other projects  like micromark, remark, mdast....

I'll close this PR later, although I didn't solve the problem very well, I learned a lot in the process. 
And thank you very much.

riccardoferretti:
Thanks for all the work @RobinKing.
Yes I think this change is not trivial, I really appreciate you working on it.
I think you might have found a solution, but I also understand that this has turned out to be more complex than originally envisioned :)
Hopefully the work will be source of inspiration/reference for future tags related PRs, by you or others.

Thanks again!

PR #697, opened 2021-07-03
Removes escape character of an alias when setting target

### LINKED ISSUE #695: Links with aliases in Tables
- Foam version: Latest
- Platform: Windows 

I have a problem with wikilinks with aliases in tables, those need to be escaped with a single backslash to work, but in Foam links with aliases that are escaped with backslash point to "notename\\".
And if you add a link with alias without the "escape" the link do not work.
Wikilinks without aliases in tables works as they should

I also have a problem that some wikilinks with aliases point to both the Alias and the Note it self, so I get one link that goes to an  existing note and one that goes to a "ghost", the alias point to the existing note, while the ghost is the filename.

pderaaij:
This resolves #695 and additionally also provides better input for the graph rendering. When setting the target, the escape character is now explicitly removed.

### DIFF
--- a/packages/foam-vscode/src/features/document-link-provider.spec.ts
+++ b/packages/foam-vscode/src/features/document-link-provider.spec.ts
@@ -22,10 +22,7 @@ describe('Document links provider', () => {
 
   afterAll(async () => {
     await cleanWorkspace();
-    await updateFoamVsCodeConfig(
-      'edit.linkReferenceDefinitions',
-      'withoutExtensions'
-    );
+    await updateFoamVsCodeConfig('edit.linkReferenceDefinitions', undefined);
   });
 
   beforeEach(async () => {
--- a/packages/foam-vscode/src/features/document-link-provider.spec.ts
+++ b/packages/foam-vscode/src/features/document-link-provider.spec.ts
@@ -10,19 +10,16 @@ import {
 import { LinkProvider } from './document-link-provider';
 import { OPEN_COMMAND } from './utility-commands';
 import { toVsCodeUri } from '../utils/vsc-utils';
-import { updateFoamVsCodeConfig } from '../services/config';
 
 describe('Document links provider', () => {
   const parser = createMarkdownParser([]);
 
   beforeAll(async () => {
     await cleanWorkspace();
-    await updateFoamVsCodeConfig('edit.linkReferenceDefinitions', 'off');
   });
 
   afterAll(async () => {
     await cleanWorkspace();
-    await updateFoamVsCodeConfig('edit.linkReferenceDefinitions', undefined);
   });
 
   beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/src/features/document-link-provider.spec.ts
```diff
@@ -120,4 +120,23 @@ describe('Document links provider', () => {
     expect(links[0].target).toEqual(OPEN_COMMAND.asURI(noteB.uri));
     expect(links[0].range).toEqual(new vscode.Range(0, 18, 0, 33));
   });
+
+  it('should support wikilink aliases in tables using escape character', async () => {
+    const fileB = await createFile('# File B');
+    const fileA = await createFile(
+      `this is a link to [[${fileB.name}\|alias]].`
```
riccardoferretti, line 127: I think you want `[[${fileB.name}\\|alias]]` here?

pderaaij: I was under the impression it was not needed to escape the escape. But will validate it.

riccardoferretti: let me know, maybe I am confused about what you are trying to achieve here (and what the proper md and js syntax for it is)

pderaaij: Yeah, first trying to get the test stable. Will let you know.

pderaaij: You're correct. It should be a double escaped. Due to the flaky tests, I didn't properly notice it. The way to solve it for now is to spread the two tests apart. For some reason, they are interfering, but I don't know why.

pderaaij: Scratch that... A new approach which works locally consistently is to clean the workspace after each test.

pderaaij:
@riccardoferretti I can use your help here. I am stuck getting the tests stable. For whatever reason, the tests are flaky and randomly fail due to disposed editors. I don't understand what is happening here. It seems to be caused from another test, but I can't find a consistent relation. Would appreciate if you can have a look at this. Perhaps you find the right thing that's causing this.

pderaaij:
@riccardoferretti Looks like I found the issue. The problems seem to be caused by the after save wikilink reference definition generation. As it's not needed for these tests, I am disabling them via updating the configuration. On the hand, this could be the default behaviour for all tests.

riccardoferretti:
You are a hero for finding this, I was not looking anywhere near that, I thought the issue was in the test setup/teardown. 🎉🎉🎉

packages/foam-vscode/src/features/document-link-provider.spec.ts
```diff
@@ -10,16 +10,22 @@ import {
...
   beforeAll(async () => {
     await cleanWorkspace();
+    await updateFoamVsCodeConfig('edit.linkReferenceDefinitions', 'off');
   });
 
   afterAll(async () => {
     await cleanWorkspace();
+    await updateFoamVsCodeConfig(
+      'edit.linkReferenceDefinitions',
+      'withoutExtensions'
```
riccardoferretti, line 27: if what you are trying to achieve is removing the config set above I believe the best approach is using `undefined` here, see docs for `vscode.workspace.getConfiguration().update()`

riccardoferretti (approved):
LGTM!

Issue #692, opened 2021-06-27
[Templates v2] Add the ability to use VS Code snippet variables in template metadata


movermeyer:
#655 added metadata to templates, including the `filepath` attribute.

As part of #534, I had dreamed of being able to use any [VS Code snippet variable](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables) in the `filepath`.

```yaml
---
foam_template:
  filepath: `journal/${CURRENT_YEAR}-${CURRENT_MONTH}-${CURRENT_DATE}.md`
---
```

This would allow `Foam: Open Daily Note` to make use of templates easily.
However, this is not currently possible. VS Code does not currently offer an interface for resolving the snippet variables without opening the snippet in an editor for the user.

---

Figure out a workaround, or get VS Code to expose an interface that would suit our needs.

pderaaij:
Would it be an option to open an editor, fill it with the template and resolve the snippets and then save the file? If we parsed the file as a note, you would have access to the frontmatter and the resolved filepath. But it's a path with many things that have to fall into place.

riccardoferretti:
I have added a request to improve the support for snippets outside of editors: https://github.com/microsoft/vscode/issues/155868

Let's upvote it to get it in their backlog

Issue #566, opened 2021-04-02
[Templates v2] Following missing Wikilinks should use a template


movermeyer:
As described in #534, clicking on a Wikilink to a non-existent note should use the `Foam: Create New Note` implementation, using the contents of the Wikilink as the value of `${FOAM_TITLE}`.

Issue #564, opened 2021-04-02
[Templates v2] Add a template variable that replaces `dateFormat`


movermeyer:
As described in #534, `Foam: Open Daily Note` uses the `foam.openDailyNote.filenameFormat` setting and passes it to `dateFormat`. This allows users the flexibility to put the current timestamp into the daily notes filename using any format available via `dateFormat`. This is much more flexible than what is available in VSCode Snippet variables.

Add a new syntax / mechanism to maintain the flexibility of `dateFormat` within a template variable.

movermeyer:
Instead of doing anything complicated in the snippet variable syntax (which I want to improve so we exactly match the VSCode/TextMate syntax), what if we introduce a new settings variable (perhaps `foam.dateFormatVariablePattern`) that a `${FOAM_DATE_FORMAT}` snippet variable uses to pass into `dateFormat` to get its value?

```json
"foam.dateFormatVariablePattern": "yyyy-mm-dd"
```

riccardoferretti:
mmm.. let's see if I understand:
- add a new setting: `foam.dateFormatVariablePattern`. This uses the `dateFormat` mechanism to populate the variable
- make the setting available as a variable in the foam template (kinda like `${FOAM_TITLE}`

IIUC I wonder if this solution would be a bit complicated to handle for the user.

Can you walk me through some use cases you are solving for?

movermeyer:
>    add a new setting: foam.dateFormatVariablePattern. This uses the dateFormat mechanism to populate the variable
>    make the setting available as a variable in the foam template (kinda like ${FOAM_TITLE}

Yeah, that's correct.

---

> I wonder if this solution would be a bit complicated to handle for the user. Can you walk me through some use cases you are solving for?

Users can already use the [`CURRENT_YEAR`/`CURRENT_MONTH`, etc.](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables) offered natively by VS Code to insert date/time values in templates.

IMO, this is probably enough flexibility to represent date times in notes:

* `2021-05-01` (`$CURRENT_YEAR-$CURRENT_MONTH-$CURRENT_DATE`)
* `05/01/2021` (`$CURRENT_MONTH/$CURRENT_DATE/$CURRENT_YEAR`)
* `01.05.2021` (`$CURRENT_DATE.$CURRENT_MONTH.$CURRENT_YEAR`)
[...]

PR #700, opened 2021-07-05
Make daily notes use templates

### LINKED ISSUE #565: [Templates v2] Make `Foam: Open Daily Note` use a template

movermeyer:

Fixes #565: Allowing daily notes to use templates.

1. I added a new template creation function, `createNoteFromDailyNoteTemplate`:
   * It differs from `createNoteFromDefaultTemplate` in a few ways:
       * It uses `daily-note.md` instead of `new-note.md`
       * It only resolves `FOAM_TITLE` if the template requires it
       * It accepts parameters for the default template text and the desired output location
2. I then changed `createDailyNoteIfNotExists` to call this function.
3. I then refactored `createNoteFromDefaultTemplate` to be able to handle all the differences.
    * This refactor will be re-used as part of the changes needed for https://github.com/foambubble/foam/issues/566

#### Limitation

We make use of the `filepath` template metadata attribute, if it is present in the template. 
However, this is currently pretty useless, since we don't yet have a way to specify the date in the `filepath`.

* If we end up [adding a Foam-specific variable](https://github.com/foambubble/foam/issues/564), then we could use that new variable in `filepath`
* Alternatively, if we solve #692, then we could use the [`CURRENT_YEAR`/`CURRENT_MONTH`/`CURRENT_DATE` VS Code snippet variables](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables).

For right now, as long as they don't use `filepath`, we continue to behave as we currently do: determining the filepath from the `foam.openDailyNote.*` settings.

**Claim:** This limitation should not prevent users from enjoying the other benefits of templates in the mean time.

...

...

### Testing

To test this change:

1. Create a new `.foam/templates/daily-note.md` template file:

```yaml
# $CURRENT_YEAR-$CURRENT_MONTH-$CURRENT_DATE

## What I plan on doing today
 
* [ ] Contribute to Foam! 🎉 
* [ ] 20 minute cardio workout 💨 

## What I did today

* 

```

2. Run `Foam: Open Daily Note`

The new note should be created at the filepath defined by `foam.openDailyNote.*` settings, using the contents of the `daily-note.md` template.

### Checklist

### DIFF
--- a/docs/features/daily-notes.md
+++ b/docs/features/daily-notes.md
@@ -25,9 +25,9 @@ The above configuration would create a file `journal/note-2020-07-25.mdx`, with 
 
 ## Daily Note Templates
 
-In the future, Foam may provide a functionality for specifying a template for new Daily Notes and other types of documents.
+Daily notes can also make use of [templates](note-templates.md), by defining a special `.foam/templates/daily-note.md` template.
 
-In the meantime, you can use [VS Code Snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets) for defining your own Daily Note template.
+See [Note Templates](note-templates.md) for details of the features available in templates.
 
 ## Roam-style Automatic Daily Notes
 
--- a/docs/features/note-templates.md
+++ b/docs/features/note-templates.md
@@ -24,10 +24,16 @@ To create a note from a template:
 
 _Theme: Ayu Light_
 
-## Default template
+## Special templates
+### Default template
 
 The `.foam/templates/new-note.md` template is special in that it is the template that will be used by the `Foam: Create New Note` command.
 Customize this template to contain content that you want included every time you create a note.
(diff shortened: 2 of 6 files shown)


riccardoferretti (approved):
This looks really good. Excited about using it too!

olaven:
I know I am late, but I just wanted to thank you for implementing this, @movermeyer😍  It really improved my workflow 🙌

PR #708, opened 2021-07-10
Introduce autocompletion for tags

### LINKED ISSUE #669: Auto-complete/intellisense for tags
Some ideas to support easier and more controllable tag management:

- auto-complete/intellisense for tags. if we consider that they are to be in some way controlled or loosely controlled it would be helpful.
- an ability to dump all tags in the alphabetical order into an auto-updated md file for easy viewing, exporting, visualization, etc...

pderaaij:
Resolves #669, allowing autocompletion for tags. 

ToDo:

- [x] verify behaviour 
- [x] optimise code
- [x] see if all edge cases are being handled in tests
- [x] Performance tests

### DIFF
--- a/packages/foam-core/src/model/tags.ts
+++ b/packages/foam-core/src/model/tags.ts
@@ -1,0 +1,159 @@
+import { FoamWorkspace } from './workspace';
+import { URI } from './uri';
+import { IDisposable } from '../index';
+import { Resource } from './note';
+
+export type TagMetadata = { uri: URI };
+
+export class FoamTags implements IDisposable {
+  public readonly tags: { [key: string]: TagMetadata[] } = {};
+
+  constructor(private readonly workspace: FoamWorkspace) {}
+
+  /**
+   * List of disposables to destroy with the tags
+   */
+  private disposables: IDisposable[] = [];
+
+  /**
+   * Computes all tags in the workspace and keep them up-to-date
+   *
+   * @param workspace the target workspace
+   * @param keepMonitoring whether to recompute the links when the workspace changes
+   * @returns the FoamTags
+   */
+  public static fromWorkspace(
+    workspace: FoamWorkspace,
+    keepMonitoring: boolean = false
+  ): FoamTags {
+    let tags = new FoamTags(workspace);
+
+    Object.values(workspace.list()).forEach(resource =>
+      tags.resolveResource(resource)
+    );
+
+    if (keepMonitoring) {
+      tags.disposables.push(
+        workspace.onDidAdd(resource => {
+          tags.resolveResource(resource);
+        }),
+        workspace.onDidUpdate(change => {
+          tags.updateTagsForUpdatedResource(change.old, change.new);
+        }),
+        workspace.onDidDelete(resource => {
+          tags.updateTagsForRemovedResource(resource);
+        })
+      );
+    }
+    return tags;
+  }
+
+  updateTagsForUpdatedResource(oldVersion: Resource, newVersion: Resource) {
+    if (oldVersion.tags.size === 0 && newVersion.tags.size === 0) {
+      return;
+    }
+
+    this.resolveUpdatedTags(newVersion, oldVersion);
+    this.resolveMovedResource(newVersion, oldVersion);
+  }
+
+  updateTagsForRemovedResource(resource: Resource): void {
+    if (resource.tags.size === 0) {
+      return;
+    }
+
+    resource.tags.forEach(tag => {
+      this.tags[tag] = this.tags[tag].filter(
+        t => !URI.isEqual(t.uri, resource.uri)
+      );
+
+      if (this.tags[tag].length === 0) {
+        delete this.tags[tag];
+      }
+    });
+  }
+
+  resolveResource(resource: Resource): void {
+    resource.tags.forEach(tag => {
+      this.tags[tag] = this.tags[tag] ?? [];
+      this.tags[tag].push({ uri: resource.uri });
+    });
+  }
+
+  dispose(): void {
+    this.disposables.forEach(d => d.dispose());
+    this.disposables = [];
+  }
+
+  /**
+   * If the file has been moved, update all relevant metadata objects of the tag
+   * @param newVersion
+   * @param oldVersion
+   */
+  private resolveMovedResource(newVersion: Resource, oldVersion: Resource) {
+    if (!URI.isEqual(oldVersion.uri, newVersion.uri)) {
+      newVersion.tags.forEach(tag => {
+        this.tags[tag].forEach((meta, idx) => {
+          if (meta.uri === oldVersion.uri) {
+            this.tags[tag][idx].uri = newVersion.uri;
+          }
+        });
+      });
+    }
+  }
+
+  /**
+   * Look for differences in the tags of the resources and update accordingly
+   *
+   * @param newVersion
+   * @param oldVersion
+   */
+  private resolveUpdatedTags(newVersion: Resource, oldVersion: Resource) {
+    const newTags = Array.from(newVersion.tags).filter(
+      t => !oldVersion.tags.has(t)
+    );
+    const removedTags = Array.from(oldVersion.tags).filter(
+      t => !newVersion.tags.has(t)
+    );
+
+    newTags.forEach(newTag => {
+      if (!this.tags[newTag]) {
+        this.tags[newTag] = [];
+      }
+
+      this.tags[newTag].push({ uri: newVersion.uri });
+    });
+
+    removedTags.forEach(removedTag => {
+      // should not happen
+      if (!this.tags[removedTag]) {
+        return;
+      }
+
+      this.tags[removedTag] = this.tags[removedTag].filter(
+        meta => !URI.isEqual(meta.uri, newVersion.uri)
+      );
+
+      if (this.tags[removedTag].length === 0) {
+        delete this.tags[removedTag];
+      }
+    });
+
+    // No change in tags detected, but might be a newly added tag to the file
+    if (
+      removedTags.length === 0 &&
+      newTags.length === 0 &&
+      URI.isEqual(oldVersion.uri, newVersion.uri)
+    ) {
+      newVersion.tags.forEach(tag => {
+        const noteRegistered = this.tags[tag].find(meta =>
+          URI.isEqual(meta.uri, newVersion.uri)
+        );
+
+        if (!noteRegistered) {
+          this.tags[tag].push({ uri: newVersion.uri });
+        }
+      });
+    }
+  }
+}
--- a/packages/foam-core/src/model/tags.ts
+++ b/packages/foam-core/src/model/tags.ts
@@ -6,7 +6,7 @@ import { Resource } from './note';
 export type TagMetadata = { uri: URI };
 
 export class FoamTags implements IDisposable {
-  public readonly tags: { [key: string]: TagMetadata[] } = {};
+  public readonly tags: Map<string, TagMetadata[]> = new Map();
 
   constructor(private readonly workspace: FoamWorkspace) {}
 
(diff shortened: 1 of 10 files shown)


pderaaij:
For this feature, I am thinking to extend the FoamGraph object with a property that holds all unique tags found in the workspace. This would reduce a lot of processing for the completion functionality, but also the tag tree view explorer. I expect this to ensure performance, but also make it easier to create features based on the tags (e.g. a graph based on tags).

@riccardoferretti is this in line with plans or any other thoughts? The current setup does not feel comfortable in means of performance.

riccardoferretti:
Yes, I think on this implementation we might have performance issues because the list is computed every time a completion is triggered.

As a first optimization we could create a local cache of tags, and monitor the workspace for changes to update it. Probably risky (cache management is tricky and better if centralized), and in any case it would duplicate what we already do in the tag explorer.

As you mentioned another approach is to add support for tags in `FoamWorkspace` or `FoamGraph`, which would consolidate the computation/management to one place. The drawback is that it increases the computation of the cache, but practically this is taken from the tag explorer's so it should be fine (although I would do a performance test).

What's your take about using `FoamWorkspace` or `FoamGraph`?

pderaaij:
`FoamGraph` makes the most sense to me. Already holds a graph of things like links and backlinks. Tag fit in here quite well. Also, the right update triggers are present here to keep the state in sync. So, that's where I would make a change.

riccardoferretti:
I see what you mean, I am a bit hesitant to put it in the graph because it already manages a fairly complex object and related state. 

I can see a couple of other options too:

1. use `FoamWorkspace` 

Tags exist in resources, so they can be handled directly in the `FoamWorkspace` as resources are loaded.

2. create a `FoamTags` object 

This would live in the main `Foam` object next to the `workspace` and `graph` properties, which would follow the same pattern as `FoamGraph` (in terms of being seeded from the workspace, and looking for updates).
This object could also have a custom API to handle tag specific behavior (e.g. parent/child tags)

What do you think?

CC @movermeyer @jmg-duarte in case they have an opinion

pderaaij:
I was balancing a trade-off between parsing the Markdown file multiples times, as multiple listeners respond to a change, and the complexity of the code. But, it might not even be a problem. Creating a `FoamTags` object would be a perfect option as well.

riccardoferretti:
I also like the `FoamTags` approach, given the tags are already in the `Resource` no additional parsing is required, and the pattern used by `FoamGraph` would work well there too.

Given the performance considerations of the current approach, I'd say let's keep all of this in one PR (that is, let's implement `FoamTags` in this same PR), which would then have:

- [ ] create `FoamTags`
- [ ] refactor tag explorer to use `FoamTags`
- [ ] change auto completion to use `FoamTags`

In a way the two use cases and the tag explorer code will also help adjust the design, which is good.

pderaaij:
@riccardoferretti Could you run the Windows tests with a clean cache for me? On my personal fork, the tests are running perfectly. Don't know what's causing the issue. 

If that all runs fine, please have a look at the PR and give your thoughts on the implementation. Thanks!

packages/foam-core/src/model/tags.ts
```diff
@@ -0,0 +1,183 @@
+import { FoamWorkspace, IDisposable, URI } from '../index';
+import { Resource } from './note';
+
+export type TagMetadata = { title: string; uri: URI };
```
riccardoferretti, line 4: where is the title needed?

packages/foam-core/src/model/tags.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      if (this.tags[tag].length === 0) {
+        delete this.tags[tag];
+      }
+    });
+  }
+
+  resolveResource(resource: Resource): void {
+    resource.tags.forEach(tag => {
+      if (!this.tags[tag]) {
```
riccardoferretti, line 78: minor: the pattern we use elsewhere is a bit different, here it is from `workspace.ts`:

```js
    this.resourcesByName[name] = this.resourcesByName[name] ?? [];
    this.resourcesByName[name].push(id);
```

packages/foam-vscode/src/features/tag-completion.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+  cleanWorkspace,
+  closeEditors,
+  createFile,
+  createTestNote,
+  showInEditor,
+} from '../test/test-utils';
+import { TagCompletionProvider } from './tag-completion';
+
+describe('Taf Completion', () => {
```
riccardoferretti, line 12:
```suggestion
describe('Tag Completion', () => {
```

packages/foam-core/src/model/tags.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  constructor(private readonly workspace: FoamWorkspace) {}
+
+  /**
+   * List of disposables to destroy with the tags
+   */
+  private disposables: IDisposable[] = [];
+
+  /**
+   * Computes all the links in the workspace, connecting notes and
```
riccardoferretti, line 17: let's update the comment

riccardoferretti (commented):
The approach looks pretty good to me!
I have left a few comments about the `FoamTags` data structure, let me know your thoughts

packages/foam-core/src/model/tags.ts
pderaaij: Ah, I've kept that one in for the tag view explorer. But we are resolving the entire resource by URI at that point. We could stop the resolving, but then we lose the hover functionality, or remove the title from the tag object.

pderaaij: For now, I'll remove it. Saves extra work in keeping things up to date as well.

pderaaij:
@riccardoferretti Applied your comments, thanks for the review! Could you please rerun the Windows runner for me with a clear cache? Don't know why it is borking, but personal fork gives an okay.

packages/foam-core/src/model/tags.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+    return tags;
+  }
+
+  updateTagsForUpdatedResource(oldVersion: Resource, newVersion: Resource) {
+    if (oldVersion.tags.size === 0 && newVersion.tags.size === 0) {
+      return;
+    }
+
+    this.resolveUpdatedTags(newVersion, oldVersion);
```
riccardoferretti, line 56: I imagine you followed this pattern because it was the one used in `FoamGraph`.
The reason why I originally wrote it this way there was to allow for more granular updates in the graph listeners (which was never actually implemented), at the cost of the added complexity of this approach.

I wonder if for tags it would be simpler to just remove the old resource from the map, and add the new one?
This should make the code much easier to follow, wdyt?

packages/foam-core/src/model/tags.ts
```diff
@@ -0,0 +1,159 @@
...
+      if (this.tags[tag].length === 0) {
+        delete this.tags[tag];
+      }
+    });
+  }
+
+  resolveResource(resource: Resource): void {
+    resource.tags.forEach(tag => {
+      this.tags[tag] = this.tags[tag] ?? [];
+      this.tags[tag].push({ uri: resource.uri });
```
riccardoferretti, line 79: what happens if a resource defines a tag more than once?

e.g.

```
this is a #good day. And I like #good things.
```

packages/foam-vscode/src/features/tags-tree-view/index.ts
```diff
@@ -136,7 +124,6 @@ export class Tag extends vscode.TreeItem {
     this.description = `${this.notes.length} reference${
       this.notes.length !== 1 ? 's' : ''
     }`;
-    this.tooltip = getContainsTooltip(this.notes.map(n => n.title));
```
riccardoferretti, line 139: this can actually be achieved with the new data structure in I think an elegant way using `resolveTreeItem` above.

Let me know if it makes sense

riccardoferretti:
This issue with the tests is a bit baffling...

I have a feeling it's not picking up your `foam-core` from the code, but something from the cache.
Can you do a quick experiment and update `foam-vscode/yarn.lock:L2509` with the following:

```
"foam-core@^0.14.1":
  "integrity" "sha512-bzUpaQhGCDCMGX7T9EbeIGmdfLn5CXlXDfguTs/35N2pl1XPp0r/zk2JAw1QsMr+eyCyTTToOy3ijt0gfUEhxQ=="
  "resolved" "https://registry.npmjs.org/foam-core/-/foam-core-0.14.1.tgz"
  "version" "0.14.1"
```

packages/foam-core/src/model/tags.ts
pderaaij: As far as I can see, the tags on the resource are unique. Did a little test run, and then I also end up with a single URI for a tag. So, feels like this is good as it is?

packages/foam-core/src/model/tags.ts
pderaaij: I am not confident if I am reading this right, so let's see if it matches. If I read it correctly, you'd propose the following 

```
loop through all tags of the old resource
for each tag, remove each instance of old resoure uri in the metadata

loop through all tags of the new resource
for each tag, add each instance of new resoure uri in the metadata
```

That would reduce a bit of the complexity of computing the difference of tags between old and new resource. Other than that, it's relatively the same. I only don't know what the impact on performance could be. But before I make a change, have I read it correctly or do you mean something else?

packages/foam-vscode/src/features/tags-tree-view/index.ts
pderaaij: Yes, the only thing I am doubting here is the performance of it. It basically means it will start to resolve the entire tree for this tooltip. Given the content of the tooltip, I don't know if that's worth the effort.

packages/foam-core/src/model/tags.ts
riccardoferretti: yes, basically we would have a function to add a resource (a)
```
loop through all tags of the new resource
for each tag, add each instance of new resoure uri in the metadata
```

and a function to remove a resource (b)
```
loop through all tags of the old resource
for each tag, remove each instance of old resoure uri in the metadata
```

updating a resource is basically b+a

```
loop through all tags of the old resource
for each tag, remove each instance of old resoure uri in the metadata

loop through all tags of the new resource
for each tag, add each instance of new resoure uri in the metadata
```

I think we are saying the same thing, right?

packages/foam-core/src/model/tags.ts
riccardoferretti: oh yes, that's a good point, we already make tags unique in the `tags` field.

FYI in the future tags won't be just `string`, but 
```
Tag {
  name: string
  range: Range
}
```

Which means that it would be possible to have more than one with the same name within a given resource.
But probably we can enforce the uniqueness when we'll need to do the refactoring here at that point.

So I think we are good here!

packages/foam-vscode/src/features/tags-tree-view/index.ts
riccardoferretti: > It basically means it will start to resolve the entire tree for this tooltip

not sure what you mean by "entire tree". The `resolveTreeItem` is meant to be lazily evaluated for a single tree item when hovered. See https://code.visualstudio.com/api/references/vscode-api#TreeDataProvider%3CT%3E

I agree the tooltip is not essential, so I am happy to go with your judgement as to whether to re-implement it lazily or not 👍

packages/foam-core/src/model/tags.ts
```diff
@@ -0,0 +1,159 @@
+import { FoamWorkspace } from './workspace';
+import { URI } from './uri';
+import { IDisposable } from '../index';
+import { Resource } from './note';
+
+export type TagMetadata = { uri: URI };
+
+export class FoamTags implements IDisposable {
+  public readonly tags: { [key: string]: TagMetadata[] } = {};
```
riccardoferretti, line 9: a very good point that comes from https://github.com/foambubble/foam/pull/729:
we should use `Map` instead of plain objects.

Otherwise a `#constructor` tag would break the code - can you check that is the case?

packages/foam-core/src/model/tags.ts
pderaaij: Altered! Simplified the code a bit ;-)

packages/foam-core/src/model/tags.ts
pderaaij: I've changed it to a `Map`.

riccardoferretti (approved):
LGTM!

jz314:
figured this out- appears to fall under general vscode intellisense; [trigger](https://code.visualstudio.com/docs/editor/intellisense#_intellisense-features) with `cntrl + space` after typing the `#` that preceeds tags.

PR #537, opened 2021-03-18
Refactored URI for Foam API v1

riccardoferretti:
As we prepare the release of Foam API v1, we want to create a sustainable API model.
Currently the `URI` is a heavy object that matches what is used by VSCode. 
Although this is convenient in a few places, it comes with various side effects and maintenance problems (e.g. see  #507).

This PR improves the `URI` object and utility methods in the following ways:
- not dependent on the VS Code URI class (now `URI` is just a simple data structure with the standard URI fields)
- consolidated `common/uri.ts` and `utils/uri.ts` into `model/uri.ts`
- consolidated the relationship amongst some URI related utility methods
- added simple utility methods to convert between Foam URI and VS Code URI (following the same pattern as `Range` and `Position` types)
- file system path is now computed via function (instead of being the `fsPath` field on the object), which improves the semantics of URI object

In terms of using VS Code URI vs Foam URI: always use Foam URI unless you need to use the VS Code one (same pattern as when using `path` vs `fsPath` in the past).

This is a pretty big refactoring 😅  but quite core, please review with attention 😉  and leave your thoughts

PR #709, opened 2021-07-12
Migrate create-from-template to use Foam's `URI` instead of VSCode's `Uri`

movermeyer:

Foam uses its own `URI` object, which is similar to VSCode's `Uri` object.
See https://github.com/foambubble/foam/pull/537 for the reasoning.

This PR changes `create-from-template` to use Foam's `URI` instead of VSCode's `Uri`.

1. Basic rename `Uri` -> `URI`
2. Corrected all the compiler failures
    * `.fsPath` -> `URI.toFsPath`
    * Use `toVsCodeUri` when calling VSCode APIs that care. 

...

No impact on functionality. It will make my changes in https://github.com/foambubble/foam/pull/700 a bit cleaner.

### Testing

...

### Checklist

### DIFF
--- a/packages/foam-core/src/model/uri.ts
+++ b/packages/foam-core/src/model/uri.ts
@@ -1,4 +1,7 @@
-// Some code in this file coming from https://github.com/microsoft/vscode/
+// `URI` is mostly compatible with VSCode's `Uri`.
+// Having a Foam-specific URI object allows for easier maintenance of the API.
+// See https://github.com/foambubble/foam/pull/537 for more context.
+// Some code in this file comes from https://github.com/microsoft/vscode/main/src/vs/base/common/uri.ts
 // See LICENSE for details
 
 import * as paths from 'path';
--- a/packages/foam-vscode/src/features/create-from-template.spec.ts
+++ b/packages/foam-vscode/src/features/create-from-template.spec.ts
@@ -1,5 +1,7 @@
+import { URI } from 'foam-core';
 import path from 'path';
-import { commands, Uri, window, workspace } from 'vscode';
+import { toVsCodeUri } from '../utils/vsc-utils';
+import { commands, window, workspace } from 'vscode';
 describe('createFromTemplate', () => {
   describe('create-note-from-template', () => {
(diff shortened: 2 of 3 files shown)


riccardoferretti:
This is something I thought about when deciding whether to use `URI` vs `Uri` in `foam-vscode`.
At the time I decided to go with `URI` because:
- I felt it made the code more "compatible" with Foam and its patterns
- it was going to highlight any issue with `URI` that might otherwise have gone unnoticed (in fact there was a Windows issue that we later had to fix)

That said, I was nonetheless feeling a bit torn about it as I can also see some advantages in using `Uri`.

A couple of questions:
1. Can you walk through the advantages of doing that?
2. Would at this point make sense to do the refactoring across the `foam-vscode` source? Or at least define the pattern of using `Uri` wherever possible?

Thoughts?

movermeyer:
@riccardoferretti 

>     1. Can you walk through the advantages of doing that?

In #700, I'm making `dated-notes` use the note creation code in `create-from-template`.
But `dated-notes` uses `URI`, while `create-from-template` uses `Uri`
So I could avoid the conversion between them within the Foam-related code, and push any conversion to `URI` to the edge where we interact with VSCode interfaces that require `URI`s.

IMO, within Foam code we should only use one type, and only do conversion when necessary.

>     2. Would at this point make sense to do the refactoring across the `foam-vscode` source? Or at least define the pattern of using `Uri` wherever possible?

I'm happy to have a rule saying "using `Uri` wherever possible". Frankly, I thought that was already the case.

riccardoferretti:
> Within Foam, use URI whenever possible, converting to VSCode's Uri only when necessary

Per our conversation, sounds good.

Let me know if this PR is still "valid" and needs to be updated, or whether you will just go for #700

movermeyer:
> Let me know if this PR is still "valid"

@riccardoferretti For the sake of keeping each PR small and independent, let's merge this one. Once it's in, I'll rebase #700 onto it.

riccardoferretti (approved):
LGTM!

Issue #648, opened 2021-05-21
.md links are opening a webpage instead of opening a note in the Markdown Preview 


YAS01-M:
- Foam version: 0.13.3
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: No

**Summary**

Hi 
Since the update to Foam v0.13.3 , when i click on any of .md links , they open a vscode webview test instead of a link of the markdown note . This issue is just in markdown preview , .md links still working  with Ctrl+Click in source page. 

screenshot of the issue when i click on [[todo]] note
![issue](https://user-images.githubusercontent.com/84593166/119190907-da386100-ba75-11eb-8965-027bbe685156.JPG)

riccardoferretti:
I am very surprised by this, and the regression on my system is happening for all versions of Foam since the feature was introduced, which makes me think something has changed in the way vscode is handling those links.
Seems related to this even though it's been fixed: https://github.com/microsoft/vscode/issues/115812

YAS01-M:
Yes , this happened also after the update of vs code studio , but before threre is no issue . I removed everything , started now with vs code insiders but nothing changed . Don't know if this issue happened with all the people who are using Foam or just many of us .
Another screenshot before clicking on some of the markdown notes  :
![issue](https://user-images.githubusercontent.com/84593166/119244418-75116800-bb68-11eb-95f8-25e5e496cb4f.JPG)

Saskia-R:
Same phenomenon on linux (x64), visual studio code version 1.56.2, Foam 0.13.3. The theory about vscode handling changing sounds logical.

riccardoferretti:
in Discord it's been reported that the problem was introduced in VS Code between version 1.55.2 and 1.56.2.
Downgrading to version 1.55.2 will solve the issue.
We still haven't figured out (or, really, deeply looked into) what is causing the issue in 1.56+ - any help there would be greatly appreciated!

dewinterjack:
I have submitted a [new issue ](https://github.com/microsoft/vscode/issues/125595)

caseychu:
[This issue](https://github.com/microsoft/vscode/issues/124678) seems related. It seems like the API for constructing links has changed to [`webview.asWebviewUri`](https://github.com/microsoft/vscode/blob/307cb32f306a527a30a5b1756e8bf548c141192c/src/vs/vscode.d.ts#L6703), and here would be the relevant place to change in Foam:
[...]

PR #710, opened 2021-07-12
Allows wikilinks to be opened in the editor again

pderaaij:
Solves #648 by adding a data-href attribute to the link. The combination of this attribute and an absolute link will do the trick for correct behaviour. Found the solution in this issue: https://github.com/microsoft/vscode/issues/126402


riccardoferretti (approved):
Amazing find!

PR #712, opened 2021-07-12
Use templates for wikilinks

### LINKED ISSUE #566: [Templates v2] Following missing Wikilinks should use a template

movermeyer:

Fixes #566: Following missing wikilinks should use a template.

Making use of the refactor done in #700...

1. I added a new template creation function, `createNoteForPlaceholderWikilink`:
   * It differs from `createNoteFromDefaultTemplate` in that:
       * It accepts parameters for the default template text and the desired output location
2. I then changed `OPEN_COMMAND` to call this function.

Wikilinks are a new area of the codebase for me. Am I missing something? Are there other code paths?

When following a placeholder wikilink, the new note will be created using the contents of `new-note.md` and using the link as the value of `$FOAM_TITLE`. 

### Testing

To test this change:

1. Create a `.foam/templates/new-note.md` template file:

```yaml
# $FOAM_TITLE

This is a new note about ${FOAM_TITLE}! 🎉
```

2. In an already existing Markdown (`.md`) file, add a placeholder wikilink:

```markdown
[[Hippos]]
```

3. `Cmd+click` the placeholder wikilink

The new note should contain the contents of the template, with the wikilink used as the value of `$FOAM_TITLE`:

```markdown
# Hippos

This is a new note about Hippos! 🎉
``` 

### Checklist


movermeyer:
Aside: [`wikilink`](https://en.wikipedia.org/wiki/Help:Link#Wikilinks) is what they are called. Not `wiki-link` or `wiki link`.
Should do a pass over the documentation and ensure we're using the right name.

packages/foam-vscode/src/features/utility-commands.ts
```diff
@@ -16,16 +16,21 @@ export const OPEN_COMMAND = {
...
-
-        if (isSome(newNote)) {
-          const title = uri.path.split('/').slice(-1);
-          const snippet = new vscode.SnippetString(
-            '# ${1:' + title + '}\n\n$0'
-          );
-          await focusNote(newNote, true);
-          await vscode.window.activeTextEditor.insertSnippet(snippet);
-        }
+        const title = uri.path.split('/').slice(-1)[0];
```
riccardoferretti, line 19: can you walk me through this change?

have you embedded the `createNoteFromPlaceholder` function and using the template pattern at the same time?

riccardoferretti (approved):
Somehow I missed this becoming ready for review!
It looks really good, and it's great to see templates being used here as well, I really like the pattern.

I have left a minor comment, but this looks good to me

packages/foam-vscode/src/features/utility-commands.ts
movermeyer: > can you walk me through this change?
@riccardoferretti 

* `OPEN_COMMAND` is the command used when other parts of Foam want to follow a placeholder Wikilink:
  * It calls `createNoteFromPlaceholder` to find the location for the new note, and create an empty file there.
  * If it could not determine where to create the file, then it returns, doing nothing.
  * `title` is the contents of the wikilink, which it writes into the new note as a snippet placeholder

I have changed `OPEN_COMMAND` so that it uses the new `createNoteForPlaceholderWikilink` to maintain the same behaviour:
  * `basedir` uses the same logic from `createNoteFromPlaceholder` to determine where to open the new note.
    * If the default template specifies a `filepath` metadata attribute, then that will be used instead
    * If it cannot be determined, `OPEN_COMMAND` still returns without doing anything.
      * In practice, it doesn't seem possible for this to happen. If there is no active note, what placeholder wikilink are you clicking on??
      * Besides, Foam doesn't (yet) work at all if you don't have a workspace open.
        * I really want to fix all the `vscode.workspace.workspaceFolders[0]` instances, but that is not something for this PR to fix.
  * `title` is the contents of the wikilink, which is used as the value for `$FOAM_TITLE`, as per #566
  * `wikilinkDefaultTemplateText` matches the current behaviour's file contents, but of course will only be used if the `new-note.md` template doesn't exist.

> have you embedded the `createNoteFromPlaceholder` function

The default filepath location logic is now embedded in `OPEN_COMMAND`. 
The file creation logic is now done by the `create-from-template` code.
Now that `createNoteFromPlaceholder`'s functionality has been subsumed, it has been removed.

riccardoferretti: Thanks for this.
I agree with your sentiment about `workspaceFolders`. We probably need to consolidate the path resolving logic for notes/placeholders/... in one place (we touched on this before). Something that would consistently apply the proper basedir prioritization, resolve relative paths, append the extension if necessary, ... once that's in place we could add some configuration to it, e.g. note extension, slugification/casification/...

But not for this PR ;)

riccardoferretti (approved):

Issue #715, opened 2021-07-12
Wikilink notes created on Cmd+<Hover>?


movermeyer:
- Foam version: v0.13.8
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**

* When I Cmd+<Hover> over a placeholder Wikilink, it produces a new file as a sibling of the file that contains the wikilink.
    * The filename is slugified (lowercase + escaped)
* When I Cmd+Click a placeholder Wikilink, it produces a new file in the root of the workspace
    * The filename is **not** slugified, as expected
    * Depending on timing, the Cmd+<Hover> may also occur.

I do not expect any notes to be created until I Cmd+Click, at which point I expect the note name to not be slugified.

**Steps to reproduce**

![weirdness](https://user-images.githubusercontent.com/1459385/125353156-2f834700-e330-11eb-95c4-c482627e0078.gif)

#### Scenario 1

1. Create a new note `<WORKSPACE>/weirdness/link_to_weirdness.md`

```markdown
[[Weirdness]]
```

2. Cmd+<Hover> over the `[[Weirdness]]` wikilink with the mouse

`<WORKSPACE>/weirdness/weirdness.md` is created (with the contents `# Weirdness`) and link reference definitions are generated and added to `<WORKSPACE>/weirdness/link_to_weirdness.md`

#### Scenario 2

1. Create a new note `<WORKSPACE>/weirdness/link_to_weirdness.md`

```markdown
[[Weirdness]]
```

2. Cmd+Click over the `[[Weirdness]]` wikilink with the mouse

`<WORKSPACE>/Weirdness.md` is created (with the contents [`# $\{1:Weirdness}\n\n$0`](https://github.com/foambubble/foam/blob/1ab9cc5f4a3cc89fa6fb9751fc5f9d07e48ccb8b/packages/foam-vscode/src/features/utility-commands.ts#L24))

Depending on timing, the behaviour of Scenario 1 may also occur.

riccardoferretti:
do you also have Markdown Notes installed?

movermeyer:
I did. 🤦 

Uninstalling `Markdown Notes` fixes the problem.

Issue #543, opened 2021-03-22
Follow link via `Go to definition` command


maciejmatczak:
With new functionality in https://github.com/foambubble/foam/pull/524, I uninstalled the https://marketplace.visualstudio.com/items?itemName=kortina.vscode-markdown-notes, but it seems the current implementation only supports a mouse click.

Hence a request - please add support for generic VS Code `Go to definition` action for wiki links. `Peek a definition` might also be handy.

Thanks

JonasSprenger:
After playing with Vscode API, it seem that the command you (@maciejmatczak) are looking for is `Open Link` (`@command:editor.action.openLink`).
It's the same command as `ctrl + click`, implemented with `interface DocumentLinkProvider<T extends DocumentLink = DocumentLink>`.
https://github.com/foambubble/foam/blob/7412d518d7aeb00976bfbf7b517db62d78f692f7/packages/foam-vscode/src/features/document-link-provider.ts#L29

The main difference is than `Go to definition` target a specific range in the destination file, whereas `Open Link` open just the file (so don't move the cursor if the file is already open).

A similar functionality which can be interesting to implement is `Go to References`/`Peek References` (it seems to me it exactly the same) with a `ReferenceProvider`. It should be the same as the backlinks of the targeted note.

maciejmatczak:
Markdown Notes mentions:
https://github.com/kortina/vscode-markdown-notes/blob/0ac9205ea909511b708d45cbca39c880688b5969/README.md#L19
> Markdown Notes provides syntax highlighting, auto-complete, Go to Definition (editor.action.revealDefinition), and Peek Definition (editor.action.peekDefinition) support for wiki-links to notes in a workspace.

Similar provider: https://github.com/kortina/vscode-markdown-notes/blob/0ac9205ea909511b708d45cbca39c880688b5969/src/MarkdownDefinitionProvider.ts

And I guess as definition is pretty much a language bases concept, it have to be supported by parser, right?

https://github.com/kortina/vscode-markdown-notes/blob/0ac9205ea909511b708d45cbca39c880688b5969/syntaxes/notes.tmLanguage.json#L11

JonasSprenger:
Thanks for the resources.

Indeed we can use `Go to Definition`, maybe by targeting the first header of the note or the whole file (using the range field).
But I think it's not a good idea to keep both commands to follow links (`Open Link` and `Go to Definition`).
`Go to Definition` seems more appropriated because it's more language specific.

[...]

PR #724, opened 2021-07-17
Implement WikilinkProvider (DocLinkProvider + DefinitionProvider) to add the 'Go to Definition' command

JonasSprenger:
A try to replace `DocumentLinkProvider` by `DefinitionProvider`.
But for creating the file if the link is a `Placeholder`, the `DefinitionProvider` API don't seems right.

I don't think it's worth to change.


JonasSprenger:
Original issuse #724

riccardoferretti:
Thanks for looking into this @JonasSprenger!
At a high level, I like the feature, it would be good to have it in Foam if we can make it fit into our UX.

Would this provider replace the `DocumentLinkProvider` or does it work alongside it? is there anything we are "losing" by dropping the document link provider?

> But for creating the file if the link is a Placeholder, the DefinitionProvider API don't seems right.

Why is that, can you elaborate?

JonasSprenger:
The call it's triggered on 'hover', so IMHO it's not adapted to create a new file if the link is a placeholder. Because it will necessarily happen inadvertently.

JonasSprenger:
> Would this provider replace the `DocumentLinkProvider` or does it work alongside it? is there anything we are "losing" by dropping the document link provider?

Yes, keeping both will be confusing, so it's one or the other.

By dropping the document link provider we are "losing":
  - The ability to keep the cursor at the last position if the linked note is already open.
  - Keeping the creation of an new note from a placeholder explicit (see below).  

> > But for creating the file if the link is a Placeholder, the DefinitionProvider API don't seems right.
> 
> Why is that, can you elaborate?

The call to the `DefinitionProvider` is triggered on 'hover', so if the link is a placeholder an new file is triggered immediately.
It will necessarily happen inadvertently, I do not expect any notes to be created until a deliberated action or command (like `Cmd+Click`).
It will imply the same "surprising" behavior as with vscode-markdown-notes #715

I have two suggestion https://github.com/foambubble/foam/issues/543#issuecomment-881911226
  - There is the possibility to use the `Open Link` command if the link doesn't exist and `Go to Definition` if it exist. But I don't thinks it's worth the increase of complexity for the user (two different commands for almost the same functionality).
 - An alternative solution is to implement the `hover-feature` to have some of the benefits of `Go to Definition`.

riccardoferretti:
If the two can coexist, I'd say let's have them coexist and use the right tool for the specific job, even if that means "splitting" functionality, no?
For code readability we could also create a `wikilink-providers.ts` that includes both the document link provider and the definition provider (as they are both related to wikilnks), do you think this would help achieve both goals (reduce confusion, full functionality)?

JonasSprenger:
> If the two can coexist, I'd say let's have them coexist and use the right tool for the specific job, even if that means "splitting" functionality, no?

You're right, there must be a way to have them coexist, and it's worth to make it configurable :
  - Navigating only with Links `DocumentLinkProvider` (like actual behavior).
  - Navigating only with Definition `DefinitionProvider`.
  - Combining both by using `DefinitionProvider` if the note already exit and `DocumentLinkProvider` for placeholder (creating and open a new note).

JonasSprenger:
> hi @JonasSprenger what is the status of this PR?

Hi @riccardoferretti,

I will try to add both Provider and make they coexist nicely with configuration.

riccardoferretti:
if you rebase from master now, the issue you are having with the tests should be resolved

packages/foam-vscode/package.json
```diff
@@ -235,10 +235,22 @@
...
+            "openLink",
+            "goToDefinitionButOpenLinkForCreation",
+            "goToDefinition",
+            "off"
+          ],
+          "enumDescriptions": [
+            "Use 'Open Link' command for navigation and creation",
+            "Use 'Go to Definition' command for navigation but use 'Open Link' for creation",
+            "Use 'Go to Definition' command for navigation and creation (Ctrl+hover over a placeholder will create and open the new note)",
+            "Disable navigation and creation"
```
JonasSprenger, line 252: What do you think about the formulation of the configuration options?

packages/foam-vscode/src/features/wikilink-provider.ts
```diff
@@ -0,0 +1,183 @@
...
+      }
+
+      const newNoteUri = toVsCodeUri(
+        URI.createResourceUriFromPlaceholder(
+          vscode.workspace.workspaceFolders[0].uri,
+          uri
+        )
+      );
+      const newNoteTitle = uri.path.split('/').slice(-1)[0];
+      await createNoteForPlaceholderWikilink(newNoteTitle, newNoteUri);
```
JonasSprenger, line 94: For creating a new note from a placeholder I used the same pattern as in `OPEN_COMMAND.execute` (`packages/foam-vscode/src/features/utility-commands.ts`), but only if a VsCode Workspace is open.

Is this the right way to go? @movermeyer  @riccardoferretti

packages/foam-vscode/src/features/wikilink-provider.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    const targetResource = this.workspace.get(uri);
+    const targetRange = findFirstLineOfContent(targetResource.source.text);
+
+    const result: vscode.LocationLink = {
+      originSelectionRange: toVsCodeRange(targetLink.range),
+      targetUri: toVsCodeUri(uri),
+      targetRange: null, // Disable the preview feature to avoid duplication with the over feature.
+      targetSelectionRange: toVsCodeRange(targetRange),
```
JonasSprenger, line 104: When following a link with the `Go To Definition` command, this allows to briefly highlight the first line of the MD content (usually the title).

packages/foam-vscode/src/features/wikilink-provider.ts
```diff
@@ -0,0 +1,183 @@
...
+    toVsCodeRange(link.range),
+    command
+  );
+  documentLink.tooltip = URI.isPlaceholder(target)
+    ? `Create note for '${target.path}'`
+    : `Go to ${URI.toFsPath(target)}`;
+  return documentLink;
+};
+
+export const findFirstLineOfContent = (rawContent: string): Range => {
```
JonasSprenger, line 149: Is it OK to implement a custom parsing algorithm ? In the long term, it's maybe not the ideal place to stay.

I tried to use `targetResource.source.contentStart` but it's not sufficient, and it skips systematically two lines after the front matter if there is one (not sure if it's the intended behavior).

JonasSprenger:
Hello @riccardoferretti,

I've finally come up with a reasonable solution (imho) to give the possibility to choose which command to use for opening links. With one possible option to combine both providers: `DefinitionProvider` to open existing notes and `DocumentLinkProvider` for creating a new note by `Ctrl+click` on a placeholder, the goal is to keep the creation of a new note (file) more explicit compared to `Ctrl+over` when using only `DefinitionProvider`.

Observation: When changing the configuration, the opens editors don't take the modifications in account, it is necessary to select another tab then re-select the original one for the changes to take effect (it's not a big deal in my opinion, but good to known when doing manual testing).

Apart from the missing tests, I have nothing to add or modify, so it's good for a review.

packages/foam-vscode/package.json
riccardoferretti: What is the difference in behavior between `openLink` and `goToDefinitionButOpenLinkForCreation`?

As in, not technically, but what changes for the user?

packages/foam-vscode/src/features/wikilink-provider.ts
```diff
@@ -0,0 +1,183 @@
...
+    token: vscode.CancellationToken
+  ): Promise<vscode.LocationLink[] | vscode.Definition> {
+    const navMode = this.getNavigationMode();
+    if (navMode === OPEN_LINK_NAV_MODE || navMode === NAVIGATION_DISABLED) {
+      return;
+    }
+
+    const resource = this.parser.parse(document.uri, document.getText());
+    const targetLink: ResourceLink | undefined = resource.links.find(link =>
+      Range.containsPosition(link.range, {
```
riccardoferretti, line 68: can't you just use `Range.containsPosition(link.range, position)` here?

packages/foam-vscode/src/features/wikilink-provider.ts
riccardoferretti: My take here is that it's not this provider that should be concerned with the workspace status, but the command itself - as we'd want the behavior to be consistent across "callers"

what's the thinking behind the "exception"?

packages/foam-vscode/src/features/wikilink-provider.ts
riccardoferretti: Is this moving the cursor to that position? what if the note was already open? would it displace the former cursor position?

packages/foam-vscode/src/features/wikilink-provider.ts
riccardoferretti: Feels like a bug with `contentStart` (but I wouldn't fix it in this PR).

riccardoferretti (commented):
Thanks for this, I left a couple of minor comments - the most important thing I want to fully understand is the effective change in behavior amongst the configuration options, so we can decide how flexible to be about them

packages/foam-vscode/src/features/wikilink-provider.ts
JonasSprenger: I agree, this PR it's not the place to fix it.

riccardoferretti:
Hi @JonasSprenger, based on this PR I have created https://github.com/foambubble/foam/pull/821

Let me know if you have any comment there, and thanks for spear-heading the effort (it was easy for me to implement that based on the work you did here!)

Closing this PR as it is now replaced by #821

JonasSprenger:
Hi @riccardoferretti,

Sorry for the long delays of my contributions and my responses. I was struggling and I didn't achieve to make a unit test for the creation of a new note within the DefinitionProvider. But most of all I struggle with procrastination... :unamused:

I'm glad my contribution was still useful.

Concerning your implementation, I like the pragmatic approach to not bother with configurability, and +1 to provide References of the Backlinks.

PR #728, opened 2021-07-18
Implement hover feature for wikilinks

### LINKED ISSUE #727: Display note content on `Hover` for wikilinks
# Feature
Displaying the content of a note targeted by a wikilink in a tooltip-like widget,
by hovering the wikilink with the mouse or by triggering a command.
  - `editor.action.showDefinitionPreviewHover`
  - or `editor.action.showHover` (`Ctrl+K Ctrl+I`).

The main benefice is to quickly read the targeted note without switching context by changing editor focus.

# Implementation
 By adding an [HoverProvider](https://code.visualstudio.com/api/references/vscode-api?ref=hackernoon.com#HoverProvider)

JonasSprenger:
resolve #727

### DIFF
--- a/packages/foam-vscode/src/features/hover-provider.spec.ts
+++ b/packages/foam-vscode/src/features/hover-provider.spec.ts
@@ -79,21 +79,11 @@ eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee`;
     const noteA = parser.parse(fileA.uri, fileA.content);
     const noteB = parser.parse(fileB.uri, fileB.content);
 
-    // Create custom workspace fixture.
-    const workspace = new FoamWorkspace();
     const matcher = new Matcher([URI.file('/')], ['**/*']);
-    const resourceProvider: MarkdownResourceProvider = new MarkdownResourceProvider(
-      matcher,
-      undefined,
-      undefined,
-      {
-        read: _ => Promise.resolve(fileBContent),
-        list: _ => Promise.resolve([]),
-      }
-    );
+    const resourceProvider = new MarkdownResourceProvider(matcher);
 
+    const workspace = new FoamWorkspace();
     workspace.registerProvider(resourceProvider);
-
     workspace.set(noteA).set(noteB);
 
     const { doc } = await showInEditor(noteA.uri);
--- a/packages/foam-vscode/src/features/hover-provider.spec.ts
+++ b/packages/foam-vscode/src/features/hover-provider.spec.ts
@@ -75,18 +75,18 @@ describe('Hover provider', () => {
     const { doc } = await showInEditor(noteA.uri);
     const pos = new vscode.Position(0, 22); // Set cursor position on the wikilink.
 
-    const result: vscode.Hover = provider.provideHover(
+    const promiseResult = provider.provideHover(
       doc,
       pos,
       noCancelToken
-    ) as vscode.Hover;
+    ) as Promise<vscode.Hover>;
 
-    expect(result.contents.length).toEqual(1);
-    const mdResult = result.contents[0] as vscode.MarkdownString;
-    expect(mdResult.value).toEqual(
-      '\n```markdown\n' + fileB.content + '\n```\n'
-    );
-    expect(mdResult.isTrusted).toBeFalsy();
+    return promiseResult.then(result => {
+      expect(result.contents).toBeDefined();
+      expect(result.contents).toEqual(
+        '\n```markdown\n' + fileB.content + '\n```\n'
+      );
+    });
   });
 
   it('should not return hover content when the cursor is not placed on a wikilink', async () => {
--- a/packages/foam-vscode/src/features/hover-provider.spec.ts
+++ b/packages/foam-vscode/src/features/hover-provider.spec.ts
@@ -1,0 +1,161 @@
+import * as vscode from 'vscode';
+import { FoamWorkspace, createMarkdownParser } from 'foam-core';
+import {
+  cleanWorkspace,
+  closeEditors,
+  createFile,
+  createTestWorkspace,
+  showInEditor,
+} from '../test/test-utils';
+import { HoverProvider } from './hover-provider';
+
+describe('Hover provider', () => {
+  const parser = createMarkdownParser([]);
+
+  beforeAll(async () => {
+    await cleanWorkspace();
+  });
+
+  afterAll(async () => {
+    await cleanWorkspace();
+  });
+
+  beforeEach(async () => {
+    await closeEditors();
+  });
+
+  it('should not return hover content for empty documents', async () => {
+    const { uri, content } = await createFile('');
+    const ws = new FoamWorkspace().set(parser.parse(uri, content));
+
+    const provider = new HoverProvider(ws, parser);
+
+    const doc = await vscode.workspace.openTextDocument(uri);
+    const pos = new vscode.Position(0, 0);
+    const result: vscode.ProviderResult<vscode.Hover> = provider.provideHover(
+      doc,
+      pos,
+      noCancelToken
+    );
+
+    expect(result).toBeUndefined();
+  });
+
+  it('should not return hover content for documents without links', async () => {
+    const { uri, content } = await createFile(
+      'This is some content without links'
+    );
+    const ws = new FoamWorkspace().set(parser.parse(uri, content));
+
+    const provider = new HoverProvider(ws, parser);
+
+    const doc = await vscode.workspace.openTextDocument(uri);
+    const pos = new vscode.Position(0, 0);
+    const result: vscode.ProviderResult<vscode.Hover> = provider.provideHover(
+      doc,
+      pos,
+      noCancelToken
+    );
+
+    expect(result).toBeUndefined();
+  });
+
+  it('should return hover content for a wikilink', async () => {
+    const fileB = await createFile('# File B\nThe content of file B');
+    const fileA = await createFile(
+      `this is a link to [[${fileB.name}]] end of the line.`
+    );
+    const noteA = parser.parse(fileA.uri, fileA.content);
+    const noteB = parser.parse(fileB.uri, fileB.content);
+    const ws = createTestWorkspace()
+      .set(noteA)
+      .set(noteB);
+
+    const provider = new HoverProvider(ws, parser);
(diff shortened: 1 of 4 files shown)


JonasSprenger:
Screenshot from 2021-07-19 01-09-43

riccardoferretti:
Looks like the right approach to me.

I noticed that the preview is not rendered, we already have a pattern for extracting/clean up markdown text for rendering, you can check the following:

- `getNoteTooltip` in `utils.ts`

- `ResourceTreeItem.resolveTreeItem()` in `grouped-resources-tree-data-provider.ts` for usage

let me know if you have any issue integrating with those, thanks!

JonasSprenger:
So we should not display the full content of the note on hover ?

packages/foam-vscode/src/features/hover-provider.spec.ts
```diff
@@ -75,18 +75,18 @@ describe('Hover provider', () => {
...
-    const mdResult = result.contents[0] as vscode.MarkdownString;
-    expect(mdResult.value).toEqual(
-      '\n```markdown\n' + fileB.content + '\n```\n'
-    );
-    expect(mdResult.isTrusted).toBeFalsy();
+    ) as Promise<vscode.Hover>;
+
+    return promiseResult.then(result => {
+      expect(result.contents).toBeDefined();
+      expect(result.contents).toEqual(
```
JonasSprenger, line 86: This test don't work we received always an empty string!
While by doing manual tests it work as expected.

I don't understand why? Maybe the fixtures are badly initialized with the promises...
(I will check an other moment)

JonasSprenger (commented):
Screenshot from 2021-07-21 00-57-08

It seems to work, but I can't make the test work correctly...

JonasSprenger:
> Looks like the right approach to me.
> 
> I noticed that the preview is not rendered, we already have a pattern for extracting/clean up markdown text for rendering, you can check the following:
> 
>     * `getNoteTooltip` in `utils.ts`
> 
>     * `ResourceTreeItem.resolveTreeItem()` in `grouped-resources-tree-data-provider.ts` for usage
> 

Thank for your feedback, I think I managed to use the utils you mentioned.
 I still have to make the test work

packages/foam-vscode/src/features/hover-provider.spec.ts
```diff
@@ -75,18 +75,18 @@ describe('Hover provider', () => {
...
       noCancelToken
-    ) as vscode.Hover;
-
-    expect(result.contents.length).toEqual(1);
-    const mdResult = result.contents[0] as vscode.MarkdownString;
-    expect(mdResult.value).toEqual(
-      '\n```markdown\n' + fileB.content + '\n```\n'
-    );
-    expect(mdResult.isTrusted).toBeFalsy();
+    ) as Promise<vscode.Hover>;
```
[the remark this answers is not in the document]
riccardoferretti: in typescript you can use `await` to make this code synchronous.

```js
const hover = await provider.provideHover(...)
expect(hover.contents).toBeDefined();
...
```

Casting shouldn't be necessary, let me know if the above doesn't work for some reason!

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+import { toVsCodeRange } from '../utils/vsc-utils';
+import { ResourceLink } from 'foam-core';
+import { getFoamVsCodeConfig } from '../services/config';
+
+const feature: FoamFeature = {
+  activate: async (
+    context: vscode.ExtensionContext,
+    foamPromise: Promise<Foam>
+  ) => {
+    if (!getFoamVsCodeConfig('links.hover.enable')) {
```
riccardoferretti, line 14: if we put this check here, changes in the configuration won't be reflected in the editor (the user would have to reload the editor).

I can see two options here:
- do the check in the `provideHover` function
- use `monitorFoamVsCodeConfig` and pass the function (see `document-decorator.ts`)

The first approach is more straightforward but I don't love the "singleton magic".
The second approach has more explicit dependency injection and is the pattern I am currently favoring (although I am open to different arguments).

Let me know your thoughts!

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const targetUri = this.workspace.resolveLink(startResource, targetLink);
+    if (URI.isPlaceholder(targetUri)) {
+      return;
+    }
+
+    const resultP: Promise<
+      vscode.Hover | undefined
+    > = this.workspace.readAsMarkdown(targetUri).then(
```
riccardoferretti, line 61: you should be able to use `async`/`await` here to simplify this code?

riccardoferretti (commented):
Very good job on the tests!
The implementation looks good to me as well, I have left a couple of comments about config management and async/await feature in TS.
Thanks!

packages/foam-vscode/src/features/hover-provider.spec.ts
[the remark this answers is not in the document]
JonasSprenger: It work properly with async/await https://github.com/foambubble/foam/pull/728/commits/b3505975b460e046e99325799fb7e821292dc38e

packages/foam-vscode/src/features/hover-provider.spec.ts
```diff
@@ -61,17 +67,43 @@ describe('Hover provider', () => {
...
+      matcher,
+      undefined,
+      undefined,
+      {
+        read: _ => Promise.resolve(fileBContent),
+        list: _ => Promise.resolve([]),
+      }
+    );
+
+    workspace.registerProvider(resourceProvider);
```
JonasSprenger, line 94: https://github.com/foambubble/foam/pull/728#discussion_r673570039#728
For this test we need a specific `resourceProvider` so we can't used the default `FoamWorkspace` fixture created by `createTestWorkspace()`

packages/foam-vscode/src/features/hover-provider.spec.ts
JonasSprenger: Resolved by https://github.com/foambubble/foam/pull/728#discussion_r676728842

packages/foam-vscode/src/features/hover-provider.spec.ts
riccardoferretti: why is that the case? 
I am a bit surprised, but we should document the situation here if there is no alternative

packages/foam-vscode/src/features/hover-provider.ts
JonasSprenger: I tried to implement the second approach in https://github.com/foambubble/foam/pull/728/commits/4f407f186ba0661e29f098c771318c1567cbf85b What do you think ?

For the first approach, I'm concerned about reading the configuration each time.

packages/foam-vscode/src/features/hover-provider.spec.ts
JonasSprenger: Because the fixture is implemented with an `MarkdownResourceProvider` returning an empty string on read, regardless of the content of the files.

```typescript
export const createTestWorkspace = () => {
  const workspace = new FoamWorkspace();
  const matcher = new Matcher([URI.file('/')], ['**/*']);
  const provider = new MarkdownResourceProvider(matcher, undefined, undefined, {
    read: _ => Promise.resolve(''), // <=== Here
    list: _ => Promise.resolve([]),
  });
// ... ... ...
};
```

JonasSprenger: Because the fixture is implemented with an `MarkdownResourceProvider` returning an empty string on read, regardless of the content of the files.

```typescript
export const createTestWorkspace = () => {
  const workspace = new FoamWorkspace();
  const matcher = new Matcher([URI.file('/')], ['**/*']);
  const provider = new MarkdownResourceProvider(matcher, undefined, undefined, {
    read: _ => Promise.resolve(''), // <=== Here
    list: _ => Promise.resolve([]),
  });
// ... ... ...
};
```

riccardoferretti: I see, but why can't we use the regular resource provider? what would break and why?

JonasSprenger:
> hi @JonasSprenger what is the status of this PR?

Hi @riccardoferretti,

Sorry, I messed up with GitHub review functionalities, so a few of my early comments was never published (in pending states).

The hover feature is working fine, and I made some changes based on your feedback. So this PR is good for a last review.

A recap of my modifications :
- Use async/await
- Use of `monitorFoamVsCodeConfig` so modification of the setting is applied directly (without reloading vscode).
- Fix test and use regular resource provider.

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+  ): Promise<vscode.Hover> {
+    if (!this.isHoverEnabled()) {
+      return;
+    }
+
+    const startResource = this.parser.parse(document.uri, document.getText());
+
+    const targetLink: ResourceLink | undefined = startResource.links.find(
+      link =>
+        link.type === 'wikilink' &&
```
riccardoferretti, line 55: why limit this feature to wikilinks? can't we add a hover for `[direct links](./path.md)` too? Foam supports them

riccardoferretti (commented):
This is good to go, I would only add support for direct links first

packages/foam-vscode/src/features/hover-provider.ts
riccardoferretti: tested locally, even just removing the line will make it work. we should add a test case though

JonasSprenger:
> This is good to go, I would only add support for direct links first

I remove the filter on `wikilink` so it works also for `direct links`.

And I refactored the test suite, mainly to have a homogeneous instantiation in each test.

packages/foam-vscode/src/features/hover-provider.ts
JonasSprenger: At first, I thought the semantics of direct links were significantly different from wikilinks.

Now it works for both https://github.com/foambubble/foam/pull/728/commits/beee772f9c9e8e026959c06f13f2d29fe275b899

riccardoferretti (approved):
Looking great!

riccardoferretti:
@allcontributors add @JonasSprenger for code

PR #730, opened 2021-07-19
Respect the ignore globs when listing all files

pderaaij:
This change ensures Foam respects the ignored folders when calling the datastore. It utilises the ignore option of the `glob` module.

Note, to properly ignore my `node_modules` folder in `layouts` I need to configure the ignore glob into: `"**/_layouts/**"`. I do not use `/*` at the end as that will cause it to traverse the entire node_modules folder.

### DIFF
--- a/packages/foam-core/src/markdown-provider.ts
+++ b/packages/foam-core/src/markdown-provider.ts
@@ -59,7 +59,9 @@ export class MarkdownResourceProvider implements ResourceProvider {
 
   async init(workspace: FoamWorkspace) {
     const filesByFolder = await Promise.all(
-      this.matcher.include.map(glob => this.dataStore.list(glob))
+      this.matcher.include.map(glob =>
+        this.dataStore.list(glob, this.matcher.exclude)
+      )
     );
     const files = this.matcher
       .match(filesByFolder.flat())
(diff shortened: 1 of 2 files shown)


riccardoferretti:
This looks really good and will probably help a lot with performance on bigger repos.

A few questions:

- why the difference between `**` vs `**/*`? I am surprised the glob library is not a bit smarter around that? This feels like a bit of quirk from a user's perspective, I wonder wether we should replace `**/*` with `**` in the globs ourselves? (but who wants to mess with globbing?? 😅 )
- I see that the function supports passing an array of string, have you checked that works correctly?
- did you get a chance to see how performance changes? (e.g. on your machine by ignoring the `node_modules` dir)

Thanks!

pderaaij:
The argument syntax is exactly the same as the one of the glob module. Tested it i with my own config that has an array of strings and that works. On my machine it saved 9 to 10 seconds when ignoring the node_modules folder. 

My guess/assumption with `**/*` versus `**` is that in the latter you only put a wildcard on the folder, without caring what's in it. The first depicts more the inside of folders and the files they contain. As such it list the entire folder structure with the first glob.

riccardoferretti (approved):

Issue #530, opened 2021-03-15
Foam Graph: What's next?


jmg-duarte:
# Foam Graph

This is a mixture between current state of affairs and a call to action.

That is, every issue and possible improvement point you see here can be thought as a place where Foam can be improved.
As improvements will be required to be PR'd into the repository, discussion is also open

### Customization

This year we added some customization settings to the graph in: <https://github.com/foambubble/foam/pull/438>

We hope to add many more during 2021.

#### Coloring by Date

Recently, I saw the following issue: <https://github.com/foambubble/foam/pull/485>

In summary, the issue suggests nodes to be colored according to "Last Modification Date".
While I do not think it is a feature that should get into the core,
I can understand why it would be an awesome feature and I have some ideas.

If you use `git` you can highlight nodes according to their status, or track the knowledge-base progress through time.
This of course may not even be possible, but would be cool nonetheless.

#### Coloring by Attributes

I consider attributes to be embedded metadata on the file.
These can be tags, or other "keys" defined in the YAML frontmatter (see <https://github.com/foambubble/foam/issues/388>).

This is inspired by <https://github.com/foambubble/foam/issues/2>.

### Filtering

Filtering should not be too hard to implement, however we are still thinking of ways to filter nodes.

#### Existing Requests

- By tags - <https://github.com/foambubble/foam/issues/496>
- By folder - <https://github.com/foambubble/foam/issues/452>

#### Text Filtering

Filtering would not be complete without filtering by text (even though I forgot about it).
If the search feature is available in the VSCode API, we can take advantage of that too.

Filtering by date could also be a thing.
Being honest, if the filtering is flexible enough we can filter on a ton of things, but do not get your hopes up just yet.

Regardless, we are open to suggestions!

### Grouping

I will preface this part by saying that *grouping is (almost) entirely dependent on the graph library*.

Why almost you ask? Because we can always add some fake nodes and hack our way into have a semi supported version of grouping.
Grouping by folder would actually work that way by nature, but tags would require a second graph to be built.

The ideas around grouping are the same as for filtering:
- By tags
- By folder

### Data Dump

This is useful for many reasons, mostly for support on the web and other visualization tools. <https://github.com/foambubble/foam/issues/308>

While it is not supported yet, adding it in is just a matter of deciding on the graph structure and commit to it.

There are several formats we can pick from, for example Turtle (<https://en.wikipedia.org/wiki/Turtle_(syntax)>).

### Performance

[...]

PR #737, opened 2021-08-04
Issue #530: Graph Improvements

dannysemi:
Hello! This is my first attempt at contributing to an open source project. I attempted to address many of the areas identified in regards to the graph visualization. I'm happy with the outcome, but I did make a lot of changes to the core graph logic. It's all backwards compatible as far as I can tell, aside from the actual visualization. It could be made to work with the old force-graph, but I chose to implement cytoscape.js both for handling the graph data in core and the visualization in the webview. I don't know how fast this is. I don't know how to test for that. The highlights:
- ~~reworked foam-core/graph by implementing cytoscape.js to handle all of the graph data~~
- implemented a basic gui and filtering in the data visualization (dat.gui made filtering kind of weird, should a panel be used instead?)
- ~~implemented a command to expose the graph json (I'm fairly certain I put this in the wrong place)~~
- add `color: "red"` to yaml frontmatter to change node colors easily. Accepts these formats: `"red" / "#ff0000" / rgb(255,0,0)`
- highlight nodes based on text search (todo: use regex instead of simple string matching)
- folder hierarchy view
- visualize tags on graph
- filter by tagged nodes (todo: implement filtering by specific tags)
- limit graph view to selected nodes (Hold shift or ctrl to add nodes to selection, or use box select. Note: this doesn't play well with any of the other filters. I need update the gui to disable the other parameters when this is selected.)

I'm still working on the styling, but it is still customizable. The formatting is different though so try something like this to test it out:
```
"foam.graph.style" : { "styles": [{
      "selector": "node[type = 'note']",
      "style": {
        "background-color": "cyan",
      }
    }, {
      "selector": "node[type = 'tag']",
      "style": {
        "background-color": "purple",
      }
    }, {
      "selector": "node[type = 'placeholder']",
      "style": {
        "background-color": "pink",
      }
    }
  ]}
```

Hopefully this works out and can be used as a basis for further improvements. I'm very new to typescript and vscode extensions, so I'm sure there is room for improvement. Let me know and I'll try to get things fixed as soon as I can.


dannysemi:
#530

dannysemi:
I just realized that there are tests. This is currently failing them. I'll work on that.

riccardoferretti:
hey @dannysemi! thanks for contributing to the project.
Let me start with a core question first, so I can give you more appropriate feedback: what is the main goal of the PR? What problem is it solving?

dannysemi:
I'm trying to make it easier to interact with the graph.

dannysemi:
@riccardoferretti When I originally started this it felt necessary to convert all of the original graph to cytoscape. I realize now that was probably unnecessary. I can revert the changes to the graph core and move the necessary conversions to dataviz. Am I headed in the right direction with the graph. Where are we at on this?

riccardoferretti:
> @riccardoferretti When I originally started this it felt necessary to convert all of the original graph to cytoscape. I realize now that was probably unnecessary. I can revert the changes to the graph core and move the necessary conversions to dataviz. Am I headed in the right direction with the graph? Where are we at on this?

Hi @dannysemi, yes, I agree with moving the cytoscape changes to the dataviz is better than changing the core graph model. I even wonder whether it needs to live in Foam VS Code, or in the graph webview deps, what's your take on that?

dannysemi:
> Hi @dannysemi, yes, I agree with moving the cytoscape changes to the dataviz is better than changing the core graph model. I even wonder whether it needs to live in Foam VS Code, or in the graph webview deps, what's your take on that?

@riccardoferretti I should be able to do everything from within the webview. I will just need to tweak `generateGraphData` to make sure note properties and tags will be included in the data sent to the webview. How do you want to handle text search? Should I send the source text to the webview or make a request from the webview and return matching URIs in dataviz? My end goal for this is to have a basic set of features that work well in the graph, but also expose the graph data so I can create my own extension and do whatever I want with the data. Exposing the full source text would present privacy concerns in that case, so maybe the second option?

riccardoferretti:
> I will just need to tweak generateGraphData to make sure note properties and tags will be included in the data sent to the webview

Yup sending the note properties sounds like a good idea

> How do you want to handle text search?

This is an interesting point, probably it requires a deeper look.
My first instinct is to not send the text along to the graph, as the payload would dramatically increase.
An approach would be to expose a search api in the channel between the webview and dataviz, and use the VS Code side to run the query and return the results. This is just off the top of my head, unclear whether there are implications I am missing; it seems like it should handle the base case pretty well though.

> My end goal for this is to have a basic set of features that work well in the graph, but also expose the graph data so I can create my own extension and do whatever I want with the data

Can you share a couple of examples? Especially when it comes to e.g. search.
I agree about the privacy concerns (even though that could be said for any extension installed in VS Code)

> I couldn't find a way to utilize vscode's native search at all. Let me know if I missed something.

That could be an interesting approach, again it kinda depends on the flow you are thinking about, let me know

dannysemi:
> > My end goal for this is to have a basic set of features that work well in the graph, but also expose the graph data so I can create my own extension and do whatever I want with the data
> 
> Can you share a couple of examples? Especially when it comes to e.g. search.
> I agree about the privacy concerns (even though that could be said for any extension installed in VS Code)

Sorry, I wasn't necessarily referring to search there. I had too many ideas going at once. I am working on implementing search just like you mentioned.

What I meant by exposing graph data to other extensions is essentially converting the api between foam-vscode and the webview so that other extensions can have access to the data. So that if I want to encode custom properties in my frontmatter I can then have my own extension that visualizes that data. For example, I could create a property `Priority: 1` and then request the graph data from my extension and create a webview that displays a table of notes sorted by priority. That way foam doesn't have to support my specific properties or visualization and I have more freedom for how I use my data. In that simple example it wouldn't be very difficult to just extract that information from the raw markdown in my extension, but I want access to the whole graph.

^This probably belongs in a separate issue/PR.

dannysemi:
@riccardoferretti Okay, I reverted the changes in graph core and moved all of my graph changes to the webview. Only had to make a few small changes in dataviz and I'm liking where the graph is at. From here I think it's mostly styling and maybe some small edits to the UI or filters to get the desired result.

riccardoferretti:
I tried the new graph code and I like the flexibility and the options that cytoscape + DAT provide, but I am having serious performance issues on my test repo (~500 notes). Are you seeing the same?
Also looking at their own performance tests (https://cytoscape.org/js-perf/) I am concerned that the library might not cope well with medium and large repos, what do you think?

dannysemi:
> I tried the new graph code and I like the flexibility and the options that cytoscape + DAT provide, but I am having serious performance issues on my test repo (~500 notes). Are you seeing the same?
> Also looking at their own performance tests (https://cytoscape.org/js-perf/) I am concerned that the library might not cope well with medium and large repos, what do you think?

Do you have something you can share with a lot of nodes? Mine doesn't have enough to produce an issue. Is the performance a problem during graph updates or just the initial load? I can look at optimization. Did you have any ideas on an alternative? 'Roll your own' isn't very appealing to me.

chrisUsick:
Hi, I’ve been following this thread just because I've researched JavaScript
graph/network visualization tools in the past.
https://visjs.org/
Is the best open source option I can find. Besides having the performance
to do 1000s of nodes (I have >1000 Notes in my personal foam) it has a ton
of flexibility.
The data model allows you to efficiently load dynamic data, so you could do
things like incremental loading easily. That would help if there was a
initialization performance issue. The data model also supports filtering I
believe. Next, the library supports a Ton of layout options with dozens of
examples to work off of.

dannysemi:
Okay I tried to optimize the graph using these recommendations: [https://js.cytoscape.org/#performance/optimisations](https://js.cytoscape.org/#performance/optimisations). That should make viewport updates a lot faster with large graphs. I can't really do anything about how long it takes for the layouts to run. That may be a library limitation. The layout updates should only occur when the graph data is updated or changing view options. The initial load time is slow because I'm doing all of the data formatting at load and pre-computing the hierarchy layout so that it is predictable.

@chrisUsick I have no library loyalty. If this isn't fast enough I will check out vis.js. Thanks for the heads up. I only chose cytoscape originally because it seemed very flexible and was one of the first search results I got.

dannysemi:
I think I'm going to have to generate some test data unless somebody has a repo with a large graph I can use. If I grab some data from wikipedia I should be able to convert that into markdown with links, right?

dannysemi:
@riccardoferretti I tested graph load times between the original force-graph, cytoscape, and vis-network on a graph with 840 nodes and 1000 links. force-graph was 6x faster than cytoscape and 3x faster than vis. It shouldn't be too hard to just implement what I've done with tags and filtering using the original force-graph. I don't think speed should be the only consideration here though. My problem with the original graph is that it doesn't provide much information and the layout is unpredictable. In my opinion, the graph should have some predictability in at least one of its views. force-graph also doesn't have any support for clustering out of the box which seems like an issue for large graphs.

I think if I can use [cola.js](https://ialab.it.monash.edu/webcola/index.html) with force-graph the way the library says it supports d3 then that could supplement some of the functionality that I think the graph is missing. Hopefully that would provide a good balance of speed and functionality.

This isn't what I had in mind when I started this, but I think I'm on the right track now.

dannysemi:
I should probably clarify that I think the graph should be a means of visualizing the links in the data and navigating the data. So I want to have a view that is essentially the current graph, one that is a hierarchy based on the folder structure (to help navigate large graphs), and one that clusters/groups nodes based on tags. Additionally I want to highlight nodes based on text search, and limit what nodes appear on the graph. That's what I'm shooting for with this PR. I managed that more or less with cytoscape. The initial load time and view changes were slow, but the graph still functioned fine after the layout finished loading.

riccardoferretti:
I totally agree with the direction you are taking and I really appreciate the work you are doing.

I believe the work you have done on the DAT side is really good, and we should be able to benefit from it regardless of the graph rendering lib.

Given the better performance of force-graph, and the fact that is built on top of d3, I agree with you that investing in it is the best way forward at this point.

I would also suggest breaking things into smaller PR, and have an implementation plan, e.g.:

1. introduce DAT to graph UI and augment graph model with properties and tags
2. show/hide tags
3. text search
4. hierarchical visualization

This is just an example to illustrate the idea.

Let me know your thoughts, I really like the shape this is taking

dannysemi:
> I would also suggest breaking things into smaller PR, and have an implementation plan, e.g.:

Sorry, I'm new to this. Do you mean I should have an implementation plan for each PR or is just each step its own PR?

riccardoferretti:
What I mean is that having one PR with the code changes for various features (adding DAT, adding support for hierarchical view, support for tags, ....), would result in a big code change, and make it harder to assess what change is required by which of the features included - and therefore harder to review.

Instead breaking things down into individual units would make the code changes much more specific and easier to review.

So we can use this PR for the first change (e.g. adding basic DAT filtering and the related properties in the metadata). Then once this is reviewed and merged, create another PR for the text search. After that another PR for the hierarchical view, and so on...
(keep in mind these are just examples, you can break the features apart as you see fit to properly organize the review process)

Does that clarify? Let me know if something is not clear, I am happy to further explain

riccardoferretti:
Somewhat related, this issue was just created: https://github.com/foambubble/foam/issues/744

dannysemi:
> Does that clarify? Let me know if something is not clear, I am happy to further explain

That makes sense. There won't be much sense in implementing the DAT gui without some kind of view change though. So for this first one I will augment the graph data with the tags and properties and implement the gui to allow displaying tags on the graph.

riccardoferretti:
cool, let me know when the PR is ready for review

dannysemi:
@riccardoferretti I was able to implement the first set of changes. There is an issue that I couldn't figure out though. In order to change filters, I saved a copy of the graph data on the model object. Then when I make a view change using the gui, I create a deep copy of that original graph data, perform the filter and then update the graph using the same method as the original load but with the filtered data. This all works fine, but I think that deep copy is causing some weirdness. The hover effects no longer work, and the draw loop is throwing an error every frame looking for a node that doesn't exist. It seems like those functions are looking at a different version of the data.

Not sure what to do about that. If you have any ideas on how to perform the filter differently, let me know.

dannysemi:
Took some digging in the debugger, but now everything is working as intended. Should be ready for review.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -164,6 +167,33 @@ const Actions = {
     };
     graph.backgroundColor(model.style.background);
   },
+  evaluate: () => {
```
riccardoferretti, line 170: I am not sure I understand what this function does and why it's called `evaluate`, can you help me understand?

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -223,12 +253,34 @@ function initDataviz(channel) {
     .onBackgroundClick(event => {
       Actions.selectNode(null, event.getModifierState('Shift'));
     });
+    const gui = new dat.gui.GUI();
+    gui.add(model, 'View', {Default: "default", Tags: "tags"})
```
riccardoferretti, line 257: same as above, let's use the JS naming convention

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -344,8 +396,10 @@ try {
     switch (message.type) {
       case 'didUpdateGraphData':
         const graphData = augmentGraphInfo(message.payload);
-        console.log('didUpdateGraphData', graphData);
+        model.graphData = graphData;
         Actions.refresh(graphData);
+        console.log('didUpdateGraphData', graphData, model.graphData);
+        Actions.evaluate();
```
riccardoferretti, line 402: the fact that here we call `refresh`, then `evaluate`, which in turn calls `refresh` again, feels odd - am I missing something?

packages/foam-core/src/markdown-provider.ts
```diff
@@ -380,10 +378,7 @@ export function createMarkdownParser(
               }
             }
           } catch (e) {
-            Logger.warn(
-              `Error while parsing YAML for [${URI.toString(uri)}]`,
-              e
-            );
+            Logger.warn(`Error while parsing YAML for [${uri}]`, e);
```
riccardoferretti, line 381: how come removing the `toString` from the logging statements?

riccardoferretti (commented):
thanks for this, I have left a few comments, let me know your thoughts

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -63,6 +64,7 @@ let model = {
    * in the case it fails, use the fallback style values.
    */
   style: defaultStyle,
+  View: "default",
```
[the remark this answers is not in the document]
dannysemi: DAT.gui uses the property name as the label by default. I just had to look up how to change the label. Fixed.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -164,6 +167,33 @@ const Actions = {
     };
     graph.backgroundColor(model.style.background);
   },
+  evaluate: () => {
+    // create deep copy of the graph data as a reference for the filter
+    const graphData = JSON.parse(JSON.stringify(model.graphData));
+    let types = [];
+    switch(model.View){
+      case "default":
+        types = ["note","placeholder"];
```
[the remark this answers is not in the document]
dannysemi: I am aware of this. Took me a minute to figure out why `note-properties` kept getting removed from the graph. I still think that is a reasonable default view. In the future, I want the filter to be customizable. For now I could add a view that displays all nodes.

packages/foam-vscode/static/dataviz/graph.js
dannysemi: It checks the values of the gui controllers (only one currently, but there will be more), modifies the graph data accordingly, and then updates the graph. Would it be better if I called it something like `evaluateControllers`?

packages/foam-vscode/static/dataviz/graph.js
dannysemi: Fixed.

packages/foam-vscode/static/dataviz/graph.js
dannysemi: No, it was redundant. I simplified things now.

packages/foam-core/src/markdown-provider.ts
dannysemi: I accidentally reverted to the wrong commit. I thought I fixed it, but I was struggling with git.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -223,12 +243,35 @@ function initDataviz(channel) {
     .onBackgroundClick(event => {
       Actions.selectNode(null, event.getModifierState('Shift'));
     });
+    const gui = new dat.gui.GUI();
```
riccardoferretti, line 246: let's fix the indentation here

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -63,6 +64,7 @@ let model = {
    * in the case it fails, use the fallback style values.
    */
   style: defaultStyle,
+  view: "default",
```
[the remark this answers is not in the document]
riccardoferretti: Also, there is something that I am a bit concerned about, let me know your thoughts:

Imagine the scenario of having a VS Code command that resets the filtering on the notes in the graph.
DAT uses 2 way data binding, which means that in the current approach to achieve this I would have to change the model of DAT to achieve the result. This feels wrong to me.

Instead we should have a `showNodesByType` which has an array of types, or `null/undefined/[]` if we want to show everything.
This means this would need to be kept in sync with the DAT gui.
Crucially, the source of truth need to be the core model, not the DAT gui model.

I have seen two things that might help with this:
- in https://github.com/vasturiano/force-graph there is a `onRenderFramePost` that we could use to sync the state of the DAT gui with the one of the core model (the reason why I would put it there is because the `render` is already managed by the graph object, we might as well put everything there)
- in https://github.com/dataarts/dat.gui there is an `updateDisplay` function that could do just that. I wonder though if calling it on every render is going to cause issues

Not sure what the ideal here is, maybe this approach would also allow us to keep the state of the DAT gui completely outside of the model (which might be a good thing, the dependency should be `core model => DAT model`, never the other way around, and this would be a good way to enforce it).

What are your thoughts on this?

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -223,12 +243,35 @@ function initDataviz(channel) {
...
   Object.values(data.nodes).forEach(node => {
     node.neighbors = [];
     node.links = [];
+    if(node.tags && node.tags.length > 0){
+      node.tags.forEach(tag => {
+        const tagNode = {
+          id: tag.label,
+          title: tag.label,
+          type: 'tag',
+          properties: {color: d3.interpolateTurbo(Math.random())},
```
riccardoferretti, line 264: this should follow the stying pattern of all other nodes. In fact, I believe it's supported out of the box because we support styling nodes by type. see https://github.com/foambubble/foam/blob/master/docs/features/graph-visualisation.md

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -343,9 +407,9 @@ try {
     const message = event.data;
     switch (message.type) {
       case 'didUpdateGraphData':
-        const graphData = augmentGraphInfo(message.payload);
-        console.log('didUpdateGraphData', graphData);
-        Actions.refresh(graphData);
+        model.graphData = augmentGraphInfo(message.payload);
```
riccardoferretti, line 410: I don't think we should touch the `model` here. The model should only be touched by `Actions` and the `update` fn so we keep the unidirectional data flow - am I missing something?

riccardoferretti (commented):
Managed to get a deep review in, let me know if it's all clear, thanks!

packages/foam-vscode/static/dataviz/graph.js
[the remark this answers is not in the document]
dannysemi: Okay this sounds good. I mostly just work with data so before I started this I read up on MVC, but some things were confusing me. This definitely makes more sense to me. I thought we should have only one model.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -53,6 +53,7 @@ let model = {
   hoverNode: null,
   focusNodes: new Set(),
   focusLinks: new Set(),
+  graphData: {},
```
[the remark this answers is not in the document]
dannysemi: This may be the most challenging part for me. Naming things is way more difficult than I thought it would be.

packages/foam-vscode/static/dataviz/graph.js
dannysemi: It helps me to differentiate between tags to have them multicolored, but a uniform style is probably the more conventional approach. I originally used that approach, but then I found myself searching for the same tag over and over again. I have a much better approach to tag filtering sitting in a branch though, so once we have these initial changes worked out, I will improve the tag filtering.

packages/foam-vscode/static/dataviz/graph.js
dannysemi: No. I'm still working out MVC. I'll get it though.

packages/foam-vscode/static/dataviz/graph.js
dannysemi: Fixed.

packages/foam-vscode/static/dataviz/graph.js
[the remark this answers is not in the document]
dannysemi: I have this exactly working in a branch. I have to figure out this MVC stuff first and then I will move that over.

packages/foam-vscode/static/dataviz/graph.js
dannysemi: That makes more sense. I'm going to try again, but I don't know if I'm fully understanding how this should be structured just yet.

packages/foam-vscode/static/dataviz/graph.js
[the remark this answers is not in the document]
riccardoferretti: The pattern I am using for the graph is closer a mix of redux (https://redux.js.org/), SAM (https://sam.js.org/), and meiosis (https://meiosis.js.org/) - in general though this is a unidirectional data flow architecture (with some caveat due to the API of the graph library), as opposed to an MVC architecture.
Let me know if you have any question about this, happy to assist

packages/foam-vscode/static/dataviz/graph.js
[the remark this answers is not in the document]
riccardoferretti: You are not alone ;) https://www.mediawiki.org/wiki/Naming_things

packages/foam-vscode/static/dataviz/graph.js
riccardoferretti: that sounds good. I am also open to improvements around the styling of the graph, I just prefer a consistent approach, so that:
1. there is one way of going things for users
2. when there are updates, they can pushed across the domain

So yeah, let's start with our standard styling approach, and in another PR I am happy to consider changes to the approach

packages/foam-vscode/static/dataviz/graph.js
riccardoferretti: See my [comment above](https://github.com/foambubble/foam/pull/737#discussion_r692994457), let's not use MVC, let's use a unidirectional data flow pattern.

packages/foam-vscode/static/dataviz/graph.js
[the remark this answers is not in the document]
dannysemi: Great. I honestly didn't know where to start. I thought I was on the right track, but I'll check those out.

dannysemi:
Okay I think I addressed everything except what you were saying about the views and type filtering being restrictive. I wanted to make sure I had the data flow stuff down before going any further with the gui and filtering.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -93,40 +98,10 @@ function update(patch) {
...
-      // check that selected/hovered nodes are still valid (see #397)
-      m.hoverNode = m.nodeInfo[m.hoverNode] != null ? m.hoverNode : null;
-      m.selectedNodes = new Set(
-        Array.from(m.selectedNodes).filter(nId => m.nodeInfo[nId] != null)
-      );
-
-      // annoying we need to call this function, but I haven't found a good workaround
-      graph.graphData(m.data);
+      m.fullGraphData = graphInfo;
+      //patchGraphData(m,graphInfo);
```
riccardoferretti, line 104: I think by calling this function here you won't need to call `filterByType` in `didUpdateGraphData` - what was your thinking in this regard?

riccardoferretti (commented):
yes we are definitely heading in the right direction. I have not commented on `filterByType` as I understand it's still WIP.
I have a left some comments on the rest

packages/foam-vscode/static/dataviz/graph.js
dannysemi: I'm fine either way. I prefer not to show every node in the graph on load. So I was filtering by default.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -343,9 +412,10 @@ try {
     const message = event.data;
     switch (message.type) {
       case 'didUpdateGraphData':
-        const graphData = augmentGraphInfo(message.payload);
+        graphData = augmentGraphInfo(message.payload);
+        Actions.refreshWorkspaceData(graphData);
         console.log('didUpdateGraphData', graphData);
-        Actions.refresh(graphData);
+        Actions.filterByType();
```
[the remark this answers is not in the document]
dannysemi: Oh I started from the bottom on these, I just prefer filtering by default instead of showing every node in the graph on load. If that's not desirable then it's an easy fix.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -241,10 +274,46 @@ function augmentGraphInfo(data) {
...
+
+  // annoying we need to call this function, but I haven't found a good workaround
+  graph.graphData(m.data);
+}
+
 function getNodeColor(nodeId, model) {
   const info = model.nodeInfo[nodeId];
   const style = model.style;
-  const typeFill = d3.rgb(
+  const typeFill = info.properties.color ? d3.rgb(info.properties.color) : d3.rgb(
```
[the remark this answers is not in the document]
dannysemi: > does this mean that if a note has a `color` property it will show in that color within the graph?

Yes.

> also, for this operator we use the formatting:
> 
> ```
> const x = condition
>   ? when_true
>   : when_false
> ```

Am I not using that right? I thought that's what I was doing.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -164,6 +139,41 @@ const Actions = {
     };
     graph.backgroundColor(model.style.background);
   },
+  filterByType: () => {
```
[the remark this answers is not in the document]
dannysemi: Fixed.

StoltHD:
If you work on Cytoscape, why not talk with Emile that develope JUGGL?

He has cytoscape working for Obsidian, with a lot of config and customization already done...

riccardoferretti:
@dannysemi I think @StoltHD point is very valid, worth a try?

riccardoferretti:
(I should clarify, not for this PR, but for future ones)

dannysemi:
JUGGL looks great. Would you be looking to use that as the primary graph or as an extension?

riccardoferretti:
I have spoken to the creator of JUGGL, he doesn't have the bandwidth to port it to Foam, but he is open to helping out someone else integrate it.
In the meantime I am keen to merge this, as it's already a good step in the right direction

riccardoferretti (approved):
Thanks for this, I like the extra data sent to the graph, as well as the DAT setup, which will help us down the line.

There might be a couple of adjustments for future PRs, but will merge this in the meantime as it's already a great step forward!

krisztian-hofstadter-tedor:
dear @riccardoferretti  and @dannysemi 

will what you are working on in this issue be able to create a graph with tags instead of wiklinks?

that would be great.

I am asking as I can put tags e.g. #tag1 and #tag in HTML comments e.g.  and still find them in the tag explorer window. However, when I put wikilinks e.g. [[forest]] in a comment e.g.  this will not be available as a reference/backlink anymore at end of the .md file nor available for the graph visualiser. 

thanks, k

PR #741, opened 2021-08-11
Proposal and basic feature to include notes into a note

pderaaij:
Evolving from an initial attempt to include a note, this PR has been matured into a proposal document. The proposal displays the possible way forward with inclusion of notes. The initial code is still part of the PR, to be seen as a little proof of concept. This will be removed before merging the proposal into the repository.

The [rendered version](https://github.com/pderaaij/foam/blob/feature/include-bubbles/docs/proposals/inclusion-of-notes.md) allows for better reading of the proposals. Any comments, ideas and feedback is welcomed on this PR.

**Proof of Concept**
Initial work on allowing notes to be included in notes via the syntax of `![[wikilink]]`. This proof of concept works and prevents cyclic inclusion. This was a short amount of time spent on it. Requires additional work to implement, but the feature seems to be relatively easy to develop.

Todo:
- [x] Add tests
- [x] Think about styling opportunities, e.g. allow for `<div>` around each inclusion with an index number to style inclusions
- [x] Add documentation on usage and limitations


riccardoferretti:
The approach looks solid, do you have a screenshot of the result? let me know when I can play with it too.

Will this support autocompletion as well?

docs/proposals/inclusion-of-notes.md
```diff
@@ -0,0 +1,92 @@
...
+
+As a user I could be interested in collecting the value of any given proeprty for a note. For example, I might want to include the tags as defined in the frontmatter of note A. This should be possible via the syntax:
+
+`![[wikilink:<property>]]`
+
+The property value should be lookedup by foam defined properties, e.g. title, **or** any property defined in the frontmatter of a note.
+
+So, the example of including the tags of a note should be:
+
+`![[wikilink:tags]]`
```
riccardoferretti, line 50: This is interesting, I don't have this case in my knowledge base, but I can see the utility.

In terms of the syntax, I personally like `:` (to be confirmed based also on compatibility with other md based note taking tool we'd want to play nicely with).

I think in this context the string:

```
![[Daily Routine:tags]]
```

Would produce a box with the title `tags property of Daily Routine` and the value as content?

docs/proposals/inclusion-of-notes.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Inclusion of backlinks overview
+
+It could be interesting to easily display the backlinks of a note in the preview, rather then using the custom window foam provides. If an user wants to easily see, or perhaps format, the references of that note this could be an option.
+
+A syntax could look like:
+
+`!{{note-backlinks}}`
```
riccardoferretti, line 58: This is trickier because it requires access to the graph data to compute the value.

Is this the rationale for changing the syntax here?

docs/proposals/inclusion-of-notes.md
```diff
@@ -0,0 +1,92 @@
...
+
+But, syntax is something to think about.
+
+### Introducing a query language
```
riccardoferretti, line 62: I wonder whether the query language per se should be treated as a different feature altogether, because although it's an expansion of embedding, it has quite a different scope and goal.
I am open to it, but wouldn't treat it in the same way as an embed. Thoughts?

pderaaij: I agree, I was on the fence here and went for: let's document it anyway ;-)

I'll extract it in a separate proposal.

docs/proposals/inclusion-of-notes.md
pderaaij: I wanted to prevent clashing with a possible wikilink called `[[note-backlinks]]`. But, agree, it's a bit weird that the syntax changes. We could also treat backlinks as a property of the note. As such, you could have the syntax:

`[[wikilink:backlinks]]` 

That might simplify the implementation as well.

docs/proposals/inclusion-of-notes.md
pderaaij: Yes, I can imagine the default being something like:

For multiple results (e.g. tags)
```
<div>
    <div> <span>[property]</span> of <span>[note title]</span><div>
   <div>
   [values] (displayed as unordered list)
</div>
</div>
```

For single results as title or count it could be:

```
<div> <span>[property]</span> of <span>[note title]</span> <span>[value]</span><div>
```

Rough version, of course, but at least offer flexibility to users to custom style the resulting field as they'd like. Giving optionality to hide elements for example.

docs/proposals/inclusion-of-notes.md
```diff
@@ -0,0 +1,92 @@
...
+- Adds minimal amount of knowledge required as syntax is based on the syntax of creating a wikilink.
+- Makes the auto-complete work ouf-of-the-box, without any additional code and listeners required.
+
+**Important**. A risk exists that a loop of including the same notes arises. E.g. Note A includes note B which includes note A. This needs to be prevented by the implementation and made visible to the user.
+
+### Include a section of a note
+
+It could be interesting to only include a section of a note instead of the entire note. In order to do so thse user should be able to use the following syntax:
+
+`![[wikilink#section-b]]`
```
kevinslin, line 36: So from observed usage in Dendron, I would say that this would be the desirable default behavior. We introduced the syntax of header ranges and wildcard block headers to facilitate this effect: https://wiki.dendron.so/notes/f1af56bb-db27-47ae-8406-61a98de6c78c.html#wildcard-headers-as-a-range-boundary

Something else you might want to consider is removing the header in a reference. It's usually not needed.  We also have a concept of an `offset` which moves the reference by `offset` lines down. This all comes together in the following link, which roughly reads - include `section-b`, skip the first line, go to next block

```
![[wikilink#section-b,1:#*]]
```

This is the most commonly used sort of reference in Dendron.  Moving forward, we'll probably set an option to make this default behavior:
```
 `![[wikilink#section-b]]` is alias for  `![[wikilink#section-b,1:#*]]`
```

docs/proposals/inclusion-of-notes.md
```diff
@@ -0,0 +1,92 @@
...
+
+As a result it will include the section title + section content until the next section *or* end of file.
+
+### Include an attribute of a file (note property or frontmatter)
+
+As a user I could be interested in collecting the value of any given proeprty for a note. For example, I might want to include the tags as defined in the frontmatter of note A. This should be possible via the syntax:
+
+`![[wikilink:<property>]]`
```
kevinslin, line 44: This is a cool proposal. We've been thinking about something similar but are still undecided on the syntax

docs/proposals/inclusion-of-notes.md
kevinslin: We have had a few issues from users requesting this feature. We currently use `{{ ... }}` syntax for [variable substitution](https://wiki.dendron.so/notes/ba97866b-889f-4ac6-86e7-bb2d97f6e376.html#frontmatter-variable-substitution) since its the convention also adopted by many templating languages (eg. 11ty, jekyll, etc)

Currently we're thinking about having `{{ backlinks }}` be a reserved frontmatter variable that links to all current backlinks of a note

docs/proposals/inclusion-of-notes.md
kevinslin: we're also writing up our query syntax right now :)
that should be out next week so might be good to compare. we're looking for an alternative syntax from `{{ ... }}` because we use it for variable substitution

docs/proposals/inclusion-of-notes.md
pderaaij: That could be in an interesting solution. I'll put this in a separate proposal, including various options like the ones already mentioned.

docs/proposals/inclusion-of-notes.md
pderaaij: Good point! Can imagine, the section header preferably is left out. I would propose to take that as a default and enable offsetting at a later stage. 

@riccardoferretti what do you think?

pderaaij:
@riccardoferretti I was on holiday so chose not to work on it ;-)

I've simplified the proposal and will work on the tests today and tomorrow. I'll add a new PR with a proposal for the backlinks and query language.

docs/proposals/inclusion-of-notes.md
riccardoferretti: yup I agree!

docs/proposals/inclusion-of-notes.md
riccardoferretti: this is interesting and I still like the syntax, I am only wondering how we distinguish a title with the colon (`:`) char in it, vs the `:` used as a separator for the property name.
I don't like the idea of having to escape the `:` (kinda like what one would need to do for `|`, which is used for aliases).
Ideas:
- use the `|` itself, which people already expect to escape, but that overloads the semantic of the symbol and could create confusion
- use something we would likely not see in a filepath, e.g. `>>` (`![[Daily Routine >> tags]]`). If a filepath includes it, they would have to escape the sequence (`![[Daily \>\> Routine >> tags]]` would get the `tags` property of the `Daily >> Routine` note)
- something else?

wdyt?

packages/foam-vscode/src/features/preview-navigation.spec.ts
```diff
@@ -50,3 +51,55 @@ describe('Stylable tag generation in preview', () => {
...
+    ).toMatch(
+      `<p>This is the root node.
+![[note-b]]</p>
+`
+    );
+  });
+
+  it('displays a warning in case of cyclical inclusions', () => {
+    expect(md.render(noteD.source.text)).toMatch(
+      `<p>This is the text of note D which includes <p>This is the text of note C which includes <p>This is the text of note D which includes <div class=\"foam-cyclic-link-warning\">Cyclic link detected for wikilink: note-c</div></p>
```
riccardoferretti, line 99: let's make the linter happy :)

riccardoferretti: also, shouldn't the cycle be detected when `note C` imports `note D`?

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -25,6 +27,47 @@ const feature: FoamFeature = {
   },
 };
 
+export const markdownItWithNoteInclusion = (
+  md: markdownit,
+  workspace: FoamWorkspace
+) => {
+  return md.use(markdownItRegex, {
+    name: 'include-notes',
+    regex: /!\[\[([^\[\]]+?)\]\]/,
```
riccardoferretti, line 36: let's make the linter happy, either by removing the escape char if really unnecessary, or by placing an es-lint directive to ignore the line

riccardoferretti (approved):
Looks really good.
I think it can already be merged, but I wanted to double check with you the cycle detection comment I left first

docs/proposals/inclusion-of-notes.md
pderaaij: Perhaps we make life easier if we would change the syntax to something like:

`(![[Daily Routine]] >> tags)`

This way we won't interfere with the Wikilinks, we just make it explicit that you include a note and perform some extraction. The addition of the parentheses is to make life of RegEx-ing this a bit easier. But other combinations are also possible, of course.

packages/foam-vscode/src/features/preview-navigation.spec.ts
pderaaij: Interesting point. I now chose the same behaviour as vscode-memo. We could change it so that a note won't be included if it contains a reference to the same page. Not sure what the best option is in terms of user experience. Personally, I like it to easily see where a cyclic reference is made, instead of looking into the file. But, I do not hold a strong opinion on this.

riccardoferretti: mmm.. maybe I wasn't clear or didn't understand the reply, what I meant is that right the cycle detection will include the source note.

`renders Note D -> renders Note C -> renders Note D -> cycle warning`

But the cycle actually happens a step ahead, so I was expecting something like:

`renders Note D -> renders Note C -> cycle warning`

does that makes sense? wdyt?

PR #774, opened 2021-10-01
Consolidate foam-core within foam-vscode

riccardoferretti:


pderaaij:
This is wonderful work. I'm still working my way through the review. But, was thinking in the meanwhile:

Might be interesting to add a custom rule to prevent `vscode` imports into the `core` folder? Maintaining a one-way dependency relation.

pderaaij (approved):

movermeyer:
> Might be interesting to add a custom rule to prevent `vscode` imports into the `core` folder? Maintaining a one-way dependency relation.

IDK what the equivalent is in Typescript, but Shopify does this with a tool called [Packwerk](https://github.com/Shopify/packwerk)

pderaaij:
Yeah, I think we can achieve it by configuring an eslint rule. For example:
https://eslint.org/docs/rules/no-restricted-imports

riccardoferretti:
Thanks for the ES lint suggestion! 
So, I looked into it, and man that was a rabbit hole!
In the end I managed to get the core module to not depend on VS Code (half the game), but tried in various ways to avoid the dependency on other modules and none seemed to work.
I left some comments in the lint configuration file, but I don't consider this a blocker for the PR to be merged, so I will just go ahead. If you are more familiar with eslint please/feel free to take a look, maybe I am missing something silly!

PR #775, opened 2021-10-01
Add some basic details about multi-root workspaces

Laptop765:


Laptop765 (commented):
It would probably be better to specify the `some settings` but I couldn't tell why some needed to be moved and others didn't.

movermeyer:
> It would probably be better to specify the `some settings` but I couldn't tell why some needed to be moved and others didn't.

Yeah... how Foam uses workspaces is not consistent (yet) 😢. 
It's a bug I hope to address as part of https://github.com/foambubble/foam/pull/679 (once we know how we're handling https://github.com/foambubble/foam/pull/683) (Soon?🤞)

PR #778, opened 2021-10-05
Allow custom theming of Foam Links

ecool:
Allow custom theming of Foam Links.

This is just a fix so that while using `workbench.colorCustomizations` the default VSCode link/editor warning colors are not also modified by using:

```json
"workbench.colorCustomizations": {
    "foam.link.found": "#88e8ffcc",
    "foam.link.placeholder": "#ffff9ccc",
},
```

instead of:
```json
"workbench.colorCustomizations": {
    "textLink.foreground": "#88e8ffcc",
    "editorWarning.foreground": "#ffff9ccc",
},
```

### Format
`foam.{{item}}.{{type}}`

This format would allow for easy extensibility in the future.

#### Example
`foam.link.found`
`foam.link.placeholder`

### Further Questions / Discussion
I'm unsure if the mouse icon should also be changed. Changing to a mouse `pointer` icon when hovering and not holding `ctrl` generates the feeling that you should be able to click on the link and have something happen.

Currently, this doesn't do anything. You must hold `ctrl` and then `click`. This is a good system to stop accidental link clicking and I fully agree with the `ctrl + click` functionality.

However as I mentioned above, the mouse `pointer` icon gives the user the feeling that clicking should accomplish an action.

This feels like a UX/UI discussion and I have very little knowledge of it. Anyone else that has any other concerns, please let me know.

### Future Actions/Possibilities
I'm also currently looking into creating a [Semantic Highlighting Provider](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide) for actually being able to use a Theme that also has Foam in mind.

This will be the first time I've tried to do a provider for semantic highlighting but looking through the example and documentation it doesn't seem too difficult to tie into Foam.

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -355,6 +355,26 @@
         "command": "foam-vscode.open-daily-note",
         "key": "alt+d"
       }
+    ],
+    "colors": [
+      {
+        "id": "foam.link.found",
+        "description": "Color decoration for links.",
+        "defaults": {
+          "dark": "textLink.foreground",
+          "light": "textLink.foreground",
+          "highContrast": "textLink.foreground"
+        }
+      },
+      {
+        "id": "foam.link.placeholder",
+        "description": "Color decoration for placeholder links.",
+        "defaults": {
+          "dark": "editorWarning.foreground",
+          "light": "editorWarning.foreground",
+          "highContrast": "editorWarning.foreground"
+        }
+      }
     ]
   },
   "scripts": {
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -358,7 +358,7 @@
     ],
     "colors": [
       {
-        "id": "foam.link.found",
+        "id": "foam.link.resource",
         "description": "Color decoration for links.",
         "defaults": {
           "dark": "textLink.foreground",
--- a/packages/foam-vscode/src/features/document-decorator.ts
+++ b/packages/foam-vscode/src/features/document-decorator.ts
@@ -12,14 +12,14 @@ export const CONFIG_KEY = 'decorations.links.enable';
 const linkDecoration = vscode.window.createTextEditorDecorationType({
   rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
   textDecoration: 'none',
-  color: { id: 'textLink.foreground' },
+  color: { id: 'foam.link.found' },
   cursor: 'pointer',
 });
 
 const placeholderDecoration = vscode.window.createTextEditorDecorationType({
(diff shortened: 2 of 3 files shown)


packages/foam-vscode/package.json
```diff
@@ -355,6 +355,26 @@
         "command": "foam-vscode.open-daily-note",
         "key": "alt+d"
       }
+    ],
+    "colors": [
+      {
+        "id": "foam.link.found",
```
riccardoferretti, line 361:
```suggestion
        "id": "foam.link.resource",
```

packages/foam-vscode/src/features/document-decorator.ts
```diff
@@ -12,14 +12,14 @@ export const CONFIG_KEY = 'decorations.links.enable';
 const linkDecoration = vscode.window.createTextEditorDecorationType({
   rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
   textDecoration: 'none',
-  color: { id: 'textLink.foreground' },
+  color: { id: 'foam.link.found' },
```
riccardoferretti, line 15:
```suggestion
  color: { id: 'foam.link.resource' },
```

riccardoferretti (commented):
I find this approach really good, thanks for your contribution!

I suggest we use the "resource" naming, as it's in line with the code and I feel it reads better.
I like the `foam.{{item}}.{{type}}` approach, I was playing it out with a few others that might be interesting in the future, such as:
- `foam.tag`
- `foam.mention` (this concept doesn't exist in Foam yet but I am thinking about it)
- `foam.link.{{type-of-note}}` (e.g. if we want to customize style depending on the YAML `type` property of a note, which is a concept we already have at the model level)

A couple of other things to think about:

1. how about styling for the dataviz graph?
2. how about styling for the preview?

Would be interesting to see how these would play out (e.g. knowing whether they would fall in the same scheme, or not), to help define the configuration structure here (e.g. `foam.editor.link.placeholder` vs `foam.preview.link.placeholder`?)

What are your thoughts?

ecool:
Committed the changes.

The naming convention `foam.link.resource` reads much better than `foam.link.found`.
Did a quick look at the DataViz feature, this morning, and I believe it is possible to use custom colors based on `foam.{{item}}.{{type}}`.

For both the DataViz and Preview features I'll be taking a closer look throughout the day today.

I really like the included ideas for other types of items: `tag`, `mention`, YAML `type`. 

Follow up later today.

memeplex:
With https://github.com/foambubble/foam/pull/876 styling "found" links will be possible using standard `editor.tokenColorCustomizations`.

riccardoferretti:
@ecool I am closing this as I believe it's superseded by #876 - if there is something I am missing please let me know

Thanks for inspiring the improvements around highlights and decorations!

PR #781, opened 2021-10-09
Added support for target date variables in daily note template

### LINKED ISSUE #776: Correct date formatting in daily-notes custom template
- Foam version: 0.14.2
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes | No

**Summary**
When using a custom template for daily-notes (`daily-notes.md`), it is not possible to format to the correct date in e.g. the header of the file. The only way known to me to format the date is using VS Code Snippets e.g.
``$CURRENT_YEAR - $CURRENT_MONTH - $CURRENT_DATE``.
But this only works for the current day.
If the daily-notes for another day are opened by using the provided snippets e.g. `/tomorrow` it does not match anymore. The date in the template is always formatted to the date of creation and not the date which corresponds to the date of the note.

**Steps to reproduce**
1. Create custom daily-notes.md template using date formatting e.g. `$CURRENT_DATE`
2. Open daily-note for another date than today e.g. `/tomorrow`
3. Date formatting does not match

riccardoferretti:
fixes #776 

We are adding new variables to the Foam daily note template.
These variables reflect the date of the note being created, so that a custom template can use them.

### DIFF
--- a/packages/foam-vscode/CHANGELOG.md
+++ b/packages/foam-vscode/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the "foam-vscode" extension will be documented in this fi
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
+## Unreleased
+
+- Added `FOAM_DATE_*` template variables (#781)
+
 ## [0.15.1] - 2021-10-21
 
 Fixes and Improvements:
@@ -25,7 +29,7 @@ Fixes and Improvements:
 
 ## [0.14.2] - 2021-07-24
 
-Features: 
+Features:
 
 - Autocompletion for tags (#708 - thanks @pderaaij)
 - Use templates for new note created from wikilink (#712 - thanks @movermeyer)
@@ -42,7 +46,7 @@ Fixes and Improvements:
 
 ## [0.14.0] - 2021-07-13
 
-Features: 
+Features:
 
 - Create new note from selection (#666 - thanks @pderaaij)
 - Use templates for daily notes (#700 - thanks @movermeyer)
@@ -70,9 +74,9 @@ Fixes and Improvements:
 
 - Fixed #667, incorrect resolution of foam-core library
 
-Internal: 
+Internal:
 
-- BREAKING CHANGE: Removed Foam local plugins 
+- BREAKING CHANGE: Removed Foam local plugins
   If you were previously using the alpha feature of Foam local plugins you will soon be able to migrate the functionality to the V1 API
 
 ## [0.13.6] - 2021-06-05


movermeyer:
I think we're actually very similar in our thinking, but I didn't put enough time into writing it all out. 😅

I think there are two separate topics to discuss:

1. Simplicity (i.e. Do we need daily-note-specific variables?)
2. Expressivity (i.e. Should we copy VSCode's date variables interface?)

----

## Simplicity

My gut instinct when I saw the issue was to create a Foam-specific variable that could replace the use of `CURRENT_*` in templates.
i.e., the new variable would be equivalent to `CURRENT_*`, but always populated according to the appropriate frame of reference.

In this PR, the daily note variables are `FOAM_DATE_* = <target_date> || null`, so they are only useful in the context of daily notes
What I had in mind is: `FOAM_DATE_* = <target_date> ?? <current_date>`. It could be useful in any context (daily note templates, but also regular ones)

**Benefits**

* **Your daily notes don't have any magic in them:** Daily and non-daily notes look and behave the exact same way
* **No context-specific variables:** IMO snippet variables should be useful in all contexts (within reason) in order to justify their existence
  * They are an API that we have to support forever. The fewer the better; IMO we should take the time to ensure we can't the interface simpler.
* 

**Detriments**

* "Perhaps it'll be too complicated to explain the frame of reference?"
  * I don't think we have to explain a new concept / I don't think users will ever be surprised. I think we can state in the documentation "`FOAM_DATE_*` variables are how you insert dates in templates" and it'll just work the way they expect. We might explain the difference between `FOAM_DATE_*` and VSCode's `CURRENT_*` in a "pro-tip" box or whatever, but the basic documentation is very simple

----

## Expressivity

### Relation to [#564](https://github.com/foambubble/foam/issues/564)

Since this involves customizing date formats in daily notes, and you will also want to use the new variables in filepaths, it is very adjacent to [this issue](https://github.com/foambubble/foam/issues/564). I think they can be solved at the same time? We were hoping for a solution more flexible than what VSCode snippet variables can offer us (and there are similar issues in the [upstream](https://github.com/microsoft/vscode/issues?q=is%3Aissue+label%3Asnippets) complaining about the limitations of the existing date variables), so I'm not convinced that copying their variables is a good idea. As discussed briefly on that issue, something similar to `$FOAM_DATE_FORMAT` (bad name, I know) might be a good candidate.

Example `daily-note.md` template:

```markdown
---
foam_template:
  filepath: 'journal/${FOAM_DATE_FORMAT}.md'
---
# ${FOAM_DATE_FORMAT}

## Notes
```

----

### Question: What if a user actually wants the current date in their "relative daily note"?

(**The current PR also works in this way.** I just wanted to comment here since I had asked myself this question.)

They might want a "created on" in their notes.

Desired note contents:

```markdown
# 2021-12-13 <- Some date in the future

# $FOAM_TITLE

Created on 2021-10-13 <- Today's date
```

**Answer:** In this case, they can make use of the existing `CURRENT_*` variables.

```markdown
# $FOAM_DATE_FORMAT <- Some date in the future

# $FOAM_TITLE

Created on $CURRENT_YEAR-$CURRENT_MONTH-$CURRENT_DATE <- Today's date
```

**The current PR also works in this way.** I just wanted to comment here since I had asked myself this question.

riccardoferretti:
I agree with everything you wrote, happy to go in that direction, except on the topic of `FOAM_DATE_FORMAT`, which feels like an add-on to this conversation (the value of which I am still questioning), so I would bracket it out from this PR

PR #808, opened 2021-10-29
Allow inclusion of note when using reference definitions

### LINKED ISSUE #794: Link References break Preview for wikilinks
- Foam version: v0.15.1
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: No

**Summary**

**Steps to reproduce**
1. In a.md, embed b.md using syntax `![[b]]` and `[[c]]`
2. Open the preview
3. Without saving the file (so the Link references won't be automatically generated), the preview works fine: 
Screen Shot 2021-10-21 at 15 50 10
4. After saving the file, the Link references are generated. The preview is broken. Expected to see b.md embedded: 
Screen Shot 2021-10-21 at 15 49 48
5. (minor) the preview for syntax like `[[c]]` also broke: there's an extra pair of bracket

pderaaij:
Fixes #794.

The idea is that when we want to include a note that is also referred in the link definition, we can just "remove" that definition. This prevents the tokenisation to distort the inclusion format.

### DIFF
--- a/packages/foam-vscode/src/features/preview-navigation.ts
+++ b/packages/foam-vscode/src/features/preview-navigation.ts
@@ -144,15 +144,19 @@ export const markdownItWithRemoveLinkReferences = (
   md: markdownit,
   workspace: FoamWorkspace
 ) => {
-  // Forget about reference links that contain an alias divider
   md.inline.ruler.before('link', 'clear-references', state => {
     if (state.env.references) {
       Object.keys(state.env.references).forEach(refKey => {
+        // Forget about reference links that contain an alias divider
+        // Aliased reference links will lead the MarkdownParser to include wrong link references
         if (refKey.includes(ALIAS_DIVIDER_CHAR)) {
           delete state.env.references[refKey];
         }
 
-        // if we try to include this note, we can remove the reference as well
+        // When the reference is present due to an inclusion of that note, we
+        // need to remove that reference. This ensures the MarkdownIt parser
+        // will not replace the wikilink syntax with an <a href> link and as a result
+        // break our inclusion logic.
         if (state.src.toLowerCase().includes(`![[${refKey.toLowerCase()}]]`)) {
           delete state.env.references[refKey];
         }
--- a/packages/foam-vscode/src/features/preview-navigation.ts
+++ b/packages/foam-vscode/src/features/preview-navigation.ts
@@ -151,6 +151,11 @@ export const markdownItWithRemoveLinkReferences = (
         if (refKey.includes(ALIAS_DIVIDER_CHAR)) {
           delete state.env.references[refKey];
         }
+
+        // if we try to include this note, we can remove the reference as well
+        if (state.src.toLowerCase().includes(`![[${refKey.toLowerCase()}]]`)) {
+          delete state.env.references[refKey];
+        }
       });
     }
     return false;


riccardoferretti:
stupid question: why don't we just remove all (generated) reference definitions from the preview? do they serve a purpose?

pderaaij:
That would lead to the problem for the explicitly defined link references. At this level in the MarkdownIt parser, we can't differentiate between implicit and explicitly generated definitions.

riccardoferretti:
IIUC:
- the goal of including link definitions is so that in preview links can be followed
  - we don't need that for our links (which are source of generated definitions) because we resolve the address in the link itself
  - but the user could have extra definitions, which are needed to resolve the link
- given that in the preview external (non generated) link definitions are useful, we cannot just remove all link definitions
- it would be ok to remove all our generated link definitions, but because we can't tell them apart from the external ones, we can't simply do that - instead we can remove the ones that cause us trouble (which we always assume to be internal?):
  - wikilinks with aliases
  - "wikilinks" for included notes

Does it even make sense to have a definition for an embedded note? there is no link to follow there
Maybe the solution here is to not create definitions for included notes?

pderaaij:
> 
>     * the goal of including link definitions is so that in preview links can be followed

In the context of retaining the definitions for preview, this is correct. In the broader sense, the definitions are there for publication.

>       
>       * we don't need that for our links (which are source of generated definitions) because we resolve the address in the link itself
>       * but the user could have extra definitions, which are needed to resolve the link
> 
>     * given that in the preview external (non generated) link definitions are useful, we cannot just remove all link definitions
> 
>     * it would be ok to remove all our generated link definitions, but because we can't tell them apart from the external ones, we can't simply do that - instead we can remove the ones that cause us trouble (which we always assume to be internal?):
>       
>       * wikilinks with aliases
>       * "wikilinks" for included notes
> 

All above, correct.

> Does it even make sense to have a definition for an embedded note? there is no link to follow there Maybe the solution here is to not create definitions for included notes?

We could create some magic to leave it out there. But, that makes life harder for publishing. Not sure what the best UX is here. I could imagine that even embedding does not work when publishing, I at least can manually follow the link because the definition is there.

riccardoferretti:
> We could create some magic to leave it out there. But, that makes life harder for publishing. Not sure what the best UX is here. I could imagine that even embedding does not work when publishing, I at least can manually follow the link because the definition is there.

Yeah I think we are aligned.
I agree that embedding is outside of our control (it's the publishing fwk's job), but it's fair to keep the link as a sensible fallback.

Ok, let's go forward with your original proposal, will comment in code

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -151,6 +151,11 @@ export const markdownItWithRemoveLinkReferences = (
         if (refKey.includes(ALIAS_DIVIDER_CHAR)) {
           delete state.env.references[refKey];
         }
+
+        // if we try to include this note, we can remove the reference as well
```
riccardoferretti, line 155: let's elaborate a bit more (possibly with a reference to this PR) the reasoning about this logic.
let's put ourselves in the shoes of other developers and future selves who are trying to understand what we wanted to do, and why, here

packages/foam-vscode/src/features/preview-navigation.ts
```diff
@@ -151,6 +151,11 @@ export const markdownItWithRemoveLinkReferences = (
         if (refKey.includes(ALIAS_DIVIDER_CHAR)) {
```
riccardoferretti, line 151: the comment from above "Forget about..." should probably be here?
Also, possibly explain why we are doing that

riccardoferretti (commented):
The approach sounds good to me, I think it would be good to provide better in-code documentation so that our future selves can understand why we went this route

riccardoferretti (approved):

riccardoferretti:
Looks great, thx!

Issue #813, opened 2021-11-03
Duplicating Inbox node in graph


AndreiD049:
- Foam version:  v0.15.2
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes 

**Summary**

In a clean copy of foam-template, whenever i just make a change in inbox.md and then save the file. the node is duplicated in the graph:
![image](https://user-images.githubusercontent.com/52671223/140170771-d043dc4f-981f-47b1-8296-43a8ceceab6c.png)

I don't see this issue for other nodes, but i believe it's linked to the inconsistency described here microsoft/vscode#116298

**Steps to reproduce**
Not sure if this will happen 100% on other machine, but the steps are simple. Just save the file -> check the graph for duplicates.

**Additional information**
When debugging the project, after initialization, those are my resources:
![image](https://user-images.githubusercontent.com/52671223/140165617-3c5cd149-f58f-4cfc-b825-5a8466f87676.png)

After i save `inbox.md` (notice doubled inbox.md, one with uppercase drive letter):
![image](https://user-images.githubusercontent.com/52671223/140165796-11b8ae29-64dc-4bb6-8c22-22481116972b.png)

PR #814, opened 2021-11-03
Replace vscode Uri with foam URI

AndreiD049:
Fixes: #813

While debugging above mentioned issue i noticed that in both of below places, an vscode `Uri` is used, while by the function definition a Foam `URI` is required.
1. https://github.com/foambubble/foam/blob/fd9bf8f04fa2cef2e3dc133cf07b3de8ff79084e/packages/foam-vscode/src/features/wikilink-reference-generation.ts#L75
2. https://github.com/foambubble/foam/blob/fd9bf8f04fa2cef2e3dc133cf07b3de8ff79084e/packages/foam-vscode/src/features/wikilink-reference-generation.ts#L141
Typescript doesn't complaint about that, since they have similar interfaces, but the inconsistency that is mentioned [here](https://github.com/microsoft/vscode/issues/116298) leads to the above issue.

In case any additional information, changes needed, please let me know.

### DIFF
--- a/packages/foam-vscode/src/features/wikilink-reference-generation.ts
+++ b/packages/foam-vscode/src/features/wikilink-reference-generation.ts
@@ -35,6 +35,7 @@ import {
   LINK_REFERENCE_DEFINITION_FOOTER,
   LINK_REFERENCE_DEFINITION_HEADER,
 } from '../core/janitor';
+import { fromVsCodeUri } from '../utils/vsc-utils';
 
 const feature: FoamFeature = {
   activate: async (context: ExtensionContext, foamPromise: Promise<Foam>) => {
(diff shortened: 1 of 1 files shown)


riccardoferretti (approved):
Very good find @AndreiD049

I think your current solution is correct, I am a bit concerned about the relationship between `URI` and `Uri` now feels a bit brittle (given the VS Code bug you referenced), but for now I can't think of a better solution.
Maybe we need to force uris being sent to Foam to be `URI` instead of `Uri`, I am wondering what would be a good solution to add this safety check without "compromising" the `URI` type.

In the meantime we'll merge this fix

riccardoferretti:
merging despite the failures b/c this CI error on Mac occurs on `master` as well

riccardoferretti:
@allcontributors add @AndreiD049 for code

PR #816, opened 2021-11-04
Update VS Code engine and test instance version

riccardoferretti:
Use the latest version of VS Code for both API and testing.

Fix all the tests that are failing because of the upgrade.

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -12,7 +12,7 @@
   "license": "MIT",
   "publisher": "foam",
   "engines": {
-    "vscode": "^1.47.1"
+    "vscode": "^1.61.0"
   },
   "icon": "icon/FOAM_ICON_256.png",
   "categories": [
@@ -390,7 +390,7 @@
     "@types/node": "^13.11.0",
     "@types/picomatch": "^2.2.1",
     "@types/remove-markdown": "^0.1.1",
-    "@types/vscode": "^1.47.1",
+    "@types/vscode": "^1.61.0",
     "@typescript-eslint/eslint-plugin": "^2.30.0",
     "@typescript-eslint/parser": "^2.30.0",
     "babel-jest": "^26.2.2",
@@ -406,7 +406,7 @@
     "tsdx": "^0.13.2",
     "tslib": "^2.0.0",
     "typescript": "^3.9.5",
-    "vscode-test": "^1.3.0"
+    "vscode-test": "^1.6.2"
   },
   "dependencies": {
     "dateformat": "^3.0.3",
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -406,7 +406,7 @@
     "tsdx": "^0.13.2",
     "tslib": "^2.0.0",
     "typescript": "^3.9.5",
-    "vscode-test": "^1.6.2"
+    "vscode-test": "^1.6.1"
   },
   "dependencies": {
     "dateformat": "^3.0.3",
--- a/packages/foam-vscode/src/features/create-from-template.spec.ts
+++ b/packages/foam-vscode/src/features/create-from-template.spec.ts
@@ -28,7 +28,7 @@ describe('createFromTemplate', () => {
       jest.clearAllMocks();
     });
 
-    it('can be cancelled while resolving FOAM_TITLE', async () => {
+    it.skip('can be cancelled while resolving FOAM_TITLE', async () => {
       const spy = jest
         .spyOn(window, 'showInputBox')
         .mockImplementation(jest.fn(() => Promise.resolve(undefined)));
--- a/packages/foam-vscode/src/test/run-tests.ts
+++ b/packages/foam-vscode/src/test/run-tests.ts
@@ -46,12 +46,6 @@ async function main() {
           '--disable-extensions',
           '--disable-workspace-trust',
         ],
-        // Running the tests with vscode 1.53.0 is causing issues in the output/error stream management,
-        // which is causing a stack overflow, possibly due to a recursive callback.
-        // Also see https://github.com/foambubble/foam/pull/479#issuecomment-774167127
-        // Forcing the version to 1.52.0 solves the problem.
-        // TODO: to review, further investigate, and roll back this workaround.
-        version: '1.52.0',
       });
     } catch (err) {
       console.log('Error occurred while running Foam e2e tests:', err);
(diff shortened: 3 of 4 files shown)


riccardoferretti:
in reference to https://github.com/foambubble/foam/pull/814
@AndreiD049, I have been taking a closer look at this as part of #816 because the update surfaced a bunch of URI related issues that are exactly due to the casing of the drive letter.

I am considering normalizing all `URI` to lower case drive, which I don't love because I think that:
- URI => Uri => URI should be idempotent
- Uri => URI => Uri should be idempotent

But it sounds like that given not even VS Code can be consistent on the drive letter casing, maybe we should just pick one convention (lowercase) and go with it.

I don't know exactly whether this is problematic on a Windows machine (I am on OSX). I assume you have one, can you please test this branch locally? (be aware that some tests are still failing, I am a bit surprised but it's not easy to debug Windows-only issues on a Mac 😅 - if you have an idea of what's going on please let me know)

AndreiD049:
@riccardoferretti i will test this tomorrow and let you know ;)

AndreiD049:
@riccardoferretti i am surprised but it's neither easy to debug those issue while you're on windows 🤣 
When i clone your branch and run `npm run e2e`, i get the same failing tests.
Then, when i want to debug them, i create a task in launch.json like below:
```json
  "configurations": [
   ...
    {
      "name": "Extension Tests",
      "type": "extensionHost",
      "request": "launch",
      "runtimeExecutable": "${execPath}",
      "args": [
        "C:\\Users\\AndreiD\\AppData\\Local\\Temp\\foam-C4zCFm",
        "--disable-extensions",
        "--disable-workspace-trust",
        "--extensionDevelopmentPath=${workspaceFolder}/packages/foam-vscode",
        "--extensionTestsPath=${workspaceFolder}/packages/foam-vscode/out/test/suite"
      ],
      "outFiles": ["${workspaceFolder}/packages/foam-vscode/out/test/**/*.js"]
    }
```

And then they all run successfully.
image

I thought it might be linked to the version of vscode used when using `npm run e2e` (it's 32bit)
I downloaded the 32bit vscode, changed my env variables but still same result.

riccardoferretti:
thanks for trying, I can't believe this! haaha

have you tried using the `Debug Jest Tests` task defined in Foam? that should work

AndreiD049:
I've tried it, the tests fail, but i can break only on unit tests. On integration tests, the breakpoint becomes "unbound":
image
Does it work for you?

riccardoferretti:
yes it works for me... that's odd..

image

AndreiD049:
tried another windows machine, same behavior. I cannot place a breakpoint in e2e tests.
I can do it only if i run the tests via above mentioned task in `launch.json`, but then, all passes.
I've just put a couple of console logs to confirm that `Uri` is the problem and indeed, when i run `Debug Jest Tests`, the uri has the drive letter uppercased.
Also, i am able to replicate this issue manually on the current version of the extension (here it says that [[placeholder]] is referenced in 5 more notes, but 2 of them are duplicates):
image

riccardoferretti:
The example you show is very interesting, also `fileA` is a duplicate, because in that list we take out the current note (and you are in `fileA.md`, which means what you see there is the duplicate).

Given that I have put the letter case normalization in `URI.create` I reckon be that `URI` and `Uri` are getting mixed in such a way as to create this problem. (what happens to you if you remove my change in `URI.create`?)

Thanks for bearing with me on this, windows compatibility has been stinging me from time to time and it's always a drag to sort out, I appreciate the help!

AndreiD049:
Hey, if i reset the implementation of URI.create to below:
```js
  static create(from: Partial<URI>): URI {
    // Normalize the Windows drive letter to lower case
    return {
      scheme: from.scheme ?? _empty,
      authority: from.authority ?? _empty,
      path: from.path ?? _empty,
      query: from.query ?? _empty,
      fragment: from.fragment ?? _empty,
    };
  }
```
Then there are some more failing tests (11), all related to Uri's.

The problem with failing tests is that they use the vsCode's Uri. For example if i just adjust the `createFile` function to return a Foam.Uri:
```js
export const createFile = async (content: string, filepath?: string[]) => {
  const rootUri = vscode.workspace.workspaceFolders[0].uri;
  filepath = filepath ?? [randomString() + '.md'];
  const uri = vscode.Uri.joinPath(rootUri, ...filepath);
  const filenameComponents = path.parse(uri.fsPath);
  await vscode.workspace.fs.writeFile(uri, new TextEncoder().encode(content));
  return { uri: fromVsCodeUri(uri), content, ...filenameComponents }; // here we return Foam.URI instead of vscode.Uri
};
```
Then there is only 1 test failing:
image

I believe this test will always be flaky on windows with current behavior.

AndreiD049:
This of course doesn't fix the issues that was described earlier. I believe there should be some kind of adapter, and Foam should use only one kind of object.

riccardoferretti:
Ok, this is becoming a wider issue, I will organize a couple of PR to address some of the issues we found here and keep things cleaner, but as we are talking about the URI...

I am not surprised that test would fail, as its goal is to check that we don't deviate from the original `vscode.Uri`, which our normalization does. So that test would have to go.

The bigger question here is what you brought up, the compatibility between `vscode.Uri` and `foam.URI`. 
My goal originally was to have them interchangeable, but this is hitting a road block here with the flaky `vscode.Uri` behavior.
We have adapter functions, but we currently only really need `toVsCodeUri` as `vscode.Uri` is a superset of `foam.URI`. This was convenient because you could pass `vscode.Uri` directly into Foam functions, but this is the cause of the issue here..

I can also see a silver lining in going this route, which is the clear distinction of the two classes, and the more I play with the change in the code, the more I come to believe it is a viable approach:
- there are many places where we are using `Uri` instead of `URI`, which was kinda of the original point of the compatible interface, but it does blur the lines, and I am no longer sure that's a good thing
- this level of indirection would allow us to deal with `vscode.Uri` issues, like https://github.com/microsoft/vscode/issues/116298 which caused https://github.com/foambubble/foam/issues/813 (I am making this point because the alternative here would be to just `vscode.Uri` in Foam itself, but that also comes with shortcomings as the Foam API would now depend on VS Code)

Going the route of formally splitting `foam.URI` and `vscode.Uri` would also hit an implementation challenge, because for this to make sense typescript has to enforce it, but typescript uses duck typing and I haven't found a clean way to force the conversion (as a workaround for my testing I just added a `isFoam: true` field to our `URI` interface) short of renaming one of the fields or adding extra ones (or artificially being more restrictive on the types of existing ones). 

Related to this point: https://github.com/microsoft/TypeScript/issues/42698
Which means we could use classes in the future once that issue is resolved, but for the time being it seems like we'd have to artificially add another.

Thoughts?

riccardoferretti:
@AndreiD049 in the end I found a way through, but this PR has become a bit messy, so I have pulled out the `URI` refactoring into https://github.com/foambubble/foam/pull/820

Would love to hear your thoughts (there)

riccardoferretti:
superseded by #881

PR #817, opened 2021-11-04
[WIP] Handle rename and moving of notes

pderaaij:
Very initial experiment to figure out how Foam could handle the renaming and moving of files. Not much to be expected soon, but I like to learn in public.

### DIFF
--- a/packages/foam-vscode/src/core/markdown-provider.ts
+++ b/packages/foam-vscode/src/core/markdown-provider.ts
@@ -43,6 +43,12 @@ export interface ParserPlugin {
   onDidFindProperties?: (properties: any, note: Resource, node: Node) => void;
 }
 
+interface FileRenameEvent {
+  /**
+   * The files that got renamed.
+   */
+  readonly files: ReadonlyArray<{ oldUri: URI; newUri: URI }>;
+}
 export class MarkdownResourceProvider implements ResourceProvider {
   private disposables: IDisposable[] = [];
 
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/src/core/markdown-provider.ts
```diff
@@ -52,6 +58,7 @@ export class MarkdownResourceProvider implements ResourceProvider {
       onDidChange: (uri: URI) => void;
       onDidCreate: (uri: URI) => void;
       onDidDelete: (uri: URI) => void;
+      onDidRenameFiles: (e: FileRenameEvent) => any;
```
riccardoferretti, line 61: I would use a `onDidMove`, as renaming is a subset of moving, and we need to monitor both (e.g. a file that was moved to another folder might need links pointing to it by path to now point to the new location)

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -69,6 +70,43 @@ export class FoamWorkspace implements IDisposable {
     return this;
   }
 
+  rename(oldResource: Resource, newResource: Resource) {
```
riccardoferretti, line 73:
```suggestion
  move(oldResource: URI, newResource: URI) {
```

riccardoferretti:
thanks for taking this on!

Here are my thoughts at a high level:

- in terms of flow FS/VS Code event -> workspace.move(oldUri, newUri) -> providers listening on `workspace.onDidMove` will update references (probably in terms of proposed edits, committed at a second stage, a good reference point for the API can be https://code.visualstudio.com/api/language-extensions/programmatic-language-features#rename-symbols)
- I wonder whether we should support moving a directory, which I reckon would require the API to be array based, and changes to the workspace committed to disk at the end (in case a file that is being moved also needs to be updated)

riccardoferretti:
superseded by #969

PR #825, opened 2021-11-11
Refactored note templates code

riccardoferretti:
## Reason for refactoring

I was struggling to navigate the template code, as there are different layers in the same file and I found it hard to see the dependency relationships amongst functions and entities.
I am focusing on better defining the different services that Foam provides, and cleaning up this code also helps a lot.

## What reviewers should focus on

I believe it's easier to focus on the end result, instead of the changes per se.
Does the resulting design make sense?
Also, I have already done some manual testing, but here I have also leaned into the test coverage (which generally speaking is pretty good in this part of the codebase) - so additional manual testing would be very useful.


movermeyer (commented):
> Does the resulting design make sense?

Seems reasonable to me at a first read.

---

Perhaps add some tests for the new files? Beyond being generally useful, if the tests are hard to write it might reveal that the boundaries are not quite right.

---

I noticed that the selected text is getting replaced with the note title, but it is not a Wikilink:

Given this file:

```markdown
# Existing Note

My text
```

Select `My text` and run `Foam: Create New Note`
Enter a title of `Zoom` and hit return

The existing file now has the contents:

```markdown
# Existing Note

Zoom
```

Instead of the expected (existing behaviour):

```markdown
# Existing Note

[[Zoom]]
```

riccardoferretti:
yup, good catch

I am adding some tests, so far it looks good. I put the PR back in draft mode as I am making a bit more changes

riccardoferretti:
This is now ready for review

pderaaij (approved):
Amazing work!

Issue #806, opened 2021-10-29
Error linking file in subdirectory


memeplex:
- Foam version: v0.15.2
- Platform: macOS Big Sur
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**

**Steps to reproduce**
1. Create a directory d with files a.md and b.md
2. Add `[[d/b]]` to a.md
3. Follow that link: it will offer to create b and the fail because b already exists.

https://user-images.githubusercontent.com/2845433/139410901-3630da90-297c-4d15-8e14-09901a1c4add.mov

memeplex:
The problem is here:

```ts
 static computeRelativeURI(reference: URI, relativeSlug: string): URI {
    // if no extension is provided, use the same extension as the source file
    const slug =
      posix.extname(relativeSlug) !== ''
        ? relativeSlug
        : `${relativeSlug}${posix.extname(reference.path)}`;
    return URI.create({
      ...reference,
      path: posix.join(posix.dirname(reference.path), slug),
    });
  }
```

In `posix.join(posix.dirname(reference.path), slug)` the slug is `dir/file.md` and since dirname only removes the file from the url you end up with `.../dir/dir/file.md`. I will try to fix this.

memeplex:
I will need some guidance regarding the expected behavior here:

1. Is `[[dir/file]]` a supported wikilink?
2. If it is, should it be interpreted as starting:
     a. from the root of the workspace
     b. one level above the source file (as suggested by the current code, but fixed)
     c. relative to the source file (but then the "already exists" error is wrong)

b makes no sense to me.

I would like to interpret it as in c and reserve paths starting with `/` for a, but maybe this isn't intercompatible with other tools, I don't know.

memeplex:
This is the way it's done for Cmd-click:

```ts
        const basedir =
          vscode.workspace.workspaceFolders.length > 0
            ? vscode.workspace.workspaceFolders[0].uri
            : vscode.window.activeTextEditor?.document.uri
            ? URI.getDir(vscode.window.activeTextEditor!.document.uri)
            : undefined;
    ....
    URI.createResourceUriFromPlaceholder(basedir, uri)
```

So here the wikilink is taken to hang from the root of the workspace as in (a) instead of (c) above. Hence this results in an error message.

riccardoferretti:
Hi @memeplex, I have been doing some thinking around it and spoke to other project leads in the space.

Here is where my head is at:

[...]

PR #841, opened 2021-11-22
Add support for wikilinks disambiguation 

riccardoferretti:
In this PR the way wikilinks are handled changes in order to support disambiguation across directories.

See the documentation in the PR for more information.

In summary, the resource `/work/projects/foam/refactoring.md` can be identified by `[[refactoring]]`, `[[foam/refactoring]]`, `[[projects/foam/refactoring]]` and `[[work/projects/foam/refactoring]]` (and of course via a path-like wikilink).

Related: https://github.com/foambubble/foam/issues/806

### Backwards compatibility

- Existing links to notes are backwards compatible
- This PR introduces a **breaking change** in the links to attachments, which now  must include the extension (which was beforehand optional)

### DIFF
--- a/docs/proposals/wikilinks-in-foam.md
+++ b/docs/proposals/wikilinks-in-foam.md
@@ -1,0 +1,97 @@
+# Wikilinks in Foam
+
+Foam supports standard wikilinks in the format `[[wikilink]]`.
+
+Wikilinks can refer to any note or attachment in the repo: `[[note.md]]`, `[[doc.pdf]]`, `[[image.jpg]]`.
+
+The usual wikilink syntax without extension refers to notes: `[[wikilink]]` and `[[wikilink.md]]` are equivalent.
+
+The goal of wikilinks is to uniquily identify a file in a repo, no matter in which directory it lives.
+
+Sometimes in a repo you can have files with the same name in different directories.
+Foam allows you to identify those files using the minimum effort needed to disambiguate them.
+
+This is achieved by adding as many directories above the file needed to uniquely identify the link, e.g. `[[house/todo]]`.
+
+See below for more details.
+
+## Goals for wikilinks in Foam
+
+Wikilinks in Foam are meant to satisfy the following:
+- make it easy for users to identify a resource
+- make it interoperable with other similar note taking systems (Obsidian, Dendron, ...)
+- be easy to get started with, but satisfy growing needs
+
+## Types of wikilinks supported in Foam
+
+Foam supports two types of keys inside a wikilink: a **path** reference and an **identifier** reference:
+
+- `[[./file]]` and `[[../to/another/file]]` are **path** links to a resource, relative _from the source_
+- `[[/path/to/file]]` is a **path** link to a resource, relative _from the repo root_
+- `[[file]]` is an **identifier** of a resource (based on the filename)
+- `[[path/to/file]]` is an **identifier** of a resource (based on the path), the same is true for `[[to/file]]`
+
+It's important to note that sometimes identifier keys can't uniquely locale a resource.
+
+A more concrete example will help:
+
+```
+/
+  projects/
+    house/
+      todo.md
+    buy-car/
+      todo.md
+      cars.md
+  work/
+    todo.md
+    notes.md
+```
+
+In the above repo:
+
+- `[[cars]]` is a unique identifier of a resource - it can be used anywhere in the repo
+- `[[todo]]` is an non-unique identifier as it can refer to multiple resources
+- `[[house/todo]]` is a unique identifier of a resource - it can be used anywhere in the repo
+- `[[projects/house/todo]]` is a unique identifier of a resource - it can be used anywhere in the repo
+- `[[/projects/house/todo]]` is a path reference to a resource
+- `[[./todo]]` is a path reference to a resource (e.g. from `/projects/buy-car/cars.md`)
+
+Basically we could say as a rule:
+
+- if the link starts with `/` or `.` we consider it a **path** reference, in the first case from the repo root, otherwise from the source note
+- if a link doesn't start with `/` or `.` it is an **identifier**
+  - generally speaking we use the shortest identifier available to identify a resource, **but all are valid**
+    - `[[projects/buy-car/cars]]`, `[[buy-car/cars]]`, `[[cars]]` are all unique identifier to the same resource, and are all valid in a document
+    - the same can be said for `[[projects/house/todo]]` and `[[house/todo]]` - but not for `[[todo]]`, because it can refer to more than one resource
+
+## Compatibility with other apps
+
+Foam's identifiers are a super set of Obsidian's: all Obsidian links are supported by Foam, but Foam multi-part identifier (scenario 6) is only supported by Foam.
+
+To improve compatibility this option should either be behind a configuration key, or it should be easily updated e.g. via the janitor.
+
+| Scenario                    | Obsidian                        | Foam                            |
+| --------------------------- | ------------------------------- | ------------------------------- |
+| 1 `[[notes]]`               | ✔ unique identifier in repo     | ✔ unique identifier in repo     |
+| 2 `[[/work/notes]]`         | ✔ valid path from repo root     | ✔ valid path from repo root     |
+| 3 `[[work/notes]]`          | ✔ valid path from repo root     | ✔ valid identifier in repo      |
+| 4 `[[project/house/todo]]`  | ✔ valid path from repo root     | ✔ valid unique identifier       |
+| 5 `[[/project/house/todo]]` | ✔ valid path from repo root     | ✔ valid path from repo root     |
+| 6 `[[house/todo]]`          | ✘ incorrect path from repo root | ✔ valid unique identifier       |
+| 7 `[[todo]]`                | ✘ ambiguous identifier          | ✘ ambiguous identifier          |
+| 8 `[[/house/todo]]`         | ✘ incorrect path from repo root | ✘ incorrect path from repo root |
+
+## Non-unique identifiers
+
+We can't prevent non-unique identifiers from occurring in Foam (first and foremost because a file could be edited with another editor) but we can flag them. 
+
+Therefore Foam follows the following strategy instead:
+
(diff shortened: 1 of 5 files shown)


docs/proposals/wikilinks-in-foam.md
```diff
@@ -0,0 +1,97 @@
+# Wikilinks in Foam
+
+Foam supports standard wikilinks in the format `[[wikilink]]`.
+
+Wikilinks can refer to any note or attachment in the repo: `[[note.md]]`, `[[doc.pdf]]`, `[[image.jpg]]`.
+
+The usual wikilink syntax without extension refers to notes: `[[wikilink]]` and `[[wikilink.md]]` are equivalent.
+
+The goal of wikilinks is to uniquily identify a file in a repo, no matter in which directory it lives.
```
memeplex, line 9: Small typo here: uniquely.

docs/proposals/wikilinks-in-foam.md
```diff
@@ -0,0 +1,97 @@
...
+
+- if the link starts with `/` or `.` we consider it a **path** reference, in the first case from the repo root, otherwise from the source note
+- if a link doesn't start with `/` or `.` it is an **identifier**
+  - generally speaking we use the shortest identifier available to identify a resource, **but all are valid**
+    - `[[projects/buy-car/cars]]`, `[[buy-car/cars]]`, `[[cars]]` are all unique identifier to the same resource, and are all valid in a document
+    - the same can be said for `[[projects/house/todo]]` and `[[house/todo]]` - but not for `[[todo]]`, because it can refer to more than one resource
+
+## Compatibility with other apps
+
+Foam's identifiers are a super set of Obsidian's: all Obsidian links are supported by Foam, but Foam multi-part identifier (scenario 6) is only supported by Foam.
```
memeplex, line 70: This seems to be working in Obsidian after all as I commented in the issue tracker. Obsidian inserts paths as those in 4 but can follow abridged ones as those in 6.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -12,23 +11,19 @@ export function getReferenceType(
...
-};
-const uriToResourceId = (uri: URI) => pathToResourceId(uri.path);
-
-const pathToResourceName = (pathValue: string) =>
-  path.parse(pathValue).name.toLowerCase();
-export const uriToResourceName = (uri: URI) => pathToResourceName(uri.path);
+function hasExtension(path: string): boolean {
+  const dotIdx = path.lastIndexOf('.');
+  return dotIdx > 0 && path.length - dotIdx <= 4;
+}
```
memeplex, line 26: Is this test reliable? There are extensions with more than 3 characters. Perhaps checking if the dot index is greater than the last path separator index? Or using path.extname from node (I'm not a node programmer, I don't really know what I'm talking about)?

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -107,44 +81,53 @@ export class FoamWorkspace implements IDisposable {
...
 
       case 'relative-path':
         if (isNone(reference)) {
           return null;
         }
+        if (!hasExtension(resourceId as string)) {
+          resourceId = resourceId + '.md';
+        }
         const relativePath = resourceId as string;
         const targetUri = URI.computeRelativeURI(reference, relativePath);
```
memeplex, line 129: Above you used the name `resourceUri`, I'm not sure whether there is a difference in meaning here that justifies to call it otherwise.

memeplex (commented):
Some minor and late (sorry) comments.

packages/foam-vscode/src/core/model/workspace.ts
memeplex: There seems to be conflicting criteria for this in other parts of the code.

For example, in utils.ts:

```ts
export function dropExtension(path: string): string {
  const parts = path.split('.');
  parts.pop();
  return parts.join('.');
}
```

This code is strangely repeated in index.ts: 

```ts
export function dropExtension(path: string): string {
  const parts = path.split('.');
  parts.pop();
  return parts.join('.');
}
```

which probably requires some cleanup and deduplication.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -12,23 +11,19 @@ export function getReferenceType(
   if (URI.isUri(reference)) {
```
memeplex, line 11: `getReferenceType` may be a method of `URI` I guess. That said, there is (already was) some overloading of the meaning of URI here: a `URI` may be a reference of different types, including `uri`. I think something like `full-uri` would have been a better name for the reference type, but perhaps it's not worth the refactoring.

packages/foam-vscode/src/core/model/workspace.ts
memeplex: Also this method of URI:

```
  static getFileNameWithoutExtension(uri: URI) {
    return URI.getBasename(uri).replace(/\.[^.]+$/, '');
  }
```

riccardoferretti: Thanks for these comments @memeplex, all very good observations.

For what regards `hasExtension`, yeah, it is an approximation. 
Basically you are suggesting simply checking if a `.` exists in the last path segment... thinking of it, it's a good approach generally speaking, but I don't think it would work in the context in which `hasExtension` is used in `workspace` (which makes me think maybe the name of the function is incorrect in the first place).
You will see what I mean if you change the implementation of the function.

For what regards `dropExtension`, yes we should dedup them, and consolidate where all these URI/path related utility functions live. Currently I think they should mostly be in `URI`, although this one for example is not strictly about the `URI` but adjacent.

For what regards `posix`, yes, I agree, with a big caveat that I haven't checked yet: I want Foam to soon become available as a web extension, which means there are certain limitations on what OS specific libraries we can use. I don't know whether `posix` falls under that. If we can use it in a web extension, then I agree everything should leverage the library.

Does this provide a bit of background? What are your thoughts?

packages/foam-vscode/src/core/model/workspace.ts
riccardoferretti: This is because JS is stupid ;)

Because I prefer to use `const` instead of `let` I need to find different names in the various `case` statements, or it will trigger a compiler error. 
If you use in this line `const resourceUri = URI.computeRelativeURI(reference, relativePath);`, you will see what I mean.

The alternative would be to define `let resourceUri` beforehand and they assign it, which is actually something I am doing now in https://github.com/foambubble/foam/pull/856 exactly for the reason above.

docs/proposals/wikilinks-in-foam.md
riccardoferretti: this is great! I will update the docs

docs/proposals/wikilinks-in-foam.md
riccardoferretti: thanks! fixed in https://github.com/foambubble/foam/commit/59bb2eb38f08056867834cf2ee1ae432d68da67a

docs/proposals/wikilinks-in-foam.md
riccardoferretti: fixed in https://github.com/foambubble/foam/commit/59bb2eb38f08056867834cf2ee1ae432d68da67a

packages/foam-vscode/src/core/model/workspace.ts
memeplex: I think the most important point is to unify all these criteria in one place, then it will be easy to change it to/from posix or whatever when the time comes. The primitive may be `getExtension`, it's easy to implement `hasExtension` and `dropExtension` from it, maybe they're not even worth of independently existing as small utilities and may be inlined in each usage site.  It's unclear to me where something like that should live, the situation with utilities looks a bit messy: utils.ts, utils/*, core/utils/*. Since there already is `dropExtension` in utils.ts that may be the place to implement `getExtension` (and, perhaps, also `dropExtension` and `hasExtension`). Having these extension related functions suggests a path utility class and that already exists in node but given your concerns I would stop at a few simple facade functions. Another option is to make them part of URI but that's already the most bloated and convoluted class in foam, it think it needs to be simplified and not made more complex than it already is.

memeplex: I see URI already uses posix quite a lot: extname, dirname, join. I would like to avoid NIH but if you feel that the dependency on posix is worth abstracting in order to be removed later, something like utils.Path or utils.path could make sense. Of course, not a full path utility but just a place where to put path related utilities (object oriented like `Path(path).getExtension()` or just functions like `utils.path.getExtension(path)`). Again, an alternative is to hide extension manipulation behind URI so everything posix is encapsulated there.

riccardoferretti: I agree, where these functions would live in `utils/path.ts` or even in `URI.ts` (but I tend to prefer the former).
I need to check whether `posix` can be used in a web extension (it might), but in any case I think it would be good to consolidate all path related utility functions (which are an overlapping, but not super nor sub set, of `posix`), and put there whatever is required.
In the codebase we use more the functions approach.
I would expect `URI` to use these utility functions where necessary.

memeplex: From https://code.visualstudio.com/api/extension-guides/web-extensions:

> Node.js globals and libraries such as process, os, setImmediate, path, util, url are not available at runtime. They can, however, be added with tools like webpack. The webpack configuration section explains how this is done.

https://code.visualstudio.com/api/extension-guides/web-extensions#webpack-configuration

Not sure what the meaning of something like process or os would be inside such an environment, though.

riccardoferretti: Do you understand how https://code.visualstudio.com/api/extension-guides/web-extensions#webpack-configuration helps with adding those modules?

Because I sure don't 😄 
(also, I am not a webpack expert, I'd need to take a deeper look at it, maybe @digitarald can also point us in the right direction)

memeplex: No, I haven't the faintest idea. I understand that it bundles the modules but how these modules may be meaningful outside of their natural environment I don't know. Path manipulation is thinkable inside a browser, but most of process, os and other modules mentioned there would be useless.

PR #844, opened 2021-11-23
Added diagnostic for ambiguous wikilinks with quick fix actions

riccardoferretti:
When a wikilink can refer to more than a resource, flag it as a warning for the user.

The user can also quickly correct the error by pressing `cmd+.` and selecting which resource the link is meant for

feature-wikilink-diagnostics


faydanube:
It didn't work...

faydanube:
How can I block this `problem` completely, apart from uninstalling the plugin ?

Issue #294, opened 2020-10-22
Add support for block-level linking


robgrace:
(This is more of a core feature, rather than an enhancement, inasmuch as foam is trying to replicate Roam behavior):

Roam, and now Obsidian, provide support for block-level linking between markdown notes.  It would be great if Roam could follow suit and provide this capapbilty.

Unfortunately, there are now three styles of intra-note linking to choose from:
- **Roam**:   ((block-id))      to text block (bullet)
- **Obsidian**:  [[NoteTitle#^block-id]]     -- link to a text block annotated with `^block-id`
- **Extended Markdown**:    [Description](#heading-id)     -- link to a heading annotated with `{#custom-id}`

I vote for the Obsidian-style linking, since this is the format that I'd imagine most people will be co-editing with vscode.

ingalless:
Hey @robgrace!
Great suggestion, and one that had work started on it a while back:

[Foam Sneak Peek: Block Referencing and Block Embedding](https://youtu.be/ZqLEGGmOw-8)

It's one of the features we want in Foam

5hay:
I really like the `^hash` syntax at the end of a block. Is there any progress made or timeline specified when we can expect a first version that includes this feature? Have a nice Sunday everyone.

kisnows:
Really need this feature.

khramoff:
Hey, @ingalless. Definitely looks great! Is there any news, progress? 

Long time, no see. 
If there's any way, I can participate in this, it would be nice.

ingalless:
@khramoff @5hay I've not heard of any progress recently. Foam has been undergoing a lot of underyling changes as of late. If I hear any whispers I'll try remember to update this 👍🏻

Kcalb35:
I'm looking forward to this feature, because I really need to re-arrange my blocks of notes in another way. Hope to hear of any progress.

ChenguangZhang:
It has been more than 1 year since the sneak peek video was uploaded and there is no block referencing (?). Any workaround?

huyz:
> I vote for the Obsidian-style linking, since this is the format that I'd imagine most people will be co-editing with vscode.

+1.  Dendron also uses the `#^hash` link format.

briansunter:
I would still really like to see this

riccardoferretti:
Yes I agree, and the support for sections (#856) implicitly added support for blocks in the model (as a block can modeled as a "type" of section), now it would be a matter of changing the markdown parser to recognize the anchors.
[...]

PR #856, opened 2021-12-02
Add support for sections

riccardoferretti:
This PR adds support for sections in Foam.

That means that:
- `[[resource#section]]` will be recognized as a valid resource and not as a placeholder
- `[[resource#section]]` will navigate to the `section` line in the `resource` 
- Autocompletion for sections
- `![[resource#section]]` will embed only the text in `section` in the preview
- Diagnostics will report the usage of incorrect sections (e.g. `[[resource#section3]]` when no `section 3` exists in the file), with quick suggestions to fix the error

Note for reviewers: Even after the PR has been merged, feel free to leave comments, I will address them in a follow-up PR.  This is to not slow down work while waiting for feedback.

### DIFF
--- a/packages/foam-vscode/src/core/markdown-provider.ts
+++ b/packages/foam-vscode/src/core/markdown-provider.ts
@@ -237,7 +237,10 @@ const blocksPlugin: ParserPlugin = {
   },
   onDidVisitTree: (tree, note) => {
     if (note.blocks.length > 0) {
-      note.blocks[note.blocks.length - 1].range.end = note.source.end;
+      note.blocks[note.blocks.length - 1].range.end = Position.create(
+        note.source.end.line + 1,
+        0
+      );
     }
   },
 };
--- a/packages/foam-vscode/src/core/markdown-provider.ts
+++ b/packages/foam-vscode/src/core/markdown-provider.ts
@@ -107,8 +107,20 @@ export class MarkdownResourceProvider implements ResourceProvider {
     return this.dataStore.read(uri);
   }
 
-  readAsMarkdown(uri: URI): Promise<string | null> {
-    return this.dataStore.read(uri);
+  async readAsMarkdown(uri: URI): Promise<string | null> {
+    let content = await this.dataStore.read(uri);
+    if (isSome(content) && uri.fragment) {
+      const block = this.parser
+        .parse(uri, content)
+        .blocks.find(b => b.label === uri.fragment);
+      if (isSome(block)) {
+        const rows = content.split('\n');
+        content = rows
+          .slice(block.range.start.line, block.range.end.line)
+          .join('\n');
+      }
+    }
+    return content;
   }
 
   async fetch(uri: URI) {
@@ -133,9 +145,14 @@ export class MarkdownResourceProvider implements ResourceProvider {
             workspace.find(definedUri, resource.uri)?.uri ??
             URI.placeholder(definedUri.path);
         } else {
+          const [target, section] = link.target.split('#');
           targetUri =
-            workspace.find(link.target, resource.uri)?.uri ??
+            workspace.find(target, resource.uri)?.uri ??
             URI.placeholder(link.target);
+
+          if (section && !URI.isPlaceholder(targetUri)) {
+            targetUri = URI.create({ ...targetUri, fragment: section });
+          }
         }
         break;
 
@@ -197,6 +214,30 @@ const tagsPlugin: ParserPlugin = {
           range: Range.createFromPosition(start, end),
         });
       });
+    }
+  },
+};
+
+const blocksPlugin: ParserPlugin = {
+  name: 'block',
+  visit: (node, note) => {
+    if (node.type === 'heading') {
+      const label = ((node as Parent)!.children?.[0] as any)?.value;
+      if (label) {
+        const range = astPositionToFoamRange(node.position!);
+        if (note.blocks.length > 0) {
+          note.blocks[note.blocks.length - 1].range.end = range.start;
+        }
+        note.blocks.push({
+          label,
+          range: range,
+        });
+      }
+    }
+  },
+  onDidVisitTree: (tree, note) => {
+    if (note.blocks.length > 0) {
+      note.blocks[note.blocks.length - 1].range.end = note.source.end;
     }
   },
 };
@@ -314,6 +355,7 @@ export function createMarkdownParser(
     wikilinkPlugin,
     definitionsPlugin,
     tagsPlugin,
+    blocksPlugin,
     ...extraPlugins,
   ];
 
@@ -344,6 +386,7 @@ export function createMarkdownParser(
         type: 'note',
         properties: {},
         title: '',
+        blocks: [],
         tags: [],
         links: [],
         definitions: [],
--- a/packages/foam-vscode/src/core/markdown-provider.ts
+++ b/packages/foam-vscode/src/core/markdown-provider.ts
@@ -110,13 +110,13 @@ export class MarkdownResourceProvider implements ResourceProvider {
   async readAsMarkdown(uri: URI): Promise<string | null> {
     let content = await this.dataStore.read(uri);
     if (isSome(content) && uri.fragment) {
-      const block = this.parser
+      const section = this.parser
         .parse(uri, content)
-        .blocks.find(b => b.label === uri.fragment);
-      if (isSome(block)) {
+        .sections.find(b => b.label === uri.fragment);
+      if (isSome(section)) {
         const rows = content.split('\n');
         content = rows
-          .slice(block.range.start.line, block.range.end.line)
+          .slice(section.range.start.line, section.range.end.line)
           .join('\n');
       }
     }
@@ -222,11 +222,11 @@ const tagsPlugin: ParserPlugin = {
   },
 };
 
-let blockStack: Array<{ label: string; level: number; start: Position }> = [];
-const blocksPlugin: ParserPlugin = {
-  name: 'block',
+let sectionStack: Array<{ label: string; level: number; start: Position }> = [];
+const sectionsPlugin: ParserPlugin = {
+  name: 'section',
   onWillVisitTree: () => {
-    blockStack = [];
+    sectionStack = [];
   },
   visit: (node, note) => {
     if (node.type === 'heading') {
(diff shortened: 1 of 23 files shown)


packages/foam-vscode/src/core/model/note.ts
```diff
@@ -38,12 +38,17 @@ export interface Tag {
...
+  range: Range;
+}
+
 export interface Resource {
   uri: URI;
   type: string;
   title: string;
   properties: any;
-  // sections: NoteSection[]
+  blocks: NoteBlock[];
```
riccardoferretti, line 51: Calling this `blocks` so that in the future it will also house inline blocks (https://github.com/foambubble/foam/issues/294) - sections are just a type of block

packages/foam-vscode/src/features/utility-commands.ts
```diff
@@ -1,54 +1,72 @@
...
 };
 
 const feature: FoamFeature = {
-  activate: (context: vscode.ExtensionContext) => {
+  activate: (context: vscode.ExtensionContext, foamPromise: Promise<Foam>) => {
     context.subscriptions.push(
       vscode.commands.registerCommand(
         OPEN_COMMAND.command,
-        OPEN_COMMAND.execute
+        async (params: { uri: URI }) => {
```
riccardoferretti, line 23: moved the `execute` here as we need `foam` here and trying to be smart about initialization was more convoluted than just doing this.

Commands need a bit of love/refactoring, but this is out of scope for this PR

packages/foam-vscode/src/core/markdown-provider.ts
```diff
@@ -201,6 +222,53 @@ const tagsPlugin: ParserPlugin = {
   },
 };
 
+let blockStack: Array<{ label: string; level: number; start: Position }> = [];
+const blocksPlugin: ParserPlugin = {
```
memeplex, line 226: Not specific of this change, but how well does reparsing the entire file on each change scale? Do you know? I see it's an async operation, but anyway it could be pretty expensive for big files. In the ZK many-small-files scenario is not problem, but if you have some big files in the same workspace, especially now that there is section linking support, couldn't it be a problem?

riccardoferretti: Performance is something that I have been postponing. At some point we'll need to take a serious look at it (starting from whether using a parser is better than regexes, which other tools have preferred). But I have (semi-)consciously decided to leave it for later

packages/foam-vscode/src/core/markdown-provider.ts
```diff
@@ -201,6 +222,53 @@ const tagsPlugin: ParserPlugin = {
...
+  name: 'block',
+  onWillVisitTree: () => {
+    blockStack = [];
+  },
+  visit: (node, note) => {
+    if (node.type === 'heading') {
+      const level = (node as any).depth;
+      const label = ((node as Parent)!.children?.[0] as any)?.value;
```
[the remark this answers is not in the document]
riccardoferretti: Not sure I follow, how would you change the line? 
I admit sometimes I get lost in TS's types so I have overdone it haha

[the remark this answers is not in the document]
memeplex: (anyway it's done this way in other places so it's probably better to leave it alone just for the sake of consistency, but would you mind explaining it?)

packages/foam-vscode/src/core/model/note.ts
riccardoferretti: I see your point about blocks, it is inspired by names we have used before, but I also don't love it.
I am not sure about Section as I don't think a section represents well an "inline block".. mm, I am torn about it.
It's almost an Anchor, but that is a point, not a range.
I guess it could also be considered a Fragment, but that feels confusing with `URI.fragment`.

Other ideas?

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -152,11 +165,23 @@ export class FoamWorkspace implements IDisposable {
...
+    if (!resource) {
+      return null;
+    }
+    if (!fragment) {
+      return resource;
+    }
+    return {
+      ...resource,
+      uri: URI.withFragment(resource.uri, fragment),
+    };
```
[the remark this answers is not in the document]
riccardoferretti: Yes, my thinking here was that if the section doesn't exist we simply return the whole thing. 
We fallback gracefully, returning `null` feels too strict to me.
That being said, there might be some unexpected consequences, which might require us to reconsider this in the future

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -19,11 +18,64 @@ const feature: FoamFeature = {
         mdDocSelector,
         new CompletionProvider(foam.workspace, foam.graph),
         '['
+      ),
+      vscode.languages.registerCompletionItemProvider(
+        mdDocSelector,
+        new SectionCompletionProvider(foam.workspace, foam.graph),
```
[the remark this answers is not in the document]
riccardoferretti: Ooops, good point. This needs to be consolidated once we figure out the most appropriate name for the field/concept

packages/foam-vscode/src/features/navigation-provider.ts
```diff
@@ -110,20 +110,22 @@ export class NavigationProvider
...
-        )
-      ),
+      targetRange: toVsCodeRange(targetRange),
       targetSelectionRange: toVsCodeRange(
-        Range.createFromPosition(
-          targetResource.source.contentStart,
-          targetResource.source.contentStart
-        )
+        Range.createFromPosition(targetRange.start, targetRange.start)
       ),
```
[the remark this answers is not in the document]
riccardoferretti: `targetSelectionRange` with same start and end is so is so that there will be no selection in the editor when navigating the link

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -152,11 +165,23 @@ export class FoamWorkspace implements IDisposable {
...
+
+    if (!resource) {
+      return null;
+    }
+    if (!fragment) {
+      return resource;
+    }
+    return {
+      ...resource,
+      uri: URI.withFragment(resource.uri, fragment),
```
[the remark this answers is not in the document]
riccardoferretti: Yes, I see your point, and this is related to the comment below.

> the representation of a section is a resource that is identical to its entire note except that the uri includes a fragment? That is, all sections in a note have the same representation except for the different fragments in their uris

That's correct.

> may be worth encapsulating as part of the Resource abstraction

I think that's a very good idea

riccardoferretti (commented):
Thanks for the comments, some good pointers in there. 

I am mainly thinking about the Block vs Section debate, I see your point, and the more I think about it (even from when I wrote my reply), the more I tend to agree with you.
Let me think about that one

packages/foam-vscode/src/core/markdown-provider.ts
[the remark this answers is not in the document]
memeplex: I'm just learning about the type system as we speak but I think that once you say `x is T` you don't need to also signal the compiler it can't be null by adding the `!`. Consider this example:

```ts
interface InnerThing {
    realThing: number;
}

interface Thing {
    innerThing?: InnerThing;
}

function doStuffWith(aThing: Thing, anotherThing: Thing) : number {
    return (aThing.innerThing as InnerThing).realThing + anotherThing.innerThing!.realThing;
}
```

When  compiled with strict null checking the last line can't just be  `aThing.innerThing.realThing + anotherThing.innerThing.realThing`, but  both  doing the cast  or adding the `!` would be enough to silence the compiler, you don't need to do both of them.

That said, it doesn't do any harm and since it's consistently used in a similar way at other locations maybe it's better not to touch it.

memeplex:
A couple other things:
1. How will this work with the graph? As it is now, are sections shown at all? What kind of relationship they have with their containing node?
2. What about links to a section in the same file? Suppose I type `[[#`, will that trigger a completion? Will foam be able to solve a link `#...` to a section in its containing file?

memeplex:
Also:

3. What about backlinks?

By looking at  the code I see that they are now based on paths, not uris, so backlinks linking to the file or any of its sections will be shown in the backlink view. Another possibility is to dynamically update this view according to the current section (the section where the cursor is). The second possibility is more sophisticated and is what org-roam does, but Obsidian implements the first one and, since it's simpler and AFAICS (without actually running it) it should be already working in Foam, perhaps it's better to leave it like that. It could be a problem for very big files with many linked sections but I don't see that as a recommended usage pattern for Foam.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -112,36 +112,49 @@ export class FoamWorkspace implements IDisposable {
...
-        const uri = resourceId as URI;
-        return this.exists(uri)
-          ? this.resources.get(normalize(uri.path)) ?? null
-          : null;
+    if (refType === 'uri') {
+      const uri = resourceId as URI;
+      return this.exists(uri)
+        ? this.resources.get(normalize(uri.path)) ?? null
+        : null;
+    }
```
memeplex, line 138: I believe you can reorganize the code in this function in a different way:

```ts
switch (refType) {
      case 'key':
           ...
           return sorted[0];
      case 'absolute-path':
           ...
           resourceUri = ...;
           break;
      case 'relative-path':
           ...
           resourceUri = ...;
           break;
      case 'uri':
           resourceUri = resourceId;
           break;
}
return this.exists(resourceUri)
        ? this.resources.get(normalize(resourceUri.path)) ?? null
        : null;
```

This:
- Makes it clearer that ultimately what is resolved in the workspace is a uri.
- Deduplicates some normalization & resolution logic.
- Deals with the confusing fact of having three names (`uri`, `resourceUri` and `targetUri`) for what is essentially the same thing.

Unfortunately the `key` case has to be dealt in a different way because `listById` already returns resources, but all in all I think this will be an improvement.

riccardoferretti:
> How will this work with the graph? As it is now, are sections shown at all? What kind of relationship they have with their containing node?

The graph only highlights connections amongst resources, it won't go into the detail of the section

> What about links to a section in the same file? Suppose I type [[#, will that trigger a completion? Will foam be able to solve a link #... to a section in its containing file?

Currently it doesn't but that's a good point, I will add that.

> What about backlinks?

Like the graph, backlinks are based on connections amongst files, not sections. I don't believe it's worth going down that level of detail, and it could actually be unintuitive and hard to convey to the user.

packages/foam-vscode/src/core/model/workspace.ts
memeplex: Also the code:

```ts
      return this.exists(uri)
        ? this.resources.get(normalize(uri.path)) ?? null
        : null;
```

in the `uri` clause seems a bit redundant: `exists` checks if the uri is a placeholder and then if `this.resources.get(normalize(uri.path))` is not null. So here you can just check if the uri is a placeholder, since `this.resources.get(normalize(uri.path))` will return null anyway when not found. This way you avoid the double resolution. Moreover, I don't believe the `?? null` is adding anything.

memeplex: Indeed `exists` is only used here and doesn't add too much to the API. After the change above, you can remove `exists` since it's easily replaced by `isSome(ws.find(uri))`, or keep it in the `API` just like an unchecked version of `ws.get(uri)`:

```ts
  public exists(uri: URI): boolean {  // <-- doesn't add too much, maybe remove it
    return isSome(this.find(uri));
  }

  public get(uri: URI): Resource {
    const note = this.find(uri);
    if (isSome(note)) {
      return note;
    } else {
      throw new Error('Resource not found: ' + uri.path);
    }
  }
```

packages/foam-vscode/src/core/model/note.ts
riccardoferretti: decided to go for "section"

packages/foam-vscode/src/core/model/workspace.ts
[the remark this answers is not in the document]
riccardoferretti: done

packages/foam-vscode/src/core/model/workspace.ts
riccardoferretti: Good pointers, I have implemented some in the PR

PR #858, opened 2021-12-05
Refactor uri/path-related code

memeplex:
NOTE: I deleted the description since it was very outdated.


packages/foam-vscode/src/features/dataviz.ts
```diff
@@ -170,15 +171,15 @@ async function getWebviewContent(
   context: vscode.ExtensionContext,
   panel: vscode.WebviewPanel
 ) {
-  const datavizPath = [context.extensionPath, 'static', 'dataviz'];
+  const datavizPath = path.join(context.extensionPath, 'static', 'dataviz');
```
riccardoferretti, line 174: the reason why i didn't do the join here is because I wanted to delegate to VS Code to find/use the right separator. Let's use VS Code capabilities where possible (this will also make it easier to make Foam a web extension)

packages/foam-vscode/src/features/janitor.ts
```diff
@@ -69,7 +70,7 @@ async function janitor(foam: Foam) {
 async function runJanitor(foam: Foam) {
   const notes: Resource[] = foam.workspace
     .list()
-    .filter(r => URI.isMarkdownFile(r.uri));
+    .filter(r => getExtension(r.uri.path) === '.md');
```
riccardoferretti, line 73: I still believe we should have a `isMarkdownFile` function, so that application code doesn't need to know what extension(s) we use (e.g. we might support `.mdx`, `.markdown`, .. in the future)

packages/foam-vscode/src/features/open-random-note.ts
```diff
@@ -12,7 +12,7 @@ const feature: FoamFeature = {
         const currentFile = window.activeTextEditor?.document.uri.path;
         const notes = foam.workspace
           .list()
-          .filter(r => URI.isMarkdownFile(r.uri));
+          .filter(r => getExtension(r.uri.path) === '.md');
```
riccardoferretti, line 15: same here

riccardoferretti (commented):
This is great work @memeplex, quite a big refactoring.
I have left some comments, let me know your thoughts.

memeplex:
@riccardoferretti many of your comments are about the issue of not going through URI but directly using path related functions. My reasoning was this:

> most of the new path functions deal with URI paths, but I have deliberately not made them part of the URI API because (i) they can be used without going through a URI and also (ii) to avoid having a fat URI API that merely delegates to path functions. 

I would like to avoid having the API replicated, that is URI being 50% dumb wrapper for path. And inevitably one has to draw the line somewhere. For example, `removeExtension` is part of URI or an utility?

I see two options:

- Remove the `core.utils.path` module and move anything under `URI`:

```
URI.getPathName(...)
URI.getPathBasename(...)
URI.removePathExtension(...)
URI.joinPaths(...)
URI.toFsPath()
...
```

etc. This will be a fat interface, I would keep it separated below a `// --- path-related API` comment from the rest of the API.

- Make path utilities more object oriented:

```
URI.path.getName(...)
URI.path.getBasename(...)
...
```
This way I implement path-related functionality in one place but it's automatically leveraged by URI. But there is a quirk with functions that modify paths, for example you will have to do `uri.path = uri.path.joinPaths(...)`.

Anyway, now you have to do `uri = URI.joinPaths(uri, ...)` which is not less cumbersome. This raises a different concern that I was wanting to ask you.  Don't you think it would be desirable to have code like `uri = uri.joinPaths(...)` instead? That is, there are many factory static methods to build URIs, which is fine, but also many functions that are clearly instance related, yet the URI class is used as a kind of namespace for purely static functions now, which is not the nicest pattern. The oddness of this will increase by adding the fat path interface. For example, `uri.getPathBase()` vs `URI.getPathBase(uri)`, `uri.isPathRelative()` vs `URI.isPathRelative(uri)` and many others.

If you are really sure all code will manipulate paths through URI (and eventually get a local path using `uri.getFsPath()` if needed), and notice that there is a slight overhead in doing that because the uri object has to be created (also I believe this is negligible), then this is my preferred option:

- Remove the new `core.utils.path` module.
- Move all functions there inside `URI` separated by a comment header from the rest.
- Refactor all naturally instance functions in URI so that they're not static anymore.

The last point is optional, but I do believe it's more relevant after adding all the path related stuff to URI.

memeplex:
@riccardoferretti I've been experimenting with all the variants I described above. All have some drawbacks so I would like to further discuss the matter with you.

* Option 1: move path inside URI. There are many places in the code where paths are plain strings. Some of them should probably be rewritten to use URI instead. But other are legit, low-level path manipulation as in identifier resolution. Building an entire URI for this seems too much to me and some manipulations are very specific/low-level as to include them in the URI API.

* Option 2: make path more OO and expose API through `uri.path : Path`. This is pretty nice, but in places where path-as-string is still needed you have to do `path.toString()`. Still better than the previous option.

* Option 3: duplicate the entire API, as plain functions in `core.utils.path` and as wrappers in `core.model.URI`.  This is a lot of duplication.

* Option 4: current one.

packages/foam-vscode/src/core/janitor/generateLinkReferences.test.ts
```diff
@@ -8,19 +8,21 @@ import { URI } from '../model/uri';
...
 
 Logger.setLevel('error');
 
 describe('generateLinkReferences', () => {
   let _workspace: FoamWorkspace;
+  // TODO slug must be reserved for actual slugs, not file names
   const findBySlug = (slug: string): Resource => {
     return _workspace
       .list()
-      .find(res => URI.getBasename(res.uri) === slug) as Resource;
```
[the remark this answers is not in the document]
memeplex: Please see my comments below. This is a slippery slope, if I add `getBasename` I should as well add a bunch of other methods. This problem is not apparent now because path manipulation logic is scattered across the codebase, but as soon as I tried to unify all usages in a coherent path API it became clear that there are many functions and having them twice, one as functions over plain strings and one as URI wrappers to those functions, is not very nice. I've suggested a number of alternatives, but I still find the current one desirable since IMO better balances the different concerns here: code duplication, low-level (string) vs high-level path access, etc.

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -0,0 +1,88 @@
+import { posix } from 'path';
+import { promises, constants } from 'fs';
+
+export function isIdentifier(uriPath: string): boolean {
```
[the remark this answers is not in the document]
memeplex: `getShortestIdentifier` is in a similar position now, living in `core/utils/core.ts` because it's used in `wikilinks-diagnostics.ts`. So, there are a couple of identifier related utilities (`getShortestIdentifier`  and `isIdentifier`) that haven't a clear home. Having to choose between `core/utils/core.ts` and `core/utils/path.ts` I believe I prefer the second one. And I don't like the idea of putting them in different places. Alternatively, both can be made workspace static public functions.

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -0,0 +1,88 @@
...
+export function toFsPath(uriPath: string): string {
+  if (uriPath[0] === '/' && hasDrive(uriPath, 1)) {
+    return uriPath[1].toUpperCase() + uriPath.substr(2).replace(/\//g, '\\');
+  }
+  return uriPath;
+}
+
+function hasDrive(path: string, idx: number): boolean {
+  return (
+    path.length > idx && path[idx].match(/[a-zA-Z]/) && path[idx + 1] === ':'
```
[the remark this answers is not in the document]
memeplex: Ok, but notice that next to this usage there are trivial replacements `\ <-> /` over the entire path that are  likely much more intensive.

packages/foam-vscode/src/core/janitor/generateHeadings.test.ts
```diff
@@ -16,11 +17,11 @@ describe('generateHeadings', () => {
   const findBySlug = (slug: string): Resource => {
     return _workspace
       .list()
-      .find(res => URI.getBasename(res.uri) === slug) as Resource;
+      .find(res => getName(res.uri.path) === slug) as Resource;
```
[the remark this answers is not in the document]
memeplex: I agree with the rationale, but this is the reason why all the functions in `core/utils/path.ts` manipulate uri/posix paths. So `uri.path` is an object manipulated by functions in `core/utils/path.ts`. Since those in `core/utils/path.ts` are functions over strings, I'm not able to leverage the type system. Having a `Path` object has some drawbacks. Having everything encapsulated inside `URI` has some drawbacks. I've commented about this below and also done a lot of experimentation. So, although I agree with the not having to deal with filesystem specific paths part, my preference is *against* having the path API replicated inside URI, and also *against* having a path abstraction represented as a `Path` class. Instead, I proposed to use the new platform-agnostic functions on the uri.path string.

As a halfway solution I'm adding some path functions to URI, but I'm not very enthusiastic about it.

packages/foam-vscode/src/core/model/uri.ts
```diff
@@ -72,164 +68,82 @@ export abstract class URI {
...
+   * @param paths the paths to add to the URI path
+   * @returns the resulting URI
    */
-  static createResourceUriFromPlaceholder(
-    basedir: URI,
-    placeholderUri: URI
-  ): URI {
-    if (isAbsolute(placeholderUri.path)) {
-      return URI.file(placeholderUri.path);
+  static joinPaths(uri: URI, ...paths: string[]): URI {
```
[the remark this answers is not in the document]
memeplex: Ok.

packages/foam-vscode/src/core/model/workspace.test.ts
```diff
@@ -76,7 +58,7 @@ describe('Workspace resources', () => {
     const ws = createTestWorkspace()
       .set(createTestNote({ uri: 'test-file.md' }))
       .set(createTestNote({ uri: 'file.md' }));
-    expect(ws.listById('file').length).toEqual(1);
+    expect(ws.findByIdentifier('file').length).toEqual(1);
```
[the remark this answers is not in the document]
memeplex: Ok.

packages/foam-vscode/src/features/janitor.ts
memeplex: Ok

packages/foam-vscode/src/features/open-random-note.ts
memeplex: Ok

packages/foam-vscode/src/core/model/uri.ts
```diff
@@ -72,164 +68,82 @@ export abstract class URI {
...
+    if (!uri.path) {
+      throw new Error(`[UriError]: cannot call join path to URI without path`);
     }
-    const tokens = placeholderUri.path.split('/');
-    const path = tokens.slice(0, -1);
-    const filename = tokens.slice(-1);
-    return URI.joinPath(basedir, ...path, `${filename}.md`);
+    let path = pathUtils.joinPaths(uri.path, ...paths);
+    return URI.create({ ...uri, path: path });
```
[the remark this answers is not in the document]
memeplex: Ok.

packages/foam-vscode/src/core/model/uri.ts
```diff
@@ -303,33 +205,6 @@ function percentDecode(str: string): string {
...
- * - always start with /
- *
- * see https://github.com/foambubble/foam/issues/813
- * see https://github.com/microsoft/vscode/issues/43959
- * see https://github.com/microsoft/vscode/issues/116298
- *
- * @param path the path to convert
- * @returns the URI compatible path
- */
-function windowsPathToUriPath(path: string): string {
```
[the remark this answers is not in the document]
memeplex: As I said before, uriPath is used in the sense of platform-agnostic path, as opposed to platform-specific paths like POSIX or Win32. Moreover, this clearly conveys the fact that these paths are often found in `uri.path`. It's a soft way of communicating the missing abstraction that would otherwise require a `Path` class or enlarging the `URI` API too much. I've done some further changes in that direction, but it's not possible to include all path manipulation use cases in the URI API in a way that makes complete sense. Some manipulations are intrinsically low-level and at that level paths are little more than strings, for example when resolving identifiers in the workspace. So I need functions over strings on the one hand, and a higher level abstraction at the other hand. It's not easy to find a fully satisfying middle ground.

packages/foam-vscode/src/features/dataviz.ts
memeplex: I agree with you rationale, but `path.join` *was used* in this case nevertheless, I just splitted its usage in two here. Indeed, this split wasn't deliberate, it's part of a previous refactoring that I'd done in order to replace `path.join` with `joinPaths` as you want. I reverted it because this is not the only  place where `path.join` is used to load foam resources, so I decided to keep the original code in places where foam was loading its resources, in order to avoid problems. If you prefer, I can replace all `path.join` usages with `uri.joinPath` or `path.joinPath` as I see fit (and as I'd originally done). I've commented about this in the PR description.

memeplex:
@riccardoferretti I've tried to address most of your concerns in my new commit. Even though I'm not fond of it, I've added some extra path methods to the URI interface. As discussed elsewhere, some path manipulations still have to go through low-level routines that work on strings, having them as part of URI make no sense.

There are two important changes:
- The URI API is now more object oriented, static factories are still used but functions that were obviously instance related are now instance methods. This change was in part motivated by the inclusion of new instance path-related  functions. But also by its own merits, I find the resulting code less verbose and more natural in most cases. It was a lot of work, but most of the changes are formal in nature.
- `isIdentifier` and `getShortestIdentifier` live now in `core.utils.path`. I've commented about this above. I'm far from satisfied with this solution and open to suggestions.

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -0,0 +1,88 @@
...
+export function parseUNCShare(uncPath: string): [string, string] {
+  const idx = uncPath.indexOf('\\', 2);
+  if (idx === -1) {
+    return [uncPath.substring(2), '\\'];
+  } else {
+    return [uncPath.substring(2, idx), uncPath.substring(idx) || '\\'];
+  }
+}
+
+export function toUriPath(path: string): string {
```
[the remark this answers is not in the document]
memeplex: A way to solve this by ignoring it is:
- Rename `toFsPath/toUriPath` to `toFsPath/fromFsPath` (moreover, this nicely mimics `toVsCodeUri/fromVsCodeUri`).
- Rename all `uriPath` arguments to just `path` on the understanding that `path` means platform-agnostic paths here (the ones in URIs, similar to POSIX).

There are some shortcomings though:
- The relationship between "path in uri.path" and "platform-agnostic path" is not conveyed by the new naming, although since now most path functionality is accessed through URI wrappers this is not as relevant as it was before.
- Some path functions can receive uriPaths as well as fsPaths (specifically, `toUriPath` is able to deal with both). In these cases `path` meant `uriPath|fsPath`. I can call this `anyPath` so as to reserve `path` to mean the former `uriPath`.

So we get:

- `toFsPath/fromFsPath`  to move to/from platform-specific paths.
- A `path` argument means platform-agnostic path.
- A `fsPath` argument means platform-specific path.
- A `uncPath` argument means UNC share.
- A `anyPath` argument means `path|fsPath`.

I prefer the old naming convention for arguments, but all in all I believe this change would be for the best.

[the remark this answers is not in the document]
memeplex: I've addressed this in a slightly different way since the `fromFsPath` name doesn't cut the mustard after all: the input is indeed an `anyPath`, in the sense of `path|fsPath` (for example, when parsing a URI both kind of paths may arrive). So the name is now simply `asPath`. The signature is `asPath(value: string): path`. This function prepares a path: normalizes drive letter case, replaces backslashes with slashes, etc. I'm still reluctant to call it anything posix-related because I think of it as an agnostic path, not a posix path. `asFoamPath` is another option, but `asPath` seems quite decent to me.

memeplex:
For now I've strived to limit myself to formal changes, but here's one thing that doesn't make a lot of sense to me and maybe it's better to raise it now: the criterion to detect Windows paths is to look for a leading drive letter, but this only works with absolute paths AFAICS. Moreover, do we want the platform-specific path for the *current* platform or for some platform heuristically inferred from the path structure?

memeplex:
Another thing I noticed while doing the refactoring. There is some fair amount of code dealing with github style slugs that don't seem to be actually used anywhere. Is this a remnant that should be removed?

packages/foam-vscode/src/core/model/uri.ts
```diff
@@ -72,164 +68,82 @@ export abstract class URI {
...
+    const path = pathUtils.toUriPath(fsPath);
     return URI.create({ scheme: 'file', authority, path });
   }
 
-  static placeholder(key: string): URI {
-    return URI.create({
-      scheme: 'placeholder',
-      path: key,
-    });
+  static placeholder(uriPath: string): URI {
```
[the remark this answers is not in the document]
memeplex: `uriPath` is now simply `path`, the internal format used by foam that is mostly POSIX (but also stuff like C:/p1/p2 that I don't think is actually POSIX, but Windows with slashes that happens to work with POSIX path manipulation functions). I've removed the uriPath name anywhere. The meaning of `uriPath` was (still is) explained in the PR description, is not "slash format" but rather more like "internal generic format required by the functions in core.utils.path". But now that's just `path`.

packages/foam-vscode/src/core/model/uri.ts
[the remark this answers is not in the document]
memeplex: `uriPath` is now just `path` everywhere. This is explained in detail somewhere above or below.

packages/foam-vscode/src/core/utils/path.ts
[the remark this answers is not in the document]
memeplex: `getShortestIdentifier` and `isIdentifier` are now static functions of `FoamWorkspace`. If you prefer I can make them functions of the workspace module, but it's easier to just import `FoamWorkspace` and get everything there.

packages/foam-vscode/src/features/janitor.ts
memeplex: I've moved this into a `isMarkdown` predicate but the current implementation just checks for `.md`. If you want me to add checks for other extensions just let me know.

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -7,8 +7,8 @@ import { FoamFeature } from '../types';
 import { getNoteTooltip, mdDocSelector } from '../utils';
 import { fromVsCodeUri, toVsCodeUri } from '../utils/vsc-utils';
 
-export const WIKILINK_REGEX = /\[\[[^\[\]]*(?!.*\]\])/;
-export const SECTION_REGEX = /\[\[([^\[\]]*#(?!.*\]\]))/;
+export const WIKILINK_REGEX = /\[\[[^[\]]*(?!.*\]\])/;
+export const SECTION_REGEX = /\[\[([^[\]]*#(?!.*\]\]))/;
```
memeplex, line 11: Not related but it keeps the linter happy.

packages/foam-vscode/src/core/markdown-provider.ts
```diff
@@ -530,9 +522,8 @@ export function createMarkdownReferences(
   // Should never occur since we're already in a file,
   if (source?.type !== 'note') {
     console.warn(
-      `Note ${URI.toString(
-        noteUri
-      )} note found in workspace when attempting to generate markdown reference list`
+      `Note ${noteUri.toString()} note found in workspace when attempting \
+      to generate markdown reference list`
```
memeplex, line 526: I was not able to shut up prettier other than doing this.

memeplex:
I've further advanced in the direction of wrapping path utilities as URI methods. Putting aside the API duplication I've extensively pointed out, sometimes it's hard to tell if the rest adds up to a net improvement. Consider:

```ts
      let relativeUri = target.uri.relativeTo(noteUri.getDirectory());
      if (!includeExtension) {
        relativeUri = relativeUri.removeExtension();
      }
```

It's slightly more readable than the bare string path version, but it requires three URI object creations from which only the path key is ever used.

One can live with this until it actually becomes a performance bottleneck, maybe it's just fine.

As I commented before, there are no hard and fast rules except for the extreme cases:

- Always use path-as-string functions in uri.path
- Vs always use path wrapper methods in URI

Notice that this is not the difference between platform-agnostic/dependent paths. In both cases the routines used are platform-agnostic. The URI level provides some convenience (especially now that the wrappers are methods and not static functions), but little in terms of abstraction/encapsulation, and of course a slight overhead. Anything between these two options will be nuanced and require common sense. For now I've used URI wrappers everything except for workspace, where raw path-as-string manipulations are done.

packages/foam-vscode/src/features/dataviz.ts
riccardoferretti: oops I didn't notice, let's keep it as it was before for now (although I agree it needs to be changed), as the PR is already big and we should limit moving pieces here

packages/foam-vscode/src/features/utility-commands.ts
```diff
@@ -27,7 +22,7 @@ const feature: FoamFeature = {
       vscode.commands.registerCommand(
         OPEN_COMMAND.command,
         async (params: { uri: URI }) => {
-          const { uri } = params;
+          const uri = new URI(params.uri);
```
riccardoferretti, line 25: minor - why are we creating a new object here?

packages/foam-vscode/src/features/wikilink-diagnostics.spec.ts
```diff
@@ -103,7 +103,7 @@ describe('Wikilink diagnostics', () => {
     expect(items[0].severity).toEqual(vscode.DiagnosticSeverity.Warning);
     expect(
       items[0].relatedInformation.map(info => info.location.uri.path)
-    ).toEqual([fileA.uri.path, fileB.uri.path]);
+    ).toEqual([fileB.uri.path, fileA.uri.path]);
```
riccardoferretti, line 106: why this change between `fileA` and `fileB`?

riccardoferretti:
A few comments about the spec above (we usually do them in the form of PRs so they can be commented in-line, just like code):

> URI path: is a platform-independent path, similar to POSIX paths, that can act as the path component of an URI. They can be further classified into:
> 
> Relative: a URI path that starts with ./ or ../.
> Absolute: a URI path that starts with /.
> Identifier: a URI path that is neither relative nor absolute.

When does an identifier URI path happen?

> There are a couple of methods (toFsPath and toUriPath

I think somewhere you proposed `toFsPath` and `fromFsPath`, I like these very much

> Consider the URI path /dir1/dir2/name.ext, then:
> /dir1/dir2 is the directory, as returned by getDir.

what is `getDir` for `/dir1/dir2/`/`/dir1/dir2`?

> Path API and URI API

I wouldn't document these here as they are subject to drift from the code.
Instead I would add documentation directly in the code.

> Renaming of listById to findByIdentifier

`listByIdentifier` (this points to what I said above about the drift, I wouldn't go into too much code detail here, let's just make sure the code is properly documented)

I am in general a bit confused about the `References` section.

> Removal of getReferenceType, which is made mostly redundant by other changes proposed here

In what way? I saw it missing, I am not sure I understand "how"

> Path names are sometimes referred as "slugs" in the code

Legacy from the very early days of Foam

> The criterion to identify Windows paths requires the path to be absolute and to start with a drive letter (with an optional slash prefix). An alternative criterion would have been to take a filesystem path to be of the hosting platform's kind, but here I'm being conservative by keeping the preexisting criterion

This was a conscious choice as the second option you listed, which was the case before, caused massive headaches


---

Review will follow shortly

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -0,0 +1,102 @@
+import { CharCode } from '../common/charCode';
+import { posix } from 'path';
+import { promises, constants } from 'fs';
+
+export function asPath(value: string): [string, string] {
```
riccardoferretti, line 5: I am confused by this fn, is this `fromFsPath`? I can't seem to find that

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (value[0] === '/' && hasDrive(value, 1)) {
+    value = '/' + value[1].toUpperCase() + value.substr(2);
+  }
+  if (hasDrive(value, 0)) {
+    value = '/' + value[0].toUpperCase() + value.substr(1).replace(/\\/g, '/');
+  }
+  return [value, authority];
+}
+
+export function toFsPath(path: string, authority?: string): string {
```
riccardoferretti, line 19: next to this one we should have `fromFsPath`

packages/foam-vscode/src/core/utils/path.ts
riccardoferretti: let's add documentation for these functions, let's help future developers working on this codebase :)

riccardoferretti (commented):
Left some comments, I can't say I have wrapped my head around the full scope of the changes, specifically around the win/posix relationship in `path` (and to a lesser extent `uri`). I am not sure if my confusion comes from actual changes, or simply some of the wording in the description of the change (see also my previous PR comment)

Some of the previous comments/threads can't be continued (as they refer to outdated code), so that breaks a bit the conversation and hopefully I am not missing anything here

Let me know your thoughts

packages/foam-vscode/src/core/utils/path.ts
memeplex: I've wrote about `fromFsPath` indeed but later delete the comment as part of some clean up before requesting revision, so as to not overwhelm you yet more. `fromFsPath` is `asPath`, the entry point to the module (originally `toUriPath`). The problem with `fromFsPath` is that the argument is not actually a filesystem path since the same function is used as an entry point both from paths encoded in URLs (for example, in order to normalize the drive letter) and from filesystem paths. This was also the case for the original `toUriPath`. So there is an asymmetry that works against having the pair `fromFsPath / toFsPath`.

packages/foam-vscode/src/features/dataviz.ts
memeplex: Sure.

memeplex:
> A few comments about the spec above

Notice that at this point it's significatively outdated, I will update it as soon as we agree on the approach.

> When does an identifier URI path happen?

There is nothing special here. In the old terminology uriPath = system independent path. And in the context of resolution a path is classified in absolute `/...`, relative `./...` `../...` and everything else (identifier). 

> I think somewhere you proposed toFsPath and fromFsPath, I like these very much

I'd proposed them but unfortunately `fromFsPath` is not a general enough name, they aren't a symmetric pair, the entry method takes anything (a path encoded in an URL, a native path, etc) and produces a system independent path ready to be used by the rest of the functions, while the exit method just produces native paths. So the pair is now `asPath / toFsPath`.

> what is getDir for /dir1/dir2///dir1/dir2?

I assume you meant `/dir1/dir2` vs `/dir1/dir2/file.txt` ? There is no way to tell one situation from the other at the path level, a (possibly async) system call would be required in order to behave in different ways. Personally I would like:
`getDirectory(/dir1/dir2) = getDirectory(/dir1/dir2/file.txt) = /dir1/dir2` but that's not possible and the standard behavior is to remove the last component, be it a file or a directory. That's why I added an optional `isDirectory?` flag to some functions/methods. It's usually clear from the calling context which is the case.

> I wouldn't document these here as they are subject to drift from the code.
> Instead I would add documentation directly in the code.

Ok, I'm going to add docstrings for the most complex or central functions. I usually don't add trivial documentation for functions that are clear enough, but I can do that for the entire public API if you prefer.

> Renaming of listById to findByIdentifier

I've already done this a few commits ago, isn't it there?

> I am in general a bit confused about the References section.
> Removal of getReferenceType, which is made mostly redundant by other changes proposed here
> In what way? I saw it missing, I am not sure I understand "how"

You have `isAbsolute` in `path.ts` and `isIdentifier` in `FoamWorkspace`. There are 4 kinds of references:

* URI -> has a distinct type
* string
   - isIdentifier
   - isAbsolute
   - other (is relative)

This is what's documented in the section. The code is structured just like that:

```ts
    if (reference instanceof URI) {
     ...
    }
    let [path, fragment] = (reference as string).split('#');
    if (FoamWorkspace.isIdentifier(path)) {
      ...
    } else {
      if (isAbsolute(path) || isSome(baseUri)) {
        ....
      }
    }
```

packages/foam-vscode/src/core/utils/path.ts
memeplex: See comment above.

packages/foam-vscode/src/features/wikilink-diagnostics.spec.ts
memeplex: Because in the specification it's now `listByIdentifier` that takes care of returning its results in a stable order, instead of its user `find`. I thought that it would be better to enforce this behavior from the source itself, but if you don't agree I can move it downwards again. Anyway, that's the reason why the order was different, this happens in like 135 of `wikilinks-diagnostics.ts`:

```ts
const targets = workspace.listByIdentifier(target);
```

packages/foam-vscode/src/features/utility-commands.ts
memeplex: Because what is coming is a parse of the json encoded object, which is just a bare object, not a proper instance of `URI`. So object-things do work but instance-things don't. I can't remember which was the problem right now, but I guess that `uri.getName()` was.

packages/foam-vscode/src/core/model/uri.ts
```diff
@@ -24,258 +23,133 @@ import { CharCode } from '../common/charCode';
...
 
-  static getDir(uri: URI) {
-    return URI.file(posix.dirname(uri.path));
+  getExtension(): string {
+    return pathUtils.getExtension(this.path);
   }
 
-  static getFileNameWithoutExtension(uri: URI) {
-    return URI.getBasename(uri).replace(/\.[^.]+$/, '');
+  changeExtension(from: string, to: string): URI {
```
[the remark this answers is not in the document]
memeplex: :)

packages/foam-vscode/src/core/model/uri.ts
```diff
@@ -24,258 +23,133 @@ import { CharCode } from '../common/charCode';
...
  * ```
  */
-export interface URI {
+
+const _empty = '';
+const _slash = '/';
+const _regexp = /^(([^:/?#]{2,}?):)?(\/\/([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?/;
+
+export class URI {
   scheme: string;
```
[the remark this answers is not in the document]
memeplex: Good idea.

memeplex:
> Left some comments, I can't say I have wrapped my head around the full scope of the changes, specifically around the win/posix relationship in path (and to a lesser extent uri).

The uri path terminology was eradicated from the code, I believe part of your confusion comes from the fact that you're reading an outdated PR description (sorry for that, I've done many revisions after your first batch of comments and wasn't able to update the spec yet).

Actually the matter is not obscure:

```
asPath(value) -> pathFunction1(path), pathFunction2(path), ... -> toFsPath(path)
```

`asPath` is an entry "normalizer" of paths that prepare them to work with the other path functions. `toFsPath` is an exit function that returns a local  path. Internally all functions work with the normalized format (POSIX-like). Originally this internal format was called "uri path" because it was indeed used for paths living in `uri.path` and all path functions were intended to work mostly on `uri.path`. Later I renamed this format to just "path". But it's the same format Foam uses in uris, with or without this PR: uppercase drive letter, forward slashes, etc.


In terms of URI the above looks like:

```
URI.file(value) -> uri.pathFunction1(), uri.pathFunction2(), ... -> uri.toFsPath()
```

Feel free to contact me in Discord to ask me about this.

packages/foam-vscode/src/core/utils/path.ts
memeplex: I've introduced the `fromFsPath` name after all.

memeplex:
@riccardoferretti after further thinking about it I decided to introduce the `fromFsPath` name. There is some ambiguity, as I've remarked before, in that POSIX paths are used both as inputs to `fromFsPath` and as platform-agnostic paths, I've used a neutral language in the docstring trying not to highlight this dual role too much.

I've also documented the entire public path API.

memeplex:
>> The criterion to identify Windows paths requires the path to be absolute and to start with a drive letter 

> This was a conscious choice as the second option you listed

Ok, but the criterion is not without pitfalls. For example, `c:` is a valid POSIX file name. Sure, naming directories like that is asking for trouble, but are you fine with this?

The remaining `path.join` usages are mostly related to Foam resources loading, I would like to replace them by `core.utils.path.joinPath` and I have two alternatives here:
* Check for a drive letter in the first directory and if there is one use `win32.join`.
* Add a `isFs?: boolean` argument and use `path.join` in that case.

The first option is compatible with `fromFsPath`, the second option is compatible with current usages to be replaced.

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -52,7 +50,7 @@ export function toFsPath(path: string, authority?: string): string {
  *    - /d1/d2 -> /d1
  *
  * @param path A POSIX path.
- * @returns true is the path is absolute, false otherwise.
+ * @returns true if the path is absolute, false otherwise.
  */
 export function isAbsolute(path: string): boolean {
   return posix.isAbsolute(path);
```
riccardoferretti, line 56: can we check if it starts with `/`?

packages/foam-vscode/src/features/create-from-template.spec.ts
```diff
@@ -142,12 +142,12 @@ Template A
 
     it('can be cancelled', async () => {
       // This is the default template which would be created.
-      const template = path.join(
-        workspace.workspaceFolders[0].uri.fsPath,
+      const template = Uri.joinPath(
```
riccardoferretti, line 145: I like this

riccardoferretti (commented):
Only a question around the posix usage, otherwise everything looks good in the latest delta

packages/foam-vscode/src/core/utils/path.ts
riccardoferretti: btw, this is minor, as we agreed that the removal of `posix` can happen as a second step

riccardoferretti (approved):
Nevermind my previous point, I think this is all good now.
Let's merge it and if something arises address it individually, this is already a many notches better than the existing path code/design.

riccardoferretti:
@allcontributors add @memeplex for code

riccardoferretti:
just wanted to be sure 🤣

PR #860, opened 2021-12-07
Error fix for janitor update links

jeffrywalker:
`i` was able to have a value of `i = lines.length` which will result in `undefined TypeError` when attempting `lines[i].length`

### DIFF
--- a/packages/foam-vscode/src/core/janitor/apply-text-edit.ts
+++ b/packages/foam-vscode/src/core/janitor/apply-text-edit.ts
@@ -34,5 +34,6 @@ const getOffset = (
     offset = offset + lines[i].length + eolLen;
     i++;
   }
+  i = Math.min(i, lines.length - 1);
   return offset + Math.min(position.character, lines[i].length);
 };


riccardoferretti:
Can you share a scenario to reproduce the problem? 
It might help to better understand the issue and the fix, as I am unclear by looking at the code how `i` could get to `lines.length`... maybe I am missing something

jeffrywalker:
To reproduce the issue as I initially observed it:
1. extract the foam template
2. change `foam.edit.linkReferenceDefinitions` to `withExtensions`
3. extract the attached **zotero.zip** in the template root directory. This will provide some entries generated by _Zotero MdNotes_ 
```
...
- attachments
- docs
- zotero
     /author
     /collection
    - nist2008.md
    - stevens2016.md
    - zipfelModeling2007.md
```
4. run foam Janitor
[zotero.zip](https://github.com/foambubble/foam/files/7678459/zotero.zip)

image

memeplex:
There is indeed an error here:  `i` tracks the next line, if `position.line == lines.length` the loop will end up with `i == lines.length`.

But the way this is fixed in this PR is wrong. By making `i` the last line, you end up adding its length twice (it's already included in `offset`).

Here are some alternatives:

```ts
return offset + Math.min(position.character, lines[i]?.length ?? 0);
```

or:

```ts
const length = i < lines.length ? lines[i].length : 0;
return offset + Math.min(position.character, length);
```

riccardoferretti:
good point @memeplex 

@jeffrywalker have you had a chance to look at the proposal? would also be good to add a couple of tests to `apply-text-edit.test.ts`

jeffrywalker:
@riccardoferretti I have not had the opportunity to look though glancing at the suggestion seems like a pretty straight forward minor change.

riccardoferretti:
hi @jeffrywalker let me know if you will be able to implement the suggested change and some tests for it (I don't like to step on other people's work, at the same time I am keen to bring the PR to a resolution :) ), thanks!

jeffrywalker:
@riccardoferretti I have been too busy to revisit this; if you have the time to resolve this matter feel free to do so.

riccardoferretti:
Thanks for pointing out the issue, and to @memeplex for sharing the fix. I have committed it and it will be out in the next version.

Issue #852, opened 2021-12-01
Bad theme of graph filter in light mode


memeplex:
- Foam version: v0.16.0
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**

When using vscode light themes, including the default light theme, the graph filter panel looks like:

![image](https://user-images.githubusercontent.com/2845433/144169311-56c8fba0-0ab5-4807-b69c-543c73f513ae.png)

Besides the panel being the wrong color, there is an usability issue because some labels are not visible and other ones are very low contrast.

**Steps to reproduce**
1. Open the foam template project
2. Set a light theme
3. Open foam graph
4. Expand the filter panel

Issue #854, opened 2021-12-01
Weird layout of graph filter panel


memeplex:
- Foam version: v0.16.0
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**

The panel starts like this:

![image](https://user-images.githubusercontent.com/2845433/144170364-36b1d83e-5815-4e69-9a71-a5800e03cabb.png)

When I resize it it gets like this:

![image](https://user-images.githubusercontent.com/2845433/144170398-ec20c20e-d8cb-4ae4-92e2-725dab6ee63a.png)

It's like the right half of the panel is kept blank, even when that means truncating labels.

Issue #853, opened 2021-12-01
Graph starts too small


memeplex:
- Foam version: v0.16.0
- Platform: Mac
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes | No

**Summary**

The initial size of the graph is unnecessarily small:

![image](https://user-images.githubusercontent.com/2845433/144170201-d4545db8-6e7d-4ee6-8543-24de7dd4ac59.png)

PR #875, opened 2021-12-13
Graph improvements: light theme, zoom to fit canvas, dat.gui layout

memeplex:
Fix https://github.com/foambubble/foam/issues/852,  https://github.com/foambubble/foam/issues/853 and https://github.com/foambubble/foam/issues/854.

Generalize mechanism to load external resources. Then load new graph.css which:
- Resizes property/value from 40%/60% to 90%/10% (better suited to checkboxes).
- Adds a new light theme based on [1] (but with a darker arrow that I've created in order to increase contrast). This theme gets activated if the graph is loaded when vscode theme is light.

Also zooms the graph to fit the canvas at startup. This is a bit tricky because the graph engine needs some iterations (ticks) before cooling down. Initially the graph zoom level is  heuristically computed in a way that's inversely proportional to the number of nodes. I find this heuristic tends to compress the graph too much at the center of the webview, so I scaled the initial value x 1.5 (which still seems to be pretty conservative). Then the number of ticks to cool down is set to 100. Finally the graph is scaled to fit the canvas with a 0.5 seconds animation. Overall, the effect is nice and the resulting graph takes advantage of all available space. Values are based on [2].

```ts
        graph.zoom(graph.zoom() * 1.5);
        graph.cooldownTicks(100);
        graph.onEngineStop(() => graph.zoomToFit(500));
```

Little refactor: a number of `==` and `!=` in graph.js that the linter was complaining about were changed to `===` and `!==`.

---
[1] https://brm.io/dat-gui-light-theme/
[2] https://github.com/vasturiano/force-graph/blob/master/example/fit-to-canvas/index.html

### DIFF
--- a/packages/foam-vscode/static/dataviz/graph.js
+++ b/packages/foam-vscode/static/dataviz/graph.js
@@ -136,7 +136,7 @@ const Actions = {
         }
       });
       types.forEach(type => {
-        if (model.showNodesOfType[type] == null) {
+        if (model.showNodesOfType[type] === null) {
           model.showNodesOfType[type] = true;
         }
       });
@@ -148,7 +148,7 @@ const Actions = {
       if (!isAppend) {
         m.selectedNodes.clear();
       }
-      if (nodeId != null) {
+      if (nodeId !== null) {
         m.selectedNodes.add(nodeId);
       }
     }),
@@ -205,7 +205,7 @@ function initDataviz(channel) {
     )
     .nodeCanvasObject((node, ctx, globalScale) => {
       const info = model.graph.nodeInfo[node.id];
-      if (info == null) {
+      if (info === null) {
         console.error(`Could not find info for node ${node.id} - skipping`);
         return;
       }
@@ -299,7 +299,7 @@ function updateForceGraphDataFromModel(m) {
   });
   // apply the delta
   nodeIdsToRemove.forEach(id => {
-    const index = m.data.nodes.findIndex(n => n.id == id);
+    const index = m.data.nodes.findIndex(n => n.id === id);
     m.data.nodes.splice(index, 1); // delete the element
   });
   nodeIdsToAdd.forEach(nodeId => {
@@ -312,19 +312,19 @@ function updateForceGraphDataFromModel(m) {
   m.data.links = m.graph.links
     .filter(link => {
       const isSource = Object.values(m.data.nodes).some(
-        node => node.id == link.source
+        node => node.id === link.source
       );
       const isTarget = Object.values(m.data.nodes).some(
-        node => node.id == link.target
+        node => node.id === link.target
       );
       return isSource && isTarget;
     })
     .map(link => ({ ...link }));
 
   // check that selected/hovered nodes are still valid (see #397)
-  m.hoverNode = m.graph.nodeInfo[m.hoverNode] != null ? m.hoverNode : null;
+  m.hoverNode = m.graph.nodeInfo[m.hoverNode] !== null ? m.hoverNode : null;
   m.selectedNodes = new Set(
-    Array.from(m.selectedNodes).filter(nId => m.graph.nodeInfo[nId] != null)
+    Array.from(m.selectedNodes).filter(nId => m.graph.nodeInfo[nId] !== null)
   );
 
   // annoying we need to call this function, but I haven't found a good workaround
@@ -399,7 +399,7 @@ function getLinkState(link, model) {
     ? 'regular'
     : Array.from(model.focusLinks).some(
         fLink =>
-          fLink.source == link.source.id && fLink.target == link.target.id
+          fLink.source === link.source.id && fLink.target === link.target.id
       )
     ? 'highlighted'
     : 'lessened';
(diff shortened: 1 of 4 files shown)


packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -205,7 +205,7 @@ function initDataviz(channel) {
     )
     .nodeCanvasObject((node, ctx, globalScale) => {
       const info = model.graph.nodeInfo[node.id];
-      if (info == null) {
+      if (info === null) {
```
riccardoferretti, line 208: `== null` is different from `=== null`.  The former is also `true` against `undefined`

`undefined === null  => false`
`undefined == null => true`

That is why for these checks we use `==`

memeplex: I'm aware of the difference, I've just assumed that it was a stylistical matter here because the same style is used all over the place, in most cases it doesn't seem to have any semantic motivation, but if there is any doubt regarding that I prefer to revert all those changes altogether.

riccardoferretti: yeah, if there is no `model.graph.nodeInfo[node.id]`, `info` would be `undefined`.

but the places where you changed it for the string comparison I think look good, I would keep those

packages/foam-vscode/static/dataviz/index.html
```diff
@@ -2,14 +2,10 @@
...
     <meta charset="utf-8" />
-    <script data-replace src="./d3.v6.min.js"></script>
-    <script data-replace src="./force-graph.1.40.5.min.js"></script>
-    <script data-replace src="./dat.gui.min.js"></script>
-    <style>
-      body {
-        overflow: hidden;
-      }
-    </style>
+    <script replace-src="./d3.v6.min.js"></script>
```
riccardoferretti, line 5: what's the reason for using `replace-src`?

the reason why we use `data-replace src=` is that it's a valid html tag, which means you can still open the webpage in a browser (outside of vscode) if you want to do some testing (the dev cycle tends to be faster with that approach, although I have never documented it)

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -501,6 +501,9 @@ try {
       case 'didUpdateGraphData':
         graphData = augmentGraphInfo(message.payload);
         Actions.refreshWorkspaceData(graphData);
+        graph.zoom(graph.zoom() * 1.5);
+        graph.cooldownTicks(100);
+        graph.onEngineStop(() => graph.zoomToFit(500));
```
riccardoferretti, line 506: does this mean that every time a file is saved the graph is put back into a zoom-to-fit state?

riccardoferretti (commented):
The new styling approach looks great, I left a couple of questions on other bits

packages/foam-vscode/static/dataviz/index.html
memeplex: Nothing in particular, I needed to generalize the mechanism to also include href and I just did it this way, but I see your point.

memeplex:
@riccardoferretti I believe I've now addressed all of the concerns you raised.

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -501,6 +502,15 @@ try {
...
         Actions.refreshWorkspaceData(graphData);
+        if (!graphUpdated) {
+          window.graphUpdated = true;
+          graph.zoom(graph.zoom() * 1.5);
+          graph.cooldownTicks(100);
+          graph.onEngineStop(() => {
+            graph.onEngineStop(() => {});
+            graph.zoomToFit(500);
+          });
+        }
```
memeplex, line 513: It's not easy to put this in `initDataviz` because `zoom()` depends on the number of nodes and, regarding the rest, one would have said that they should work that way but in practice it's not the case. Moreover, the API is not documented down to the required level of detail. This approach is safer and puts all the relevant code together in the same place.

packages/foam-vscode/static/dataviz/index.html
```diff
@@ -23,6 +19,6 @@
       4. open this file in a browser
     -->
     <!-- <script src="./test-data.js"></script> -->
-    <script data-replace src="./graph.js"></script>
+    <script replace-data src="./graph.js"></script>
```
riccardoferretti, line 22: to be valid html data attributes need to start with `data-` https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

packages/foam-vscode/static/dataviz/graph.js
```diff
@@ -136,7 +136,7 @@ const Actions = {
         }
       });
       types.forEach(type => {
-        if (model.showNodesOfType[type] == null) {
+        if (model.showNodesOfType[type] === null) {
```
riccardoferretti, line 139:
```suggestion
        if (model.showNodesOfType[type] == null) {
```

Otherwise custom times don't show up in the graph

riccardoferretti (commented):
looking good, a couple more comments and I think we are there

packages/foam-vscode/static/dataviz/index.html
memeplex: Oops didn't realize.

packages/foam-vscode/static/dataviz/graph.js
memeplex: Missed this one.

memeplex:
> looking good, a couple more comments and I think we are there

Done. Sorry, I'm a bit unfocused switching between work and foam PRs.

riccardoferretti (approved):

Issue #772, opened 2021-10-01
Weird wikilink highlighting


memeplex:
- Foam version: 0.14.2
- Platform: macOS Big Sur
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

**Summary**

External brackets are not colorized, just one internal bracket is colorized.

**Steps to reproduce**

1. Install foam, clone the default template, open any file with a wikilink.

**Additional information**

![image](https://user-images.githubusercontent.com/2845433/135566321-9ed6a9ef-708f-437a-9d61-95b67b7f0701.png)

memeplex:
Hey, just to know what to expect, can anyone confirm this is a known issue? I'm unable to see anything related to syntax highlighting in the code, likewise for https://github.com/foambubble/foam/issues/773.

For wikilinks, the above screenshot shows the same highlighting than the built-in markdown mode applies. Instead, the screenshot in https://github.com/foambubble/foam/issues/773 shows some failed attempt to improve over that module.

Both seem too basic stuff for a tool that works with wikilinks and tags, so I suspect something is broken in my installation, but I've repeated everything from scratch to no avail. Or maybe syntax highlighting is still in you backlog so this would be a FR.

Full disclosure: I'm analyzing moving from org-mode but it's kind of a toll order since I have accumulated many notes over the years and it makes me a little nervous to see this basic stuff not working. But I understand it's just cosmetic, so if that's because of prioritization it's just fine.

pderaaij:
I have the same behaviour. It happens when the wikilink only contains a single word. Personally, I seldom have notes with a single word, so never noticed it. I'll have a look at it. Thanks for notifying!

memeplex:
I've filed a FR in https://github.com/microsoft/vscode/issues/138173 to avoid the default weird highlighting, which is still visible before foam decorations are painted.

riccardoferretti:
thanks for this @memeplex, for what is worth you got my 👍 on your issue, let's see what the VS Code team says

memeplex:
I was thinking that perhaps a syntax injection for wikilinks that considers every `[[...]]` as a resolved link and a decorator that marks unresolved links would solve both problems:

1. No weird default syntax for `[[..]]`.
2. No decoration delay (except for the rare case of unresolved links).

riccardoferretti:
That's an interesting idea, I like it

memeplex:
I've it in my TODO list, will eventually work on that.

PR #876, opened 2021-12-13
Use syntax injection for wikilinks

memeplex:
Implements the strategy proposed in https://github.com/foambubble/foam/issues/772#issuecomment-988269789

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -33,16 +33,27 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
-    "grammars": [
-      {
-        "path": "./syntaxes/injection.json",
-        "scopeName": "wikilink.injection",
-        "injectTo": ["text.html.markdown"]
-      }
-    ],
     "markdown.markdownItPlugins": true,
     "markdown.previewStyles": [
       "./static/preview/style.css"
+    ],
+    "grammars": [
+      {
+        "path": "./syntaxes/injection.json",
+        "scopeName": "foam.wikilink.injection",
+        "injectTo": ["text.html.markdown"]
+      }
+    ],
+    "colors": [
+      {
+        "id": "foam.placeholder",
+        "description": "Color of foam placeholders.",
+        "defaults": {
+          "dark": "editorWarning.foreground",
+          "light": "editorWarning.foreground",
+          "highContrast": "editorWarning.foreground"
+        }
+      }
     ],
     "views": {
       "explorer": [
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -33,6 +33,13 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
+    "grammars": [
+      {
+        "path": "./syntaxes/injection.json",
+        "scopeName": "wikilink.injection",
+        "injectTo": ["text.html.markdown"]
+      }
+    ],
     "markdown.markdownItPlugins": true,
     "markdown.previewStyles": [
       "./static/preview/style.css"
--- a/packages/foam-vscode/src/features/document-decorator.ts
+++ b/packages/foam-vscode/src/features/document-decorator.ts
@@ -12,13 +12,6 @@ import { fromVsCodeUri } from '../utils/vsc-utils';
 
 export const CONFIG_KEY = 'decorations.links.enable';
 
-const linkDecoration = vscode.window.createTextEditorDecorationType({
-  rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
-  textDecoration: 'none',
-  color: { id: 'textLink.foreground' },
-  cursor: 'pointer',
-});
-
 const placeholderDecoration = vscode.window.createTextEditorDecorationType({
   rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
   textDecoration: 'none',
@@ -42,17 +35,13 @@ const updateDecorations = (
     fromVsCodeUri(editor.document.uri),
     editor.document.getText()
   );
-  let linkRanges = [];
   let placeholderRanges = [];
   note.links.forEach(link => {
     const linkUri = workspace.resolveLink(note, link);
     if (linkUri.isPlaceholder()) {
       placeholderRanges.push(link.range);
-    } else {
-      linkRanges.push(link.range);
     }
   });
-  editor.setDecorations(linkDecoration, linkRanges);
   editor.setDecorations(placeholderDecoration, placeholderRanges);
 };
 
@@ -82,7 +71,6 @@ const feature: FoamFeature = {
 
     context.subscriptions.push(
       areDecorationsEnabled,
-      linkDecoration,
       placeholderDecoration,
       vscode.window.onDidChangeActiveTextEditor(editor => {
         activeEditor = editor;
--- a/packages/foam-vscode/src/features/document-decorator.ts
+++ b/packages/foam-vscode/src/features/document-decorator.ts
@@ -15,7 +15,7 @@ export const CONFIG_KEY = 'decorations.links.enable';
 const placeholderDecoration = vscode.window.createTextEditorDecorationType({
   rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
   textDecoration: 'none',
-  color: { id: 'editorWarning.foreground' },
+  color: { id: 'foam.placeholder' },
   cursor: 'pointer',
 });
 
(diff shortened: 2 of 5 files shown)


riccardoferretti:
Thanks for this, pretty cool.
I am not familiar with the grammar injection, will take a deeper look later.
How do you define/set the color? does it live in the theme?

riccardoferretti:
Nevermind, saw this: https://github.com/foambubble/foam/pull/778#issuecomment-992079800

riccardoferretti:
I have done a test with a few themes, but I am not thrilled by the end result:

image

I don't like the inconsistencies between our decorations and the syntax injection, as well as our linking and the syntax decoration. In the image above:
- the underlining includes `[[` and `]]` but the color of the resource doesn't
- the color of the resource doesn't include the brackets, but for placeholders it does
- placeholders don't have an underline but the resources do

Here are a few more themes:

image

image

image

image

Also, maybe we should use `x` instead of `y` as most themes seem to highlight the label?

image

image


Thoughts?

memeplex:
There is something wrong there, it should be like this:

image

memeplex:
Dracula:

image

Noctis:

image

Default light:

image

memeplex:
Could you run the "Developer: Inspect editor tokens and scopes" command over a wikilink?

riccardoferretti:
image

image

image

image

To make in line with the other VS Code styles, I'd suggest:
- limit the link to the text of the wikilink, without the brackets
- use the color of a regular link label (not url)

memeplex:
I believe your writer extension is interfering somehow. `text.markdown.writer.wiki-link` is not anything I'm setting. OTOH, you should see `markup.underline.link.markdown` which is what the injection sets.

Take a look at https://github.com/foambubble/foam/pull/876/files#diff-12db166a30e824ef08df70891edbe03bf7bd996d39734072672f5a93187553e3, it's pretty self explanatory. The documentation is in https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#injection-grammars. It would help if you could diagnose what's happening at your side.

I'm ok with excluding the delimiters from the highlighting if you prefer so. Should do that for placeholders too, since current decorators include delimiters too.

riccardoferretti:
> I'm ok with excluding the delimiters from the highlighting if you prefer so. Should do that for placeholders too, since current decorators include delimiters too.

yes I agree decorations for placeholders and wikilinks should have the same params and only change in color

riccardoferretti:
> I believe your writer extension is interfering somehow. text.markdown.writer.wiki-link is not anything I'm setting. OTOH, you should see markup.underline.link.markdown which is what the injection sets

In that case this is not an issue, as we can't account for all contributions.
I would default the color though to what we had before, so that customization is still possible, but the release doesn't unexpectedly break with the previous style, what do you think?

memeplex:
I've been exploring this. There  are two ways to change colors:

```json
{
  "workbench.colorCustomizations": {
    "titleBar.activeBackground": "#ff0000"
  }
}
```

for workbench colors, and

```json
{
  "editor.tokenColorCustomizations": {
    "comments": "#FF0000"
  }
}
```

for  token (defined by a grammar) colors. `colorCustomizations` can take color identifiers defined as contributions in `package.json`, for example we could define `foam.placeholder` as:

```json
    "colors": [
      {
        "id": "foam.placeholder",
        "description": "Color of foam placeholders.",
        "defaults": {
          "dark": "editorWarning.foreground",
          "light": "editorWarning.foreground",
          "highContrast": "editorWarning.foreground"
        }
      }
    ],
```

This is not only customizable but sets a default color, that's great. But `tokenColorCustomizations` can't take a color identifier (only a color RGB code) and, moreover, there is no way of programmatically setting a color for a grammar context. So the best we can do to make wikilinks work OOTB is mapping them to some existing grammar context as I did. I think we can *also* define a more specific context like `foam.wikilink` so that themes (and the user) can customize it, but for OOTB experience we need to map to a preexisting context as well.

memeplex:
Another way to put the above is:

* We can default a new color like `foam.placeholder` to an existing color like `editorWarning.foreground`.
* We can default a new grammar scope to an existing grammar scope like `string.other.link.title.markdown`.
* We can't default a new grammar scope to an existing color like `textLink.foreground`.

This kind of makes sense.

memeplex:
Regarding the exclusion of `[[` and `]]`, it's very simple to achieve it in the grammar, but there are other issues:

1. markdown-provider is parsing `[[...]]` as a link, so its range includes the delimiters.
2. navigation-provider uses this range so vscode underlines everything, including delimiters.
3. document-decorator uses this range so it decorates everything, including delimiters.

I'm reluctant to change 1, because I expect many tests to fail after that. And I don't see an easy way to do simple position arithmetic (as in `range.start + 2, range.end - 2`) in 2 and 3.

So I would say changing this is not worth the effort.

memeplex:
I've pushed the changes I consider worthy according to the above comments.

riccardoferretti:
Thanks for the thorough research, this is very helpful.
I have a few comments, but in summary this is what I am looking to better understand: 
- how does a user customize the highlight colors for placeholders and wikilinks (I am actually comfortable with not allowing that to start with, but what I would not like is if there are two different mechanisms to achieve that result)
- what sensible defaults do we have for the highlight colors for placeholders and wikilinks (the wikilinks one I like, the `a` in `[a](b)`, I am less clear about the placeholder one)

Looking at the link you sent, I wonder if the right "place" to define placeholder/resource differences is in semantic highlights https://code.visualstudio.com/docs/getstarted/themes#_editor-semantic-highlighting - which technically speaking is exactly what we do to differentiate resources from placeholders.

> Regarding the exclusion of [[ and ]]

I agree that touching option 1 can have too many ripple effects, let's leave that the way it is 😅 
I believe option 2 and 3 should be fairly straightforward to implement, e.g. look at `wikilink-diagnostics.ts:246`, would that work?

memeplex:
> how does a user customize the highlight colors for placeholders and wikilinks

For placeholders, it's done in `workbench.colorCustomizations` by changing the color for `foam.placeholder`.

For wikilinks, it's done in `workbench.tokenColorCustomizations` by changing the color for `string.other.link.title.markdown`. Also, I realized that prefix matching works here, so I'm experimenting with the `string.other.link.title.markdown.foam` scope now, so that in inherits whatever the current theme sets as link title color but still allows to set foam specific colors.

> look at wikilink-diagnostics.ts:246, would that work?

```ts
        range: new vscode.Range(
          diagnostic.range.start.line,
          diagnostic.range.start.character + 2,
          diagnostic.range.end.line,
          diagnostic.range.end.character - 2
        ),
```

Isn't this assuming that a wikilink is entirely in the same line? Do we actually require, or are we willing to require, this?

>  I wonder if the right "place" to define placeholder/resource differences is in semantic highlights

AFAIK this is still a bit experimental and, more importantly, requires a language server.

riccardoferretti:
> Isn't this assuming that a wikilink is entirely in the same line? Do we actually require, or are we willing to require, this?

Yes it's fair to assume a wikilink is on the same line

> `workbench.colorCustomizations` ... `workbench.tokenColorCustomizations`

Darn I really don't like the different way of configuring that. The user doesn't care what's what, they just want to change a color. Maybe we can put those behind a foam setting?

memeplex:
> Darn I really don't like the different way of configuring that

Unfortunately, it's not possible to configure syntax using color ids, there is no way to do anything programmatic here. It's something we have to live with. At least it's configurable and has a sensible default (and it's difficult to see why one would change it). The alternative is to do anything through decorators and register two color ids `foam.wikilink` and `foam.placeholder`. Sure it's more consistent from a configuration perspective, but having wikilinks as part of the grammar is more performant. There are some trade offs going on here, but overall the result is positive: instant highlighting and configurable colors.

riccardoferretti:
I agree, and I think there is always time later to add those settings if things get annoying.

Where does this leave us? (also in the context of what we were saying above about consistency between placeholders and wikilinks)

memeplex:
I've implemented these features:
- Only highlight/underline contents, not delimiters.
- Make wikilink color also configurable in a foam-specific way (without affecting other links title color).

riccardoferretti (approved):
LGTM!

memeplex:
I forgot to mention that a couple days ago I read this in the texmate documentation regarding scope names:

> This name should be dot separated with each additional part specializing the kind of element matched. For example a double-quoted string in C will get string.quoted.double.c assigned as its scope name

So, in retrospective, adding the `.foam` suffix seems the right way to do it and not just a workaround.

Issue #880, opened 2021-12-17
Error logging test suite output


memeplex:
- Foam version: 0.16.0
- Platform: Max
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes

In suite.ts `stderr.write` is mapped to `console.log` but the former expects arbitrary chunks while `log` expects lines to which it appends a newline, this is the reason why there was a long sequence of characters at the end of a failing suite: jest writes a summary at the end, but in a way that sends output char by char using `write`, so `log` was adding a newline to each char. I wrote this alternative code:

```ts
  let remaining = '';
  process.stderr.write = (buffer: string) => {
    let lines = (remaining + buffer).split('\n');
    remaining = lines.pop();
    lines.forEach((l) => console.log(l));
    return true;
  };
```

and was about to create a PR when I realized that some (minor) issues still persist: random blank lines every now and then. So I started from scratch, a simple project focused on reproducing that issue (as well as the fact that stdout/err are not being shown at all) and indeed there are underlying problems in vscode-test, see my reports:

* https://github.com/microsoft/vscode-test/issues/114
* https://github.com/microsoft/vscode-test/issues/115
* https://github.com/microsoft/vscode-test/issues/116

In order to avoid the extra blank lines (not a big deal anyway) an alternative is to implement `write` as:

```ts
  let stderr = '';
  process.stderr.write = (buffer: string) => {
    stderr += buffer;
    return true;
  };
  ...
  console.log(stderr);
```

the drawback is that output is send all together at the end.

I'm going to write a PR to fix this. If you want a quick one let me know what solution do you prefer. If you're not in a hurry, then I prefer to wait for some clarification from upstream.

PR #881, opened 2021-12-20
Improve testing experience

memeplex:
Fix https://github.com/foambubble/foam/issues/880.


.yarnrc
```diff
@@ -0,0 +1 @@
+--ignore-engines true
```
memeplex, line 1: Not strictly related to testing but this suppresses some warnings and is included in the yeoman template for vscode extensions that's recommended by the vscode team.

packages/foam-vscode/jest.config.js
```diff
@@ -126,13 +126,13 @@ module.exports = {
...
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
-  // setupFilesAfterEnv: [],
+  setupFilesAfterEnv: ['jest-extended'],
 
   // A list of paths to snapshot serializer modules Jest should use for snapshot testing
   // snapshotSerializers: [],
 
   // The test environment that will be used for testing
-  testEnvironment: "node"
+  testEnvironment: 'node',
```
memeplex, line 135: This was prettier, not me, but perhaps it's better like this.

packages/foam-vscode/jest.config.js
```diff
@@ -126,13 +126,13 @@ module.exports = {
   // setupFiles: [],
 
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
-  // setupFilesAfterEnv: [],
+  setupFilesAfterEnv: ['jest-extended'],
```
memeplex, line 129: This was in every `runCLI` invocation, seems basic enough to be part of the configuration.

packages/foam-vscode/jest.config.js
```diff
@@ -91,7 +91,7 @@ module.exports = {
   // notifyMode: "failure-change",
 
   // A preset that is used as a base for Jest's configuration
-  // preset: undefined,
+  preset: 'ts-jest',
```
memeplex, line 94: Both babel-jest and ts-jest were installed. Since ts-jest seems to be the more complete and powerful option, and also it's simpler to configure, I removed babel-jest. It's pointless to have both.

packages/foam-vscode/package.json
```diff
@@ -356,7 +356,9 @@
     "build": "tsc -p ./",
     "pretest": "yarn build",
     "test": "node ./out/test/run-tests.js",
+    "pretest:unit": "yarn build",
     "test:unit": "node ./out/test/run-tests.js --unit",
+    "pretest:e2e": "yarn build",
     "test:e2e": "node ./out/test/run-tests.js --e2e",
```
memeplex, line 362: This is convenient when using `test:unit/e2e`.

packages/foam-vscode/package.json
```diff
@@ -388,12 +390,10 @@
     "@types/vscode": "^1.47.1",
     "@typescript-eslint/eslint-plugin": "^2.30.0",
     "@typescript-eslint/parser": "^2.30.0",
-    "babel-jest": "^26.2.2",
```
memeplex, line 391: Removed in favor of ts-jest as explained above.

packages/foam-vscode/package.json
```diff
@@ -388,12 +390,10 @@
...
     "eslint": "^6.8.0",
     "eslint-plugin-import": "^2.24.2",
     "husky": "^4.2.5",
     "jest": "^26.2.2",
-    "jest-environment-vscode": "^1.0.0",
```
memeplex, line 396: This was actually doing next to nothing, it just sets `globals.vscode` so I moved that to our environment (see below) and removed the dependency (moreover, it's an obscure dependency that doesn't has a public repo).

packages/foam-vscode/src/test/run-tests.ts
```diff
@@ -35,23 +33,17 @@ async function main() {
...
       const extensionTestsPath = path.join(__dirname, 'suite');
-      const tmpWorkspaceDir = fs.mkdtempSync(path.join(os.tmpdir(), 'foam-'));
 
       // Download VS Code, unzip it and run the integration test
       await runTests({
         extensionDevelopmentPath,
         extensionTestsPath,
         launchArgs: [
-          tmpWorkspaceDir,
+          path.join(extensionDevelopmentPath, '.test-workspace'),
```
memeplex, line 42: The `.test-workspace` directory is part of the repo (that's why we need the `.keep` file) but its contents are ignored by git. This avoids having to create the directory in each entry point (launcher and suite.ts).

packages/foam-vscode/src/test/run-tests.ts
```diff
@@ -35,23 +33,17 @@ async function main() {
...
           '--disable-extensions',
           '--disable-workspace-trust',
         ],
-        // Running the tests with vscode 1.53.0 is causing issues in the output/error stream management,
-        // which is causing a stack overflow, possibly due to a recursive callback.
-        // Also see https://github.com/foambubble/foam/pull/479#issuecomment-774167127
-        // Forcing the version to 1.52.0 solves the problem.
-        // TODO: to review, further investigate, and roll back this workaround.
-        version: '1.52.0',
+        version: '1.60.0',
```
memeplex, line 46: I also tried version 1.62.0 but a couple of tests were failing from time to time because of undefined values, something to revisit in the future.

packages/foam-vscode/src/test/suite-unit.ts
```diff
@@ -8,32 +8,25 @@
  *   they will make direct use of the vscode API to be invoked as commands, create editors,
  *   and so on..
  */
+
+process.env.FORCE_COLOR = '1';
+process.env.NODE_ENV = 'test';
+
```
memeplex, line 14: I moved `process.env.FORCE_COLOR = '1'` before the imports because in newer versions of foam it won't have any effect otherwise. This is due to an early import of chalk happening immediately after the import of jest (see my report https://github.com/facebook/jest/issues/12162). So this is future-proof(er). For the sake of consistency, I also moved the other environment setting here.

packages/foam-vscode/src/test/suite-unit.ts
```diff
@@ -8,32 +8,25 @@
...
 import path from 'path';
 import { runCLI } from '@jest/core';
 
 const rootDir = path.join(__dirname, '..', '..');
 
 export function runUnit(): Promise<void> {
-  process.env.FORCE_COLOR = '1';
-  process.env.NODE_ENV = 'test';
-  process.env.BABEL_ENV = 'test';
```
memeplex, line 19: I don't think we are using babel anymore (it was just a jest thing AFAICS).

packages/foam-vscode/src/test/support/vscode-environment.js
```diff
@@ -1,27 +1,27 @@
...
-    this.global.vscode = vscode;
 
     vscode.workspace
       .getConfiguration()
       .update('foam.edit.linkReferenceDefinitions', 'off');
   }
+
   async teardown() {
-    this.global.vscode = initialVscode;
+    this.global.vscode = {};
```
memeplex, line 22: The reason for this is that:

```ts
this.global.vscode = initialVscode;
await super.teardown();
```

actually ended up with `this.global.vscode = {}` since that's what the base class did during `teardown`.

Moreover, I fail to understand how can it make a difference, or how can an initial vscode global exist that's different from the current one but still somehow valid.

packages/foam-vscode/src/test/test-utils-vscode.ts
```diff
@@ -13,7 +13,7 @@ import { randomString, wait } from './test-utils';
 Logger.setLevel('error');
 
 export const cleanWorkspace = async () => {
-  const files = await vscode.workspace.findFiles('**', '.vscode');
+  const files = await vscode.workspace.findFiles('**', '{.vscode,.keep}');
```
memeplex, line 16: At the end I decided against rimraf here because I was seeing obscure errors that could be related to syncing issues between the underlying fs and vscode's view of it. And, anyway, it wasn't that easy to remove folders using glob patterns.

packages/foam-vscode/src/test/suite-unit.ts
```diff
@@ -8,32 +8,25 @@
...
-          transform: JSON.stringify({ '^.+\\.ts$': 'ts-jest' }),
           runInBand: true,
           testRegex: '\\.(test)\\.ts$',
           setupFiles: ['<rootDir>/src/test/support/jest-setup.ts'],
-          setupFilesAfterEnv: ['jest-extended'],
-          globals: JSON.stringify({
-            'ts-jest': {
-              tsconfig: path.join(rootDir, 'tsconfig.json'),
-            },
-          }),
```
memeplex, line 36: The ts-jest related settings doesn't seem to be necessary, the official documentation suggests to just set the preset to ts-jest. I did that and found no issue.

packages/foam-vscode/src/test/suite.ts
```diff
@@ -9,42 +9,43 @@
...
+
+  let remaining = '';
   process.stderr.write = (buffer: string) => {
-    console.log(buffer);
+    const lines = (remaining + buffer).split('\n');
+    remaining = lines.pop() as string;
+    // Trim long lines because some uninformative code dumps will flood the
+    // console or, worse, be suppressed altogether because of their size.
+    lines.forEach(l => console.log(l.substr(0, 300)));
     return true;
```
memeplex, line 31: This triggered a number of related reports:

* https://github.com/microsoft/vscode-test/issues/114
* https://github.com/microsoft/vscode-test/issues/115
* https://github.com/microsoft/vscode-test/issues/116
* https://github.com/microsoft/vscode/issues/139417
* https://github.com/foambubble/foam/issues/880

.vscode/launch.json
```diff
@@ -6,15 +6,18 @@
...
+      "args": [
+        "${workspaceFolder}/packages/foam-vscode/.test-workspace",
+        "--disable-extensions",
+        "--disable-workspace-trust",
+        "--extensionDevelopmentPath=${workspaceFolder}/packages/foam-vscode",
+        "--extensionTestsPath=${workspaceFolder}/packages/foam-vscode/out/test/suite"
+      ],
+      "outFiles": ["${workspaceFolder}/packages/foam-vscode/out/**/*.js"],
+      "preLaunchTask": "${defaultBuildTask}"
     },
```
memeplex, line 21: An `extensionHost` launcher is the way to go: https://github.com/microsoft/vscode-test/issues/117

packages/foam-vscode/package.json
memeplex: On a side note, calling `yarn test` at the top level is noticeably slower than calling it here, I don't if that's because of lerna or what.

.vscode/launch.json
```diff
@@ -24,8 +29,25 @@
...
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "cwd": "/Users/carlos/Desktop/foam/packages/foam-vscode",
+      "runtimeExecutable": "yarn",
+      "args": [
+        "jest",
+        "--runInBand",
+        "--watchAll=false"
+      ]
     }
```
memeplex, line 51: This launcher is for debugging from the test explorer (sidebar) when using the vscode-jest extension.

packages/foam-vscode/jest.config.js
```diff
@@ -82,7 +82,7 @@ module.exports = {
   // moduleNameMapper: {},
 
   // An array of regexp pattern strings, matched against all module paths before considered 'visible' to the module loader
-  // modulePathIgnorePatterns: [],
+  modulePathIgnorePatterns: ['.vscode-test'],
```
memeplex, line 85: Otherwise vscode-jest will try to resolve some imports there and fail because of ambiguity (e.g. the yaml module is in every vscode installation).

packages/foam-vscode/jest.config.js
```diff
@@ -152,7 +152,7 @@ module.exports = {
   // ],
 
   // The regexp pattern or array of patterns that Jest uses to detect test files
-  // testRegex: [],
+  testRegex: ['\\.test\\.ts$'],
```
memeplex, line 155: This is overridden in every `runCLI` invocation but it's here as the default for vscode-jest (since there is no extension specific setting for this AFAICS). We only want unit tests in the test explorer (sidebar), since spec tests require the entire extension host to be launched before.

.yarnrc
riccardoferretti: 👍  great, I didn't like that warning, but didn't know how to suppress it

.vscode/launch.json
```diff
@@ -24,8 +29,25 @@
...
       "preLaunchTask": "${defaultBuildTask}"
+    },
+    {
+      "type": "node",
+      "name": "vscode-jest-tests",
+      "request": "launch",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "cwd": "/Users/carlos/Desktop/foam/packages/foam-vscode",
```
riccardoferretti, line 44:
```suggestion
      "cwd": "${workspaceFolder}/packages/foam-vscode",
```

packages/foam-vscode/jest.config.js
riccardoferretti: Good call

packages/foam-vscode/jest.config.js
riccardoferretti: This is a very important point, above in the `vscode-jest-tests` I was about to ask exactly whether it was supporting both unit and integration tests.
I think we need to either highlight this connection, or better yet make the `vscode-jest-tests` self-explanatory:
- we could add this comment somewhere in the code, to "link" the `vscode-jest-tests` to this configuration, or
- we could have this setting as part of the `vscode-jest-tests` configuration, so that it doesn't have a dependency on this external configuration option
> since there is no extension specific setting for this AFAICS

Could you use the `args` param? Or maybe that's only used at runtime so the extension would still "flag" e2e tests as runnable in VS Code?

What do you think?

packages/foam-vscode/package.json
riccardoferretti: Interesting, I guess one more reason to flatten things out

packages/foam-vscode/src/test/run-tests.ts
riccardoferretti: Should we add a comment above this, to explain the reasoning around locking in the version, including the double testing done by running the tests locally (with the VS Code version matching the active one), and in CI (with the locked in version, for test predictability)?
This would help new devs possibly make sense of discrepancy in test results across environments.

packages/foam-vscode/src/test/suite-unit.ts
riccardoferretti: Good catch! I would add this comment in the code, for future reference.

packages/foam-vscode/src/test/suite.ts
riccardoferretti: What do you mean by triggered?

riccardoferretti (commented):
This looks really good, I have left a couple of questions and documentation recommendations, let me know what you think

packages/foam-vscode/src/test/run-tests.ts
memeplex: TBH any comment beyond "this is the latest version that worked without issues for me" would be a lie. This may depend on the version of vscode and the version of every other dependency, besides foam code itself, so I think it's for the best to assume any upgrade may break things and leave the resolution of that breakage to the upgrader. That's the reason why I suggested to do this in a "controlled fashion".

packages/foam-vscode/src/test/suite.ts
memeplex: These are all reports of mine, some of them related to what caused this change, some of them to what was caused by this change. My feeling is that vscode tooling is still somewhat of work in progress in this regard.

packages/foam-vscode/jest.config.js
memeplex: > `vscode-jest-tests`

Notice that the extension I'm using is `vscode-jest`, this one: https://github.com/jest-community/vscode-jest

>  I was about to ask exactly whether it was supporting both unit and integration tests.

The sidebar explorer offers to run each test separately. In principle, we could write a script that launches the entire extension host just to run some e2e test and register this script as the `vscode-jest-tests` launcher, but it doesn't seem very useful because of the high overhead of creating the environment. Moreover, the launcher would be of type `node` (I don't think `vscode-jest` is able to run other types), so really unsupported as a way of debugging a test running inside the host.

The use case of quickly running some unit test from the sidebar is convenient and simple enough to support, so I'm focusing on that.

> we could add this comment somewhere in the code, to "link" the vscode-jest-tests to this configuration

I've added the comment to `jest.config.js`.

> we could have this setting as part of the vscode-jest-tests configuration

This was the first thing I tried, but unfortunately I found no such setting.

> Could you use the args param? 

No, the launcher is used to run a test but not to build the sidebar tree.

packages/foam-vscode/src/test/suite-unit.ts
memeplex: Done.

riccardoferretti (approved):
This is massive, thanks for improving DX around testing

PR #882, opened 2021-12-20
Vendor a snippet parser

### LINKED ISSUE #602: Text elements that collide with Foam-specific variables should not be replaced
- Foam version: 0.13.1
- Platform: all

**Summary**

1. Create a template with the following content:

```
# ${CURRENT_DATE/.*/${FOAM_TITLE}/}
```

2. Run `Foam: Create New Note From Template`. Select the new template
3. Enter a file name

The file contents under all circumstances should be `# ${FOAM_TITLE}` ([variables aren't allowed as the last segment of a transform](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_grammar), so `${FOAM_TITLE}` must be a `text` string).

However the file contents are instead the user-supplied title.

**Additional information**

This is a result of doing a simplified replacement of the Foam-specific variables in a first pass before passing the resulting snippet to VSCode. Ref: #549 

The snippet parser in Foam needs to be aware enough to know not to replace `text` that look like Foam-specific variables.

### LINKED ISSUE #693: [Templates v2] Allow Foam-specific variables to use the full Snippet variable syntax
In #549, we added the `FOAM_TITLE` snippet variable.

However, it could only handle the `$FOAM_TITLE` and `${FOAM_TITLE}` forms, which is only a subset of the [VSCode snippet syntax grammar](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_grammar).

For example, one cannot make use of the [variable transforms](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variable-transforms) (e.g. `${FOAM_TITLE/[^0-9^a-z]//gi}`)

---

Figure out how to support this. It might be through:

* copying bits of [VS Code's snippetParser](https://github.com/microsoft/vscode/blob/2aa0f0f428524aedb3b22b7bf8e4f291708017a9/src/vs/editor/contrib/snippet/snippetParser.ts) 
* getting VS Code to allow for custom resolvers to be registered

movermeyer:

Fixes #693
Fixes #602

Extending our understanding of the snippet syntax to clean up the edge cases.

I started by vendoring the [`SnippetParser` code](https://github.com/microsoft/vscode/blob/ea132060dd8c71baa6b2f77e46287710b888bb01/src/vs/editor/contrib/snippet/snippetParser.ts) used in VSCode.
They are [not going to allow extensions to register their own snippet variables](https://github.com/microsoft/vscode/issues/130325#issuecomment-899338950), so we need to have our own parser.

I then used their parser to find the Foam variables. This allows us to remove the cludgy regex we were previously using to find Foam variables. However, it didn't allow us to remove the regex being used to substitute values, so #602 wasn't yet solved.

I then changed `VariableResolver` interface to be asynchronous, so that resolvers can do async compute (e.g., `FOAM_TITLE` needs to ask the user for input), and moved the variable resolve code into a new `VariableResolver`. This allowed us to resolve the variables in the same way that VSCode does, including default values and transforms, solving #693.

But without a way to substitute the values, I hadn't accomplished anything. I then changed `SnippetParser` to return start and end indices for the `Variable`s in the snippet. `substituteVariables` was then changed to do the substitution based on the `Variable`s indices, instead of the regex.

I may be easier to review this commit-by-commit.

This is not the end state for this code. There is still #692 to solve.
That will require another large refactor, but this is a step in the right direction, and solves the edge case bugs while I figure out the best way to solve the next step.

#### Open questions

Where do you want this snippet parser code? I dumped it into `packages/foam-vscode/src/core/common/snippetParser.ts` as per [this comment](https://github.com/foambubble/foam/pull/882#discussion_r773054079)

Fixes #693
Fixes #602

### Testing

...


packages/foam-vscode/src/services/variable-resolver.ts
```diff
@@ -271,6 +203,111 @@ export class Resolver {
     return result;
   }
 }
+export class Resolver {
```
riccardoferretti, line 206: Have you split the `FoamVariableResolver` from the `Resolver` to make the VS Code API happy?

I am trying to wrap my mind around the individual responsibilities and how tied they are

riccardoferretti (commented):
I like the approach, will be great to have full support for VS Code variables/placeholders.
I left a few comments below

packages/foam-vscode/src/services/variable-resolver.spec.ts
```diff
@@ -17,6 +18,18 @@ describe('substituteFoamVariables', () => {
     expect((await resolver.resolveText(input))[1]).toEqual(input);
   });
 
+  test('Ignores variable-looking text values', async () => {
```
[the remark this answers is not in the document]
movermeyer: Sure thing. Done.

packages/foam-vscode/src/utils/snippets/charCode.ts
```diff
@@ -0,0 +1,444 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
movermeyer: Moved to `src/core/common`.

packages/foam-vscode/src/utils/snippets/snippetParser.ts
```diff
@@ -0,0 +1,1088 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
```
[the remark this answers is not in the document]
movermeyer: Done. I've changed the header to mention the SHA it was taken from, and include the MIT license.

packages/foam-vscode/src/services/variable-resolver.ts
movermeyer: > to make the VS Code API happy?

Not explicitly. I did it for convenience, since that way I didn't have to change the consumers of `Resolver`. Probably worth a revisit to see if it makes sense to merge the two.

movermeyer: Alright, that was easy. Just had to rename `resolve` to `resolveForName`.

Done 👍

riccardoferretti (commented):
Looking good so far, let me know when ready for review!

packages/foam-vscode/src/services/variable-resolver.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    const variable = new Variable('FOAM_TITLE', 0, 13);
+    variable.children.push(new Text('My new note'));
+
+    const variables = [variable];
+    const expected = 'My new note';
+
+    expect(substituteVariables(input, variables)).toEqual(expected);
+  });
+
+  test('it ignores Variables that are missing positions', () => {
```
riccardoferretti, line 113: I am a bit confused here, what do you mean? 
Is the fact that the position is missing in the variable (line 120) that means it doesn't get expanded, or that the variable is not present in the text (line 115)?
What would happen if line 115 were: `const input = '\${FOAM_DATE_MONTH}\n \${FOAM_TITLE}';`?

packages/foam-vscode/src/services/variable-resolver.ts
```diff
@@ -19,42 +26,40 @@ const knownFoamVariables = new Set([
...
-      // longer variable names with the values of variables that are
-      // substrings of the longer ones (e.g. `$FOO` and `$FOOBAR`. If you
-      // replace $FOO first, and aren't careful, you replace the first
-      // characters of `$FOOBAR`)
-      'g' // 'g' => Global replacement (i.e. not just the first instance)
-    );
-    text = text.replace(regex, `${value}$1`);
-  });
+export function substituteVariables(text: string, variables: Variable[]) {
+  // Assumes/requires that the variables are sorted by start position, and non-overlapping
```
riccardoferretti, line 30: Sorting them ourselves doesn't feel like a big overhead, could we do it instead of assuming that's the case?

packages/foam-vscode/src/services/variable-resolver.ts
```diff
@@ -88,28 +93,44 @@ export class Resolver {
...
+
+    const uniqVariableNamesInTemplate = new Set(
+      variablesInTemplate.map(variable => variable.name)
     );
-    const uniqVariables = [...new Set(variables)];
 
-    const resolvedValues = await this.resolveAll(uniqVariables);
+    const variables = variablesInTemplate.concat(
+      [...this.extraVariablesToResolve]
+        .filter(name => !uniqVariableNamesInTemplate.has(name))
```
riccardoferretti, line 103: should this filtering be applied to the concat'ed array?

packages/foam-vscode/src/services/variable-resolver.test.ts
riccardoferretti: Upon finishing the review, this is my understanding: a `Variable` without position can be resolved, but cannot be replaced in some text. 

Is this correct?

If so, I reckon the reason is that we need the position to know what to replace (accounting for placeholders, transform, ...), which is too hard to find at text replacement time?

I find that possibly confusing, I am trying to understand how much we can work around it

packages/foam-vscode/src/core/common/snippetParser.ts
```diff
@@ -0,0 +1,1127 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
movermeyer: I used to have a commit on this PR that excluded these files from the linter.
Then https://github.com/foambubble/foam/pull/887 got merged with a change that excluded this directory from linting: 

https://github.com/foambubble/foam/blob/c8f1f8e03ae917dc194ccff54d20c20c2df9a0e1/.eslintrc.json#L42

Personally, I like to keep vendored code as close to the original as possible, to make it easier to bring in bugfixes/changes from the upstream. Happy to remove that `ignorePatterns` entry and reformat this code if you think it's best.

packages/foam-vscode/src/core/common/snippetParser.test.ts
```diff
@@ -0,0 +1,811 @@
...
+ *	LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *	OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *	SOFTWARE.
+ *
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { Choice, FormatString, Marker, Placeholder, Scanner, SnippetParser, Text, TextmateSnippet, TokenType, Transform, Variable } from './snippetParser';
+
+describe('SnippetParser', () => {
```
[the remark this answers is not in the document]
movermeyer: Nope. 🙂

packages/foam-vscode/src/core/common/snippetParser.ts
```diff
@@ -0,0 +1,1127 @@
...
+	resolve(value?: string): string {
+		if (this.shorthandName === 'upcase') {
+			return !value ? '' : value.toLocaleUpperCase();
+		} else if (this.shorthandName === 'downcase') {
+			return !value ? '' : value.toLocaleLowerCase();
+		} else if (this.shorthandName === 'capitalize') {
+			return !value ? '' : (value[0].toLocaleUpperCase() + value.substr(1));
+		} else if (this.shorthandName === 'pascalcase') {
+			return !value ? '' : this._toPascalCase(value);
+		} else if (this.shorthandName === 'camelcase') {
```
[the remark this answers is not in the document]
movermeyer: > I wonder if expanding on this approach we could also support the slugging of the `FOAM_TITLE` via a `shorthandName === 'slug'` (and its proper handling).
> Unrelated to this PR per se, but affects #865 as in that case we wouldn't really need a `FOAM_TITLE_SLUG` variable. 

Definitely an idea worth exploring. 👍 

> I also don't understand why VS Code itself doesn't support this, given it supports `camelCase` and `PascalCase`

I suspect its beacuse `slug` isn't an agreed upon standard. Each project implements their own algorithm for "slugification". For example, check out the 6 different ways that this Markdown extension needs to handle in order to keep users happy:

https://github.com/yzhang-gh/vscode-markdown/blob/c572cb4a981272f0b3cb3570b60766b26cbe6aab/src/util/slugify.ts#L46-L140

Perhaps they could choose the VSCode implementation to be the one. IDK. But I suspect that people would complain and want their favorite tool's slugification added too.

packages/foam-vscode/src/services/variable-resolver.ts
movermeyer: I briefly had a `sort` here, then removed it.

They shouldn't ever be out of order. If it were to ever be a problem, I think I'd rather fix the source of the issue rather than quietly hide it here.

I could remove this comment if you felt that it was drawing too much attention to this requirement.

packages/foam-vscode/src/services/variable-resolver.ts
movermeyer: I don't think so?

`uniqVariableNamesInTemplate` is derived from `variablesInTemplate`, so the filtering would do nothing?

packages/foam-vscode/src/services/variable-resolver.test.ts
movermeyer: > I am a bit confused here, what do you mean? 
> Is the fact that the position is missing in the variable (line 120) that means it doesn't get expanded, or that the variable is not present in the text (line 115)? 

Both. It isn't present in the text, so doesn't get expanded.
This test is just ensuring that `substituteVariables` can handle Variables that do not have positions (because they aren't in the text).

This can happen when variables are given as `extraVariablesToResolve`. For convenience, the caller of `substituteVariables` doesn't have to know anything about `Variables` and doesn't have to filter these out before calling `substituteVariables`.

> What would happen if line 115 were: `const input = '\${FOAM_DATE_MONTH}\n \${FOAM_TITLE}';`?

Then this case wouldn't have happened. The snippet parser would have found `FOAM_TITLE` and would have given it a position.

> Upon finishing the review, this is my understanding: a Variable without position can be resolved, but cannot be replaced in some text.
> Is this correct?

Yes. A `Variable` without a position isn't actually present in the text, so both cannot and should not have any impact on the resulting text.

> I reckon the reason is that we need the position to know what to replace (accounting for placeholders, transform, ...), which is too hard to find at text replacement time?

Our simple regex that we previously had is not powerful enough to correctly identify the bound of a variable in the text. Now that the snippet parser gives us the start and end positions, we know where to replace.

> I find that possibly confusing, I am trying to understand how much we can work around it

Happy to chat more.

packages/foam-vscode/src/core/common/snippetParser.ts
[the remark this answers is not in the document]
riccardoferretti: The reason for the ignore patterns is that there are some "more complex" linting rules that would be too much effort to fix, requiring actual change to the code that goes beyond reformatting.
But I see your point about checking against upstream, let's keep the code this way then, if we want to reformat it we can do it later

packages/foam-vscode/src/services/variable-resolver.ts
riccardoferretti: Sounds good, I don't feel strongly about this, happy to go with your recommendation and I agree that as this is the expected behavior, we should address out of order issues where they originate.

I have a minor comment here, that I think is what caused me to feel confused about this function.
The two parameters are intrinsically coupled (as the positions that live in the variables refer to the text), I wonder if this coupling should be more explicit.
Or maybe there should actually be a distinction between a `Variable`, which is just a key/value pair, and a `Substitution`, which is a `Variable` with the added metadata "position" that relates to how to apply it to a text.

E.g.
```
export function substituteVariables(text: ResolvedText) {
   ...
}

interface ResolvedText {
  template: string
  substitutions: Substitution[]
}

interface Substitution {
  variable: Variable
  pos: number
  endPos: number
}
```

Just brainstorming here. Maybe even a simpler change/documentation to the API would do the trick.

But my point is that the `Variable` atm feels like is covering two roles (key/value pair, and substitution in the context of a string), I wonder if splitting them would make some things clearer?

Wdyt?

movermeyer: @riccardoferretti Makes sense. I've made two changes that clean this up a bunch.

----

1. I removed `extraVariablesToResolve`. It was not needed anymore:
  * `FOAM_TITLE` was now being [resolved separately](https://github.com/foambubble/foam/blob/08d305b9b0c16508ca226a4d82bcf3bbcea40240/packages/foam-vscode/src/services/templates.ts#L257) in `determineNewNoteFilepath`
  * The presence of `FOAM_SELECTED_TEXT` in `extraVariablesToResolve` was being used to determine whether to append `FOAM_SELECTED_TEXT` to the template. But it could equally use the existence of `FOAM_SELECTED_TEXT` in `givenValues`, so I switched to use that.
  
This means that the confusion around "`Variable`s that weren't present in the template" has been removed, since they no longer exist.

----

> The two parameters are intrinsically coupled (as the positions that live in the variables refer to the text), I wonder if this coupling should be more explicit.

2. I moved the handling of the template text and its variable substitution into `TextmateSnippet`.

This means that `variable-resolver.ts` no longer has to know anything about how to find/substitute variables in a template, and `resolveText` can be simpler too.

One downside of this is that there is now non-trivial changes (new methods) in the vendored VSCode code, making it slightly harder to update when you want to refresh the code from the upstream. 🤷

----

I think it's ready for another round of review.

riccardoferretti: > I removed extraVariablesToResolve. It was not needed anymore:

I like that very much 👍

> I moved the handling of the template text and its variable substitution into TextmateSnippet.

The approach makes a lot of sense. Even though `TextmateSnippet` is very much stateful (which I don't love) it is in line with how that class works, so in the end the solution works really well.

> One downside of this is that there is now non-trivial changes (new methods) in the vendored VSCode code, making it slightly harder to update when you want to refresh the code from the upstream. 🤷

Good point, but I agree that it's a good tradeoff.


I don't have any other comment on the code, for me it's good to be merged.
Is there anything I am missing or that you wanted to add here?

movermeyer: :shipit: :ship: :shipit:

riccardoferretti (approved):

PR #885, opened 2021-12-22
Fix link autocompletion with tags

### LINKED ISSUE #884: Autocomplete stops working when there's a # symbol in the same line
- Foam version: v0.17.1
- Platform: Mac (macOS Big Sur)
- Issue occurs on the [foam template](https://github.com/foambubble/foam-template) repo: Yes!

**Summary**

When there's a #tag in the same line, the autocomplete stops working. I think this started happening just a few weeks ago.

**Steps to reproduce**

1. Add a #tag to the line
2. Keep writing and try to autocomplete a [[wikilink]]

I was able to reproduce it using the [foam template](https://github.com/foambubble/foam-template).

**Additional information**

The only VSCode extensions I run are Vim, Prettier, and GitGraph.

Screen Capture on 2021-12-22 at 10-48-59

memeplex:
Fix #884

Alternative PR: #886

### DIFF
--- a/packages/foam-vscode/src/features/link-completion.spec.ts
+++ b/packages/foam-vscode/src/features/link-completion.spec.ts
@@ -88,40 +88,48 @@ describe('Link Completion', () => {
   });
 
   it('should return notes with unique identifiers, and placeholders', async () => {
-    const { uri } = await createFile('[[file]] [[');
-    const { doc } = await showInEditor(uri);
-    const provider = new CompletionProvider(ws, graph);
+    for (const text of ['[[', '[[file]] [[', '[[file]] #tag [[']) {
+      const { uri } = await createFile(text);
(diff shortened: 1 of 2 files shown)


memeplex:
@riccardoferretti what do you think of simplifying the regular expressions a bit, like this:

```ts
export const WIKILINK_REGEX = /\[\[[^[\]]*$/;
export const SECTION_REGEX = /\[\[([^[\]]*#)[^[\]]*$/;
```

Instead of the negative lookahead I'm requesting a prefix that starts with `[[` and extends to the cursor position without any `[` or `]` in between.

riccardoferretti (approved):
Nice and clean. Can we add a test case for it?

memeplex:
> Can we add a test case for it?

I've added some ones.

PR #887, opened 2021-12-23
Improve linting

memeplex:
The linter is showing some warnings about skipped / commented out tests, I think these are valuable so I didn't disable them.

Most of the changes are trivial `let -> const` substitutions.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,14 +5,38 @@
     "ecmaVersion": 6,
     "sourceType": "module"
   },
-  "plugins": ["@typescript-eslint", "import"],
+  "env": { "node": true, "es6": true },
+  "plugins": ["@typescript-eslint", "import", "jest"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:import/recommended",
+    "plugin:import/typescript",
+    "plugin:jest/recommended"
+  ],
   "rules": {
-    "@typescript-eslint/class-name-casing": "warn",
-    "@typescript-eslint/semi": "warn",
-    "curly": "warn",
-    "eqeqeq": "warn",
-    "no-throw-literal": "warn",
-    "semi": "off",
-    "require-await": "warn"
-  }
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "import/no-extraneous-dependencies": [
+      "error",
+      {
+        "devDependencies": ["**/src/test/**", "**/src/**/*{test,spec}.ts"]
+      }
+    ]
+  },
+  "settings": {
+    "import/core-modules": ["vscode"],
+    "import/parsers": {
+      "@typescript-eslint/parser": [".ts", ".tsx"]
+    },
+    "import/resolver": {
+      "typescript": {
+        "alwaysTryTypes": true
+      }
+    }
+  },
+  "ignorePatterns": ["**/core/common/**", "*.js"]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,9 +17,10 @@
   "rules": {
     "no-use-before-define": "off",
     "@typescript-eslint/no-use-before-define": "off",
-    "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/interface-name-prefix": "off",
     "import/no-extraneous-dependencies": [
       "error",
       {
@@ -38,5 +39,6 @@
       }
     }
   },
-  "ignorePatterns": ["**/core/common/**", "*.js"]
+  "ignorePatterns": ["**/core/common/**", "*.js"],
+  "reportUnusedDisableDirectives": true
 }
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -394,10 +394,6 @@
     "publish-extension": "yarn publish-extension-vscode && yarn publish-extension-openvsx && yarn npm-cleanup"
   },
   "devDependencies": {
-    "@babel/core": "^7.11.0",
-    "@babel/plugin-transform-runtime": "^7.10.4",
-    "@babel/preset-env": "^7.11.0",
-    "@babel/preset-typescript": "^7.10.4",
     "@types/dateformat": "^3.0.1",
     "@types/glob": "^7.1.1",
     "@types/lodash": "^4.14.157",
@@ -410,7 +406,9 @@
     "@typescript-eslint/eslint-plugin": "^2.30.0",
     "@typescript-eslint/parser": "^2.30.0",
     "eslint": "^6.8.0",
+    "eslint-import-resolver-typescript": "^2.5.0",
     "eslint-plugin-import": "^2.24.2",
+    "eslint-plugin-jest": "^25.3.0",
     "husky": "^4.2.5",
     "jest": "^26.2.2",
     "jest-extended": "^0.11.5",
--- a/packages/foam-vscode/src/core/utils/log.ts
+++ b/packages/foam-vscode/src/core/utils/log.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
 export interface ILogger {
   debug(message?: any, ...params: any[]): void;
   info(message?: any, ...params: any[]): void;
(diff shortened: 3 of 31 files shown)


packages/foam-vscode/src/features/navigation-provider.spec.ts
```diff
@@ -232,6 +232,5 @@ describe('Document navigation', () => {
         range: new vscode.Range(0, 23, 0, 23 + 9),
       });
     });
-    it('should provide references for placeholders', async () => {});
```
riccardoferretti, line 235: I would still keep the test case (not sure why it wasn't implemented), maybe `it.todo`?

For me a test case serves 2 purposes:
- tests that the code is actually working
- provides documentation to (present and) future developers about the intended specs of a feature

Sometimes even if a test is convoluted to implement, it might still be worth keeping it as some sort of documentation, although I think those would need to be flagged properly.
I still consider this an edge case (again, not sure what happened here), but an edge case we should account for.

What do you think?

packages/foam-vscode/src/test/suite.ts
```diff
@@ -38,6 +39,7 @@ export function run(): Promise<void> {
   //   throw err;
   // });
 
+  // eslint-disable-next-line no-async-promise-executor
```
riccardoferretti, line 42: TIL https://eslint.org/docs/rules/no-async-promise-executor

I have actually tried to make the `run` function `async` and remove the `Promise` wrapper, and things ran well.. I wonder why this was done in the first place, but maybe instead of ignoring the rule we should remove the promise, wdyt?

packages/foam-vscode/src/features/tags-tree-view.spec.ts
```diff
@@ -94,7 +95,9 @@ describe('Tags tree panel', () => {
 
     childTreeItems.forEach(child => {
       if (child instanceof TagItem) {
+        // eslint-disable-next-line jest/no-conditional-expect
```
riccardoferretti, line 98: I see why this rule is helpful, if we changed the instance of `child` this test wouldn't fail, but would silently skip any check.. maybe we should actually remove the `instanceof` and find a way to incorporate that in the test

packages/foam-vscode/package.json
```diff
@@ -394,10 +394,6 @@
     "publish-extension": "yarn publish-extension-vscode && yarn publish-extension-openvsx && yarn npm-cleanup"
   },
   "devDependencies": {
-    "@babel/core": "^7.11.0",
```
riccardoferretti, line 397: this makes me happy

riccardoferretti (commented):
I like the idea, I have left a few comments for you

riccardoferretti:
Also, are the lint warnings surfaced in VS Code as well?

I would like to set things up in such a way so that we don't need to discover linting warnings in CI, but we "run into them" in DEV. Any ideas around that?

packages/foam-vscode/src/core/utils/log.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
```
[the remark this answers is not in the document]
memeplex: If we enable the rule, it's all about consistency, so it has two modes: use the `I` prefix always or never.

> could actually be replaced by a combo of interface Logger + abstract Class Logger

I think the class declaration will merge with the type defined by the interface declaration. It's not as if they lived in different namespaces (the namespace of interfaces and the namespace of classes). So AFAICS `interface Logger + abstract Class Logger` and `interface ILogger + abstract Class Logger` will have different implications for the type checker. For example:

```ts
interface C {
  a: number;
  
  b: string;
}

abstract class C {
  a: number;

  b: number;  // <-- Error: Subsequent property declarations must have the same type
}
```

or 

```ts
interface C {
  a: number;

  b: string;
}

abstract class C {
  a: number;

  b: string;

  c: number;
}

function f(c: C): C {
  return c;
}

f({ a: 0, b: '1' }); // <-- Error: c is missing
```

I've not found this strictly documented anywhere, but my understanding is that the class declaration will create both the runtime object and a compile time type that will merge with an existing type named the same, so changing the type created by the interface declaration as well.

Given that the rule is about consistency I see no point in disabling it on a case by case basis like I did. And refactoring will require coming up with different names for classes and interfaces. So let's disable this rule altogether?

packages/foam-vscode/src/features/navigation-provider.spec.ts
memeplex: I'm going to comment it so that we get a warning.

packages/foam-vscode/src/test/suite.ts
memeplex: Same comment than before:

Yes, I'd like to change it, but I'm trying to avoid non-trivial refactors in this PR. That's why I've disabled some failing checks. I think it's an overall gain anyway: in the future we will be better advised not to do some things, while we still have a mark (we can search for eslint-disable) for previous violations of the rule. Maybe I can locally make this a warning instead of disabling it, I'm not sure if that's possible.

packages/foam-vscode/src/features/tags-tree-view.spec.ts
memeplex: Yes, I'd like to change it, but I'm trying to avoid non-trivial refactors in this PR. That's why I've disabled some failing checks. I think it's an overall gain anyway: in the future we will be better advised not to do some things, while we still have a mark (we can search for eslint-disable) for previous violations of the rule. Maybe I can locally make this a warning instead of disabling it, I'm not sure if that's possible.

packages/foam-vscode/package.json
memeplex: :)

packages/foam-vscode/src/features/tags-tree-view.spec.ts
memeplex: > Maybe I can locally make this a warning instead of disabling it, I'm not sure if that's possible.

It's only possible using a cumbersome syntax that will require to first turn the rule into a warning and then, in another line, revert that setting. I think it's better to keep the oneliner.

packages/foam-vscode/src/test/suite.ts
memeplex: It's only possible using a cumbersome syntax that will require to first turn the rule into a warning and then, in another line, revert that setting. I think it's better to keep the oneliner

packages/foam-vscode/src/core/utils/log.ts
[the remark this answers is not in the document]
riccardoferretti: Sorry my example of `URI` is no longer valid, I was referring to it before the refactoring, where the `abstract class` only has static methods.

A good, current example is `Resource`

.eslintrc.json
```diff
@@ -5,14 +5,38 @@
...
-    "eqeqeq": "warn",
-    "no-throw-literal": "warn",
-    "semi": "off",
-    "require-await": "warn"
-  }
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-non-null-assertion": "off",
```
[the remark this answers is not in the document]
riccardoferretti: Sounds good, if we find one valuable we can turn it on and have a PR (per rule) to adjust the codebase

.eslintrc.json
```diff
@@ -5,14 +5,38 @@
...
+    "import/parsers": {
+      "@typescript-eslint/parser": [".ts", ".tsx"]
+    },
+    "import/resolver": {
+      "typescript": {
+        "alwaysTryTypes": true
+      }
+    }
+  },
+  "ignorePatterns": ["**/core/common/**", "*.js"]
```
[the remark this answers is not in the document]
riccardoferretti: > The ES6-style modules requirement was bothering me

What do you mean?

riccardoferretti (approved):
This looks good and ready to be merged.

I left a couple of comments, that are informational and don't require any action. 

Specifically, check the `ILogger` one, basically a better example of what I was trying to say is in `Resource` (defined in `note.ts`)

packages/foam-vscode/src/core/utils/log.ts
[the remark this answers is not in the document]
memeplex: Yes, I assumed you referred to the previous implementation. But my point is that we cannot just rename `ISomething` to `Something` without care, because any other `Something` class in the same namespace will suddenly clash with the interface. Both classes and interfaces will be adding to the same underlying type `Something`. And since I'm deliberately avoiding thoughtful refactorings here, I'm prone to disabling the rule.

memeplex:
> Also, are the lint warnings surfaced in VS Code as well?
> 
> I would like to set things up in such a way so that we don't need to discover linting warnings in CI, but we "run into them" in DEV. Any ideas around that?

Sorry I forgot to answer this one. I think all that's required is to enable the eslint extension for the workspace.

.eslintrc.json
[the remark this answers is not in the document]
memeplex: That the default config wants you to change `require(...)` stuff to `import ...` equivalents, so it's conflicting with our spare usage of javascript.

PR #890, opened 2021-12-23
Some link highlighting improvements

memeplex:


packages/foam-vscode/package.json
```diff
@@ -261,7 +261,7 @@
           "default": true
         },
         "foam.decorations.links.enable": {
-          "description": "Enable decorations for links",
+          "description": "Enable decorations for placeholders",
```
memeplex, line 264: I feel inclined to remove this setting.

packages/foam-vscode/src/features/navigation-provider.ts
```diff
@@ -125,7 +125,14 @@ export class NavigationProvider
...
     const result: vscode.LocationLink = {
-      originSelectionRange: toVsCodeRange(targetLink.range),
+      originSelectionRange: toVsCodeRange(
+        new vscode.Range(
+          targetLink.range.start.line,
+          targetLink.range.start.character + 2,
+          targetLink.range.end.line,
+          targetLink.range.end.character - 2
+        )
+      ),
```
memeplex, line 135: Otherwise pressing Ctrl/Cmd when the cursor is over a wikilink will underline the entire `[[...]]` expression, including delimiters.

packages/foam-vscode/syntaxes/injection.json
```diff
@@ -5,7 +5,13 @@
...
       "contentName": "string.other.link.title.markdown.foam",
       "begin": "\\[\\[",
-      "end": "\\]\\]"
+      "beginCaptures": {
+        "0": { "name": "punctuation.definition.metadata.markdown" }
+      },
+      "end": "\\]\\]",
+      "endCaptures": {
+        "0": { "name": "punctuation.definition.metadata.markdown" }
+      }
```
memeplex, line 14: This is what the delimiters in `[..](...)` are.

packages/foam-vscode/src/features/navigation-provider.ts
riccardoferretti: what about direct links? I think this code is used for both

packages/foam-vscode/syntaxes/injection.json
riccardoferretti: basically you are making sure that `[[` and `[` use the same style, am I reading this right?

packages/foam-vscode/src/features/navigation-provider.ts
memeplex: What do you mean? The "Go to definition" feature?

riccardoferretti: `targetLink.range` for a wikilink includes `[[` and `]]`
`targetLink.range` for a direct link includes everything from `[` to `)`

packages/foam-vscode/syntaxes/injection.json
memeplex: `punctuation.definition.metadata.markdown` is the scope for `[`, `]`, `(` and `)` in `[..](...)`. And I'm adding the `.foam` suffix so that it can be configured for foam in a different way if so desired.

packages/foam-vscode/package.json
riccardoferretti: Actually, I am pretty happy to remove it as well.
If people loudly complain about it, and have good reasons to do so, we might reintroduce it

riccardoferretti:
Hi @memeplex! With Xmas and the holidays I disconnected for a few days, and am now catching back up.
What's the status on this PR? My only remaining point was around the navigation provider

memeplex:
Hi! I'm in the same boat :) I still need to check the navigation provider issue and perhaps add support for tags. I guess I'll be doing it soon.

riccardoferretti:
Cool, thanks :) I hope you had a good break!

riccardoferretti (commented):
Hi @memeplex, let me know if you will have time to wrap up this PR, thanks!

riccardoferretti:
(also, not sure whether we wanted #892 to be included here)

packages/foam-vscode/src/features/navigation-provider.ts
memeplex: Sorry for the delay, I've been very busy moving house. I think you're right, I should check for the wikilink type, but I would like to see how the direct link case looks like, perhaps some visual enhancement is in order too.

memeplex:
> not sure whether we wanted https://github.com/foambubble/foam/issues/892 to be included here

I've been thinking about adding tag highlighting. On the one hand, I like to think about tags as a kind of link that opens a search or even support peeking references, in this regard the syntax would be the same as the one for links. On the other hand, that syntax only makes sense if that link functionality is actually in place, otherwise it would be misleading, but I don't want to implement that as part of this PR.

So if you have a more neutral syntax to suggest, I can add it as  part of the injection.

memeplex:
>  it overrides our link decorations, done via the grammar, but not with placeholder decorations, done via the document decorator.

Probably some precedence issue, I guess. I don't know of any way to set decorator precedence, not to say of any way of configuring that in all conflicting extensions. I suppose vscode calls decorators in *some* order and that's it.

I found a bug in the decoration of placeholders: for direct links it's doing the +2 -2 thing. One fix is to colorize the entire direct link:

image

Another fix is to leave direct links alone. Short of some smarter only-title or only-url decoration, what do you prefer?

memeplex:
> it overrides our link decorations, done via the grammar, but not with placeholder decorations, done via the document decorator.

See https://github.com/microsoft/vscode/issues/33852#issuecomment-345755847

riccardoferretti:
If it's a placeholder it's fine by me to highlight the whole link, the same color we would highlight a `[[placeholder]]`

packages/foam-vscode/src/features/navigation-provider.ts
memeplex: This doesn't seem to be a problem because AFAICS resource.links only include wikilinks.

packages/foam-vscode/package.json
memeplex: Done.

packages/foam-vscode/src/features/navigation-provider.ts
riccardoferretti: > AFAICS resource.links only include wikilinks

`resource.links` also contains direct links, am I missing something?

(btw, if we create a new `vscode.Range` we don't need the `toVsCodeRange` function)

memeplex: I didn't dig into this, but in a page with a wikilink and a direct link only the first one was included in resource.links. Nevertheless it wouldn't hurt to defensively check for wikilinks.

memeplex: I've added a check and removed the unnecessary conversion.

memeplex:
I've added wikilink syntax support in markdown headers too (besides paragraphs).

riccardoferretti (approved):
This looks good to me, is there anything you wanted to tweak before merging?

memeplex:
> is there anything you wanted to tweak before merging?

No, I'm good.

riccardoferretti:
Thanks @memeplex !

PR #897, opened 2022-01-02
Fix open-daily-note to only accept argument if it is a Date

MCluck90:
I was trying to make VS Code automatically open my daily note with a task like this:

```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Open daily note",
      "command": "${command:foam-vscode.open-daily-note}",
      "runOptions": { "runOn": "folderOpen" }
    }
  ]
}
```

However, running that task would produce an `Invalid date` error. It turns out that calling a command through a task passes in an array of arguments. This array was being treated as a `Date` and causing issues. This PR just makes sure that the argument coming through is an actual `Date`.


riccardoferretti (approved):
I like this, thanks for pushing a PR to add the ability to run the command via a task

PR #902, opened 2022-01-05
Fix commented test case that might randomly fail

bentongxyz:
The test `should get the input from the active editor selection` was commented for `copyWithoutBrackets` (See #468 ).

Uncommented the test and fix the error that would cause it to randomly fail.

### DIFF
--- a/packages/foam-vscode/src/features/copy-without-brackets.spec.ts
+++ b/packages/foam-vscode/src/features/copy-without-brackets.spec.ts
@@ -7,8 +7,7 @@ import {
 
 describe('copyWithoutBrackets', () => {
   it('should get the input from the active editor selection', async () => {
-    await createFile('This is my [[test-content]].', ['hello.md']);
-    const uri = getUriInWorkspace('hello.md');
+    const { uri } = await createFile('This is my [[test-content]].');
     const { editor } = await showInEditor(uri);
     editor.selection = new Selection(new Position(0, 0), new Position(1, 0));
     await commands.executeCommand('foam-vscode.copy-without-brackets');
--- a/packages/foam-vscode/src/features/copy-without-brackets.spec.ts
+++ b/packages/foam-vscode/src/features/copy-without-brackets.spec.ts
@@ -1,21 +1,23 @@
-// import { env, window, Uri, Position, Selection, commands } from 'vscode';
-// import * as vscode from 'vscode';
+import { env, Position, Selection, commands } from 'vscode';
+import {
+  createFile,
+  getUriInWorkspace,
+  showInEditor,
+} from '../test/test-utils-vscode';
 
 describe('copyWithoutBrackets', () => {
   it('should pass CI', () => {
     expect(true).toBe(true);
   });
-  // it('should get the input from the active editor selection', async () => {
-  //   const doc = await vscode.workspace.openTextDocument(
-  //     Uri.parse('untitled:/hello.md')
-  //   );
-  //   const editor = await window.showTextDocument(doc);
-  //   editor.edit(builder => {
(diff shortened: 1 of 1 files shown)


packages/foam-vscode/src/features/copy-without-brackets.spec.ts
```diff
@@ -1,21 +1,23 @@
...
-  //   expect(value).toEqual('This is my Test Content.');
-  // });
+
+  it('should get the input from the active editor selection', async () => {
+    await createFile('This is my [[test-content]].', ['hello.md']);
+    const uri = getUriInWorkspace('hello.md');
+    const { editor } = await showInEditor(uri);
+    editor.selection = new Selection(new Position(0, 0), new Position(1, 0));
+    await commands.executeCommand('foam-vscode.copy-without-brackets');
+    console.log(env.clipboard);
```
riccardoferretti, line 19: I think we can remove this

packages/foam-vscode/src/features/copy-without-brackets.spec.ts
```diff
@@ -1,21 +1,23 @@
...
-// import * as vscode from 'vscode';
+import { env, Position, Selection, commands } from 'vscode';
+import {
+  createFile,
+  getUriInWorkspace,
+  showInEditor,
+} from '../test/test-utils-vscode';
 
 describe('copyWithoutBrackets', () => {
   it('should pass CI', () => {
```
riccardoferretti, line 9: We can remove this test now as it was only there because the `describe` requires a test case inside it

riccardoferretti (commented):
Thanks @bentongxyz! I have left a couple of minor comments

bentongxyz (commented):
Got it, much thanks @riccardoferretti !

packages/foam-vscode/src/features/copy-without-brackets.spec.ts
```diff
@@ -1,21 +1,18 @@
...
+  getUriInWorkspace,
+  showInEditor,
+} from '../test/test-utils-vscode';
 
 describe('copyWithoutBrackets', () => {
-  it('should pass CI', () => {
-    expect(true).toBe(true);
+  it('should get the input from the active editor selection', async () => {
+    await createFile('This is my [[test-content]].', ['hello.md']);
+    const uri = getUriInWorkspace('hello.md');
```
riccardoferretti, line 11: One thing I missed before, you should be able to do this:

```suggestion
    const { uri } = await createFile('This is my [[test-content]].');
```

Also, what was causing the random fail in the test and in what way was the issue fixed?

bentongxyz: Thanks!

As regarding the cause, when i was running tests with previous code, I notice that sometimes other test will intercept this `should get the input from the active editor selection` test, for example:

```
 FAIL  src/dated-notes.spec.ts
  ● Daily note template › Uses the daily note variables in the template

    expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 1

      hello September 12 hello
    + This is my [[test-content]].

      86 |     const doc = await showInEditor(uri);
      87 |     const content = doc.editor.document.getText();
    > 88 |     expect(content).toEqual('hello September 12 hello');
         |                     ^
      89 |   });
      90 |
      91 |   afterAll(async () => {

      at Object.<anonymous> (src/dated-notes.spec.ts:88:21)
```

this occurs in one of my runs, where this unrelated test received `This is my [[test-content]].`.

I rewrite the likely offending lines and the failures no longer occur.

riccardoferretti (approved):
Thanks!

riccardoferretti:
@allcontributors add @bentongxyz for code

PR #935, opened 2022-02-19
Added recommended tool

radoslawkierznowski:
I added a recommended plugin that saves the page directly to the repository.


riccardoferretti (approved):
Good find!

PR #937, opened 2022-02-20
Add Config: "foam.tags.poundSignTags"

metaworm:
There are some `#xxx` in my markdown notes, which are not truly tags, and they take up most of the space in the tags explorer.

I need a approach to disable these 'fake tags', i only need the tags in markdown's metadata

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -362,6 +362,11 @@
           "type": "object",
           "description": "Custom graph styling settings. An example is present in the documentation.",
           "default": {}
+        },
+        "foam.tags.poundSignTags": {
+          "type": "boolean",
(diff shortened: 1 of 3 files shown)


riccardoferretti:
Hi @metaworm, can you please add a description to PR, with the problem you are trying to solve and the approach you are taking?

riccardoferretti:
Thanks for contributing this @metaworm.
Although I fully understand your use case, as a design principle I want to keep Foam as light as possible, including configuration options. If there is enough interest for your use case for it to be a first class citizen, I would consider merging it in.
Otherwise I would suggest to wait for the API to be available, which might be able to satisfy your requirements by other means.
Happy to hear your thoughts

Issue #941, opened 2022-02-24
Unable to load notes when opened in volume root directory


develmusa:
- Foam version: [v0.17.5](https://github.com/foambubble/foam/releases/tag/v0.17.5)
- Platform: Windows
- Issue occur on the [foam template](https://github.com/foambubble/foam-template) repo: Yes 

**Summary**
If foam is opened in the root directory of a volume, no notes are loaded.

**Steps to reproduce**
1. Clone [foam template](https://github.com/foambubble/foam-template)
2. Move [foam template](https://github.com/foambubble/foam-template) files to volume root directory (eg: C:\)
3. Open volume root directory in VSCode

**Additional information**

- VSCode Foam Output:
  ```
  Foam Logging: debug
  [info - 16:35:59] Starting Foam
  [info - 16:35:59] Workspace folders: 
  [
    "F:/"
  ]
  [info - 16:35:59] Glob patterns
  {
    "includeGlobs": [
      "F:/**/*"
    ],
    "ignoreGlobs": [
      "F:/**/.vscode/**/*",
      "F:/**/_layouts/**/*",
      "F:/**/_site/**/*",
      "F:/**/node_modules/**/*",
      "F:/**/.git",
      "F:/**/.svn",
      "F:/**/.hg",
      "F:/**/CVS",
      "F:/**/.DS_Store",
      "F:/**/Thumbs.db",
      "F:/**/$RECYCLE.BIN",
      "F:/**/System Volume Information",
      "F:/**/node_modules"
    ]
  }
  [info - 16:35:59] Loaded 0 notes
  ```
- File structure on volume

  ```ps
  # PS F:\> tree /F
  Folder PATH listing for volume New Volume
  Volume serial number is 148A-B8C0
  F:.
  │   getting-started.md
  │   inbox.md
  │   readme.md
  │   test.txt
  │   todo.md
  │   
  ├───.foam
  │   └───templates
  │           your-first-template.md
  │           
  ├───.vscode
  │       extensions.json
  │       foam.json
  │       settings.json
  │       spellright.dict
  │       
  ├───assets
  │   └───css
  │           style.scss
  │           
  ├───attachments
  │       foam-icon.png
  │       
  ├───docs
  │   ├───features
  │   │       daily-notes.md
  │   │       graph-visualization.md
  │   │       link-reference-definitions.md
  │   │       note-properties.md
  │   │       note-templates.md
  │   │       spell-checking.md
  │   │       tags-and-tag-explorer.md
[...]

PR #943, opened 2022-02-24
fix: temporary solution for root level file discovery in windows

develmusa:
According to #941

Set ```strict: false``` option for  [node-glob](https://github.com/isaacs/node-glob) to allow scanning root directories in windows. Without ```strict: false``` the scanning will be canceled with this error:
```
An error occurred while bootstrapping Foam
"Error: EPERM: operation not permitted, scandir 'C:\\StoragePool\\167GB-SSD\\System Volume Information'"
```

> ```strict``` When an unusual error is encountered when attempting to read a directory, the process will just continue on in search of other matches. Set the strict option to raise an error in these cases.
> --- [node-glob/README.md](https://github.com/isaacs/node-glob/blob/3bfec21dd180ddf4672880176ad33af6296a167e/README.md?plain=1#L232)

### DIFF
--- a/packages/foam-vscode/src/core/services/datastore.ts
+++ b/packages/foam-vscode/src/core/services/datastore.ts
@@ -118,6 +118,7 @@ export class FileDataStore implements IDataStore {
   async list(glob: string, ignoreGlob?: string | string[]): Promise<URI[]> {
     const res = await findAllFiles(glob, {
       ignore: ignoreGlob,
+      strict: false,
     });
     return res.map(URI.file);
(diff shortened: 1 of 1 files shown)


riccardoferretti (approved):
Thanks for submitting the PR @develmusa - looks to me like a good solution to the problem.

I see that unfortunately there is no way to see which errors are being skipped, but at the same time it makes sense to not be strict and have a single file block a whole scan (especially because it could be, like in this case, a file we are even interested in).

Issue #933, opened 2022-02-18
Daily Note for tomorrow?


alper:
Is there a convenient way to create the Daily Note for tomorrow?

I work a lot in the today, but usually I create the note for tomorrow and do my day prep there.

I could make the change but not sure what the design should be? Maybe a setting for the daily note that is default for "today" but that you can also set for "tomorrow"?

chrisUsick:
One design could be a “select daily note” command that would show a
calendar. You could pick a past or future date, and if the not doesn’t
exist it gets created, but if it exists it simply gets opened.

alper:
You could of course pre-generate a `2022-calendar.md` note with all the days already in it as uncreated note links…

But even then another part of the workflow that's the most annoying is having to manually set the title.

riccardoferretti:
I believe an approach could be similar to what we are doing for date snippets.
Have a `Go to day note` command, which shows a quick pick list with available dates, plus the items for the next and previous timeframes (as defined in `open-dated-notes.ts`) - similar to how `/` will trigger a list of possible completion based on the same logic.

leira:
@alper Thanks for opening this issue. I was going to request the exact same thing. It is a good practice to plan for the next day at the end of the day. A `tomorrow` command will be really useful.

alper:
I've got a proof of concept working where you can pick a note for today or tomorrow.

But I don't see where the functionality in `open-dated-note.ts` is actually used. This seems mostly unrelated?
I can't figure out where the `foam-vscode.open-dated-note` command is run.

alper:
I'm thinking of having a Quick Pick with "today", "tomorrow" and otherwise just take an isodate. Would that work?

alper:
That's cool. I was still struggling with tests.

riccardoferretti:
Yup I only added the tests for the base case, I will at least add the placeholders for a few more, but as I am planning a couple of changes in the way the command work (I actually already played with the idea locally and it's really interesting what it opens) I decided to wait for that.
What was in your tests list?
Thanks for spearheading the issue, I find it super useful and it will be released soon.

alper:
Oh cool, I'll take a look at it and look forward to see which direction you want to take it.

It would be nice if it was documented how to run these specific test suites. I've been struggling a bunch with it and running it nuked a bunch of files in one of my open VSCode instances.

riccardoferretti:
We have this document to help with the setup:
[...]

PR #957, opened 2022-03-22
Make it possible to create a daily note for other days

alper:
As discussed in #933

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -171,6 +171,10 @@
         "title": "Foam: Open Daily Note"
       },
       {
+        "command": "foam-vscode.open-daily-note-for-date",
+        "title": "Foam: Pick Date for Daily Note"
+      },
+      {
         "command": "foam-vscode.open-random-note",
         "title": "Foam: Open Random Note"
       },
@@ -377,6 +381,10 @@
       {
         "command": "foam-vscode.open-daily-note",
         "key": "alt+d"
+      },
+      {
+        "command": "foam-vscode.open-daily-note-for-date",
+        "key": "alt+h"
       }
     ]
   },
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -51,7 +51,7 @@ function generateDateItems(): DateItem[] {
   items.push(new DateItem('tomorrow', tomorrow));
 
   // Future
-  for (let offset = 2; offset < 6; offset++) {
+  for (let offset = 2; offset < 8; offset++) {
     const d = new Date();
     d.setDate(d.getDate() + offset);
     items.push(new DateItem(d.toDateString(), d));
--- a/packages/foam-vscode/src/dated-notes.ts
+++ b/packages/foam-vscode/src/dated-notes.ts
@@ -1,9 +1,32 @@
-import { workspace, WorkspaceConfiguration } from 'vscode';
+import { window, workspace, WorkspaceConfiguration } from 'vscode';
 import dateFormat from 'dateformat';
 import { focusNote } from './utils';
 import { URI } from './core/model/uri';
 import { fromVsCodeUri, toVsCodeUri } from './utils/vsc-utils';
 import { NoteFactory } from './services/templates';
+
+export async function openDailyNoteForToday() {
+  console.log('Open daily note for today');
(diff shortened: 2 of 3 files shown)


riccardoferretti:
Hi @alper thanks for taking this on. 
The approach looks good to me, the only thing that is missing is which days to include.

Do you have any intuition around that?

I was thinking:
- today
- next week days
- last week days
- +1 -> 30 days
- +1 -> 12 months
- +1 year

The default value in the quick pick would be today

alper:
I don't understand what you mean with +1.

> +1 -> 30 days
> +1 -> 12 months
> +1 year

The first two would mean a list with all the days for the last week and for the current week? That's already 14 items?

riccardoferretti:
Yup, there would be various items in this list:
- today
- from tomorrow, every day for the next 31 days
- from yesterday, every day back to 31 days ago
- 1 month from today, 2 months from today, ... for the next 12 months

Again, this is just the list I was thinking of, happy to hear your thoughts on what you believe would be useful there

alper:
I think that would be way too many. I'm no working with this and I'd offer freeform input as well so that people could type in a date like 2022-06-26 or something and still get the benefits of the daily note and template.

CleanShot 2022-03-29 at 16 24 18@2x

alper:
OK. So arbitrary text input is not possible. Neither is pre-selecting the entry for today.

So maybe I'll have to add a bunch more days to expand the options a bunch, but I feel it'll not be very clear and you'll always be missing the day somebody may want to have.

packages/foam-vscode/package.json
```diff
@@ -166,6 +166,10 @@
         "command": "foam-vscode.open-daily-note",
         "title": "Foam: Open Daily Note"
       },
+      {
+        "command": "foam-vscode.open-daily-note-for-date",
```
riccardoferretti, line 170: instead of adding a new command, could we retrofit/expand the previous `foam-vscode.open-daily-note` command to default to today (as it currently is), but to possibly also take a parameter with the date to open?

packages/foam-vscode/src/dated-notes.ts
```diff
@@ -1,10 +1,83 @@
...
+    items.push(new DateItem(d.toDateString(), d));
+  }
+
+  return items;
+}
+
+export async function openDailyNoteForPickedDate() {
+  console.log('Open daily note for date');
+
+  const result = await window.showQuickPick<DateItem>(generateDateItems(), {
```
riccardoferretti, line 67: I believe you should be able to have a default by looking at the API, am I missing something? https://code.visualstudio.com/api/references/vscode-api#QuickPickItem

packages/foam-vscode/package.json
alper: How would we pass a parameter to a command? Do you have an example for me?

alper:
I think freeform input would be possible if I used this pattern:

https://github.com/microsoft/vscode-extension-samples/blob/63d86e3a0e459b2adf26487204891f7af62898ce/quickinput-sample/src/quickOpen.ts#L15

packages/foam-vscode/src/dated-notes.ts
alper: That's what I thought but the `picked` property is about items that are selected if you allow `canPickMany`.

packages/foam-vscode/package.json
riccardoferretti: You might be able to get an idea from `utility-commands.ts` and https://stackoverflow.com/questions/34809864/arguments-for-commands-in-vs-code

does that help?

packages/foam-vscode/src/dated-notes.ts
riccardoferretti: yup, you are right, I got confused :)

riccardoferretti:
hi @alper let me know how things are going on this issue and if you need anything else at this time.
You will need to rebase because there is a conflict on the branch, I am happy to support if you want.
Thanks!

riccardoferretti:
(ping me when the PR is ready for review)

riccardoferretti:
@alper I take you have been busy but I still wanted to wrap up the good work you have done and merge it, so I have created another PR based on this one: https://github.com/foambubble/foam/pull/1026

Will close this, feel free to leave any comment on the other

Issue #903, opened 2022-01-06
Allow using emoji in tags


ricekot:
Examples:
- Purely Emoji:
  - `#📥/🟩`, `#📥/🟧`, `#📥/🟥`
  - `#⭐`, `#⭐⭐`, `#⭐⭐⭐`
  - `#🐞`
- Text and Emoji:
  - `#example1️⃣`
  - `#example/1️⃣`

riccardoferretti:
Happy to receive a contribution to support this

badsketch:
The above PR admitted that it didn't cover all the cases for this issue, but I'm curious why it ended up being closed? I'd like to take a stab at this 🙂

PR #968, opened 2022-04-03
Support for Emoji-Only Tags

njnygaard:
This does not fully address #903 as the requirements there include "text[emoji]", but I wanted to take a stab at it.
These extra captures will identify emoji-only tags. I'm thinking this might be useful for 🐛  or pithy tags like that.
It looks nice in the tag explorer:
image
I tested this with: https://github.com/foambubble/foam-template
Any feedback is appreciated.


riccardoferretti:
Thanks for taking a stab at this, can we add some tests?
Is it complex to merge the two regexes?

riccardoferretti:
Hi @njnygaard - just checking in on this PR

Issue #962, opened 2022-03-28
Jump after the closing brackets (`]]`) on autocomplete


luisherranz:
### Is your feature request related to a problem? Please describe.

I'd love Foam to jump ahead the closing brackets `]]` when I use autocomplete.

### Describe the solution you'd like

When using autocomplete (by either clicking or using `Tab/Enter` on a result), Foam moves the cursor to the last word, like this:

- `I am linking to [[some page|]]`

You then have to type `]]` to continue writing.

It'd be great if Foam could jump ahead after the `]]` instead, like this:

- `I am linking to [[some page]]|`

Then you could open double brackets `[[`, partially complete the name of the page (`some p`), use `Tab` to finish the autocomplete, and keep writing the next words.

### Describe alternatives you've considered

Write the `]]` yourself, although I can't imagine a case where you wouldn't want to jump ahead. Similar software, like LogSeq, already works this way.

riccardoferretti:
That's an interesting optimization, I can see it speeding things up when writing notes.
I don't know whether it might interfere with some use cases (e.g. using sections), so this requires a bit of deep dive.
Happy to keep it in the backlog, and if someone wants to work on it I am happy to support, at least until v1 is out, at which point I could take a look myself.

Lauviah0622:
I'm trying to implement this feature but found problem. Like @riccardoferretti  said, it will interfere the section use case. 

Here is the demo, cursor jump after the closing bracket that cannot trigger the section auto completion. 
![movecursor](https://user-images.githubusercontent.com/43416399/169541125-b06630f0-5fa6-484f-a1a3-8499035ed417.gif)

but the cursor move in section auto completion is work fine. 

![movecursor-sec](https://user-images.githubusercontent.com/43416399/169541535-d1622b79-6c8d-4cc9-be5d-c4870ac107f6.gif)

luisherranz:
Maybe leave it optional for people not using sections? (I'm not)

riccardoferretti:
@Lauviah0622 Not sure which char you were using to confirm `test` in your first gif? what happens if you confirm the wikilink with the `#`? that must trigger the section autocomplete.
If instead you press enter and you get out of the wikilink, well, I think that could be fine.

Lauviah0622:
I'm sorry forgot to record my keystroke in demo! Here is the new version. 
[...]

PR #998, opened 2022-05-23
Jump after the closing brackets (]]) on autocomplete 

Lauviah0622:
Reference to #962. 

- Add commit character `|` on link completion and section completion 
- Jump after the closing bracket on autocomplete except commit with `#` and `|`


riccardoferretti (commented):
Interesting approach!
Only comment, I would put the command code together in `link-completion.ts`: given that it's only meant to serve that functionality, I think it's a bit clearer that way.

Also, given that you had to work around the API, which means the purpose of the feature is not immediately clear by looking at the code, I would add a reference to what you are accomplishing (maybe even to the issue).

Thanks for figuring this out!

Lauviah0622:
> Only comment, I would put the command code together in link-completion.ts: given that it's only meant to serve that functionality, I think it's a bit clearer that way.

Not a problem, I split out the file cause I'm not sure are there any other feature to use it in future. But it seems too single purposely so far, I agree with move in the `link-completion.ts`

> Also, given that you had to work around the API, which means the purpose of the feature is not immediately clear by looking at the code, I would add a reference to what you are accomplishing (maybe even to the issue).

Thats a practical suggestion. I'll add some addition comment to referring to issue #962.

riccardoferretti (approved):
Thanks!

riccardoferretti:
@allcontributors add @Lauviah0622 for code

PR #999, opened 2022-05-23
fix #990: lost escape character (\) in table after rename a file

### LINKED ISSUE #990: Lost escape character (\) in table after rename a file
### Describe the bug

- Give an simple table and wikilink with an alias like this, I use `\` to escape `|` because this wikilink is in a table.
```
|                |
| -------------- |
| [[test\|Test]] |
```

- After I rename the `test.md` file to `test2.md`, the table will be updated like this (missing the `\`)
```
|                |
| -------------- |
| [[test2|Test]] |
```

_No response_

1. Create a new note with this content:
```
|                |
| -------------- |
| [[test\|Test]] |
```

2. Create a `test.md` file.
3. Rename the `test.md` to `test2.md`.
4. Verify the table in the first note.

### Expected behavior

As a user, I expect that after I rename a file, the wikilink in a table does not remove escape character.

### Screenshots or Videos

_No response_

### Operating System Version

Linux

1.66.1

### Additional context

Foam version v0.18.3

Lauviah0622:
ref to #990.

### DIFF
--- a/packages/foam-vscode/src/core/services/markdown-link.ts
+++ b/packages/foam-vscode/src/core/services/markdown-link.ts
@@ -2,8 +2,9 @@ import { ResourceLink } from '../model/note';
 
 export abstract class MarkdownLink {
   private static wikilinkRegex = new RegExp(
-    /\[\[([^#|]+)?#?([^|]+)?\|?(.*)?\]\]/
+    /\[\[([^#|]+[^\\|#])?#?([^|]+[^\\|])?(\\\||\|)?(.*)?\]\]/
   );
+
   private static directLinkRegex = new RegExp(
     /\[(.*)\]\(([^#]*)?#?([^\]]+)?\)/
   );
@@ -11,13 +12,19 @@ export abstract class MarkdownLink {
   public static analyzeLink(link: ResourceLink) {
     try {
       if (link.type === 'wikilink') {
-        const [, target, section, alias] = this.wikilinkRegex.exec(
-          link.rawText
-        );
+        const [
+          ,
+          target,
+          section,
+          aliasDivider,
+          alias,
+        ] = this.wikilinkRegex.exec(link.rawText);
+
         return {
           rawTarget: target,
           target: target?.replace(/\\/g, '') ?? '',
           section: section ?? '',
+          aliasDivider: aliasDivider ?? '',
           alias: alias ?? '',
         };
       }
@@ -41,23 +48,19 @@ export abstract class MarkdownLink {
     link: ResourceLink,
     delta: { target?: string; section?: string; alias?: string }
   ) {
-    const { target, section, alias, rawTarget } = MarkdownLink.analyzeLink(
+    const { target, section, alias, aliasDivider } = MarkdownLink.analyzeLink(
       link
     );
 
-    let replacedTarget = null;
-    if (rawTarget && delta.target) {
-      replacedTarget = rawTarget.replace(target, delta.target);
-    }
-
-    const newTarget = replacedTarget ?? delta.target ?? target;
+    const newTarget = delta.target ?? target;
     const newSection = delta.section ?? section ?? '';
     const newAlias = delta.alias ?? alias ?? '';
     const sectionDivider = newSection ? '#' : '';
-    const aliasDivider = newAlias ? '|' : '';
+    const newAliasDivider = aliasDivider || (newAlias ? '|' : '');
+
     if (link.type === 'wikilink') {
       return {
-        newText: `[[${newTarget}${sectionDivider}${newSection}${aliasDivider}${newAlias}]]`,
+        newText: `[[${newTarget}${sectionDivider}${newSection}${newAliasDivider}${newAlias}]]`,
         selection: link.range,
       };
     }
--- a/packages/foam-vscode/src/core/services/markdown-link.ts
+++ b/packages/foam-vscode/src/core/services/markdown-link.ts
@@ -1,9 +1,40 @@
 import { ResourceLink } from '../model/note';
 
 export abstract class MarkdownLink {
-  private static wikilinkRegex = new RegExp(
-    /\[\[([^#|]+[^\\|#])?#?([^|]+[^\\|])?(\\\||\|)?(.*)?\]\]/
-  );
+  private static parseWikilink = (text: string) => {
+    const textWithoutBracket = text.replace(/^\[\[|\]\]$/g, '');
+    const splitGroup = textWithoutBracket.split(/(#|\\\||\|)/);
+
+    const hasAliasDivider = splitGroup.some(group => group.match(/[\\|||]/));
+    const hasSectionDivider = splitGroup.some(group => group.match(/#/));
+    if (hasAliasDivider && hasSectionDivider) {
+      const [target, sectionDivider, section, aliasDivider, alias] = splitGroup;
+      return {
+        target,
+        sectionDivider,
+        section,
+        aliasDivider,
+        alias,
+      };
+    } else if (hasAliasDivider) {
+      const [target, aliasDivider, alias] = splitGroup;
+      return {
+        target,
+        aliasDivider,
+        alias,
+      };
+    } else if (hasSectionDivider) {
+      const [target, sectionDivider, section] = splitGroup;
+      return {
+        target,
+        sectionDivider,
+        section,
+      };
+    } else {
+      const [target] = splitGroup;
+      return { target };
+    }
+  };
 
   private static directLinkRegex = new RegExp(
     /\[(.*)\]\(([^#]*)?#?([^\]]+)?\)/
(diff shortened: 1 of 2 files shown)


packages/foam-vscode/src/core/services/markdown-link.ts
```diff
@@ -15,6 +15,7 @@ export abstract class MarkdownLink {
           link.rawText
         );
         return {
+          rawTarget: target,
```
riccardoferretti, line 18: have you considered changing the regex to capture the alias separator used and returning that?
I would say that semantically might be a slightly "cleaner" solution, as the `\` is not really part of the target.

What do you think?

Lauviah0622: I had tried that, but I found there has a so uncommon use case is `\` in [valid  character in file name on linux / Unix(MacOS) but not in windows.](https://en.wikipedia.org/wiki/Filename) 

But currently, the oddly character link `|` `\` will break down the of Foam filename analyze system a lot. If there are not going to dealing with these cases. I prefer the cleaner solution you said that use regex to split the alias separator and combined it.

riccardoferretti: Not sure if I fully understand, let me give you some context.
In Foam identifiers only use the `/` divider to describe a path (maybe `\` currently works but that's a coincidence). e.g. when we create a unique identifier, which needs to include part of the path, we use `/`.
Also, the only thing we care about is if the `\` char is used right before the `|`. It's a bit of an odd case (because in theory it only applies inside tables), but I think it's ok to generalize to the fact that the allowed separators for the alias are `|` and `\|`. 
Does the above make sense? Is that answering your questions?

Lauviah0622: I got the point. but there is another problem.

I found the case: If use `\` in filename (this is so weird, and is invalid in windows). In this case, we cannot realize the `\` char is the part of escaped in divider or filename, only if extract the special case regex of escaped divider inside table, but it will be more complicated (need another special regex rule I think)

But if does not care this case(its uncommon and only occur in Linux and Unix).  we can just improve the current regex like what you said.

riccardoferretti: If I understand, your special case only applies when `\` is the last char of the filename.
In that case, but really even in the case of `\` anywhere in the file, I feel this is an uncommon enough exception that I am ok with not supporting it.

I was thinking, there might be one more issue with having `\` in the target: will these two links point to the same file?

---

| col1 | col2 |
|-|-|
| [[wikilink\|alias]] | [[wikilink]] |

[[wikilink]]

---

I have a feeling the link with the escape would resolve to another target?
It would actually be good to have a test case for this just to be sure.
But yeah, that feels like the risk of "overloading" the `target` and the why it feels "safer" to simply expose the `aliasSeparator`.

Thoughts?

Lauviah0622: > I was thinking, there might be one more issue with having \ in the target: will these two links point to the same file?

Do you mean in the source text is

```
| col1            | col2     |
| --------------- | -------- |
| [[wikilink\|alias]] | [[wikilink]] |

[[wikilink]]
```

In the cases above,  it will link to the same file `wikilink.md`. But consider the cases below. 

```
file name with '|'
[[test|]]                                  // link to test.md

file name with '\'
[[test\]]                                  // link to test.md

alias link file name with '\'
[[test\|test]]                             // link to test.md

section link file name with '\'
[[test\#Graph Visualization]]              // link to test.md

section alias link file name with '\'
[[test\#Graph Visualization|test]]         // link to test.md

|                                      | 
| -----------------------------------  | 
| [[test\]]                            |   // link to test.md
| [[test\\|alias]]                     |   // cannot be parse as valid wikilink
| [[test\#Graph Visualization\|test]]  |   // link to test.md
| [test](test\.md)                     |   // link to test\.md
| [test](test\.md#custom-graph-styles) |   // link to test\.md
| [test](test|.md)                     |   // link to test|.md
| [test](test|.md#custom-graph-styles) |   // link to test|.md
```

These case is not complete. but the uncommon end char will break the link parser. I'm agree with `expose the aliasSeparator` is  the better way. 

So I think we should have two kind of `aliasSeperator`? 

- `\|` only use in table for escape. 
- `|` use in any other context

And in regex maybe we should split out these case.

Lauviah0622:
I just change the new replacement rule, but we still can keep in discussion and improve it!

packages/foam-vscode/src/core/services/markdown-link.ts
riccardoferretti: Thanks for the recap, it helped me think about this holistically, here are my thoughts: the way I see it is just that we need to support `\|`, for a very specific case, which is an alias inside a table.
I don't feel we need to be strict about where the link is (that would get messy pretty quickly), so what I would say in general is that `\|` is equivalent to `|`.
To me that's all we need, which means that your table above to me should be: 

```
file name with '|'
[[test|]]                                  // link to test.md

file name with '\'
[[test\]]                                  // link to test\.md

alias link file name with '\'
[[test\|test]]                             // link to test.md

section link file name with '\'
[[test\#Graph Visualization]]              // link to test\.md

section alias link file name with '\'
[[test\#Graph Visualization|test]]         // link to test\.md

|                                      | 
| -----------------------------------  | 
| [[test\]]                            |   // link to test\.md
| [[test\\|alias]]                     |   // link to test\.md
| [[test\#Graph Visualization\|test]]  |   // link to test\.md
| [test](test\.md)                     |   // link to test\.md
| [test](test\.md#custom-graph-styles) |   // link to test\.md
| [test](test|.md)                     |   // link to test|.md
| [test](test|.md#custom-graph-styles) |   // link to test|.md
```

I know it's slightly different from the current implementation, unsure whether this simplify things or not, but I do believe it's the "most appropriate" interpretation of the issue at hand, which to rephrase it for me is about supporting wikilink alias inside a table, and not supporting escape chars for section/alias separators. What do you think?

Lauviah0622: >  I would say they are both valid and when refactoring we keep track of which one was used and keep using it. Some might say this is not 100% correct, and they might be right, but it's simple for the user to understand and I'd say it covers 99.998% of cases.

I'm agree with the simplicity, and its is enough that cover most of use cases

> supporting wikilink alias inside a table, and not supporting escape chars for section/alias separators. 

Is this mean we should deal with incorrect case in current implementation below: 

```
[[test\\|alias]]                       |   // cannot be parse as valid wikilink => link to test\.md
| [[test\#Graph Visualization\|test]]  |   // link to test.md => link to test\.md
```

In addition, I would like to ask where I can start to tracking the wikilnck is better than the beginning, It will make me resolve the problem more efficiently

riccardoferretti: > Is this mean we should deal with incorrect case in current implementation below:

What do you mean by "incorrect"?
In those cases I would say

`[[test\\|alias]]` maps to `test\` (given that here the separator is `\|`)
`[[test\#graph|test]]` maps to `test\` because we only support `\|` and no other "special use" of `\`

is that answering the question?

> In addition, I would like to ask where I can start to tracking the wikilnck is better than the beginning, It will make me resolve the problem more efficiently

Not sure I understand, what do you mean?

Thanks!

Lauviah0622: I'm sorry for unclear description. 

In this PR, I only fix the issue that update wikilink after rename file. 
But on the discussion above, we should also fix the problem that parser parse the wikilink in table with escape char in these case. 

```
| [[test\\|alias]]                     |   
| [[test\#Graph Visualization\|test]]  |
```

I'm willing to deal with it. But if any info like mentioned in [this comment ](https://github.com/foambubble/foam/issues/990#issuecomment-1112350100) will be helpful.

riccardoferretti: IIUC the approach of changing `MarkdownLink.analyzeLink` to support the two alias separators (`\|` and `|`) and report which one is used should fix both getting the right identifier within a table and the file rename. If I am missing something I would say let's focus this PR on the issue of updating wikilink on file rename (as you said), and if the approach feels incompatible with the second objective we can review it, I hope this makes sense.

Lauviah0622:
```
[[test\\|alias]]
```

The case cannot been parse above. Thats because the dependency `remark-wiki-link` cannot parse it as correct wikilink, but it fix at latest version 1.0.4. Here is [the wikilink parse result of foam current dependency version](https://github.com/Lauviah0622/foam-depencies-demo/blob/d054c44383c1b494454a7377ff16080fdf04a733/log.json#L75
) and [result of latest version](https://github.com/Lauviah0622/foam-depencies-demo/blob/7215f2b13d7cb4b11fa3dad5b91e18d5110b2668/log.json#L43).

The another incorrect case is fix in new commit, but there has need some improvement.
- Compare to simple regex, the complicated function may cause performance issue.
- If the wikilink is placeholder (no correspond file) foam will automatically create the file. But the auto created title of the file is strange. But I haven't found a problem yet.

Screen Shot 2022-06-10 at 12 11 41 AM

riccardoferretti:
> Compare to simple regex, the complicated function may cause performance issue.

Yes, I was wondering about that, do you have any idea of the difference in performance?

> If the wikilink is placeholder (no correspond file) foam will automatically create the file. But the auto created title of the file is strange. But I haven't found a problem yet.

This seems like a big issue to me.
Does it always happen or only when the placeholder has an alias?

Lauviah0622:
> This seems like a big issue to me.
Does it always happen or only when the placeholder has an alias?

This only occur on the file name end with `\` char. 
Current version [always replace the `\`](https://github.com/foambubble/foam/blob/dcd46f1378dfd35ad46c381e3c3846e939802f3f/packages/foam-vscode/src/core/services/markdown-link.ts#L18), so it will not happened.

Issue #749, opened 2021-08-22
Add support for VSCode Web


YashKumarVerma:
with the feature of loading a full fledged editor in the browser with just pressing the `.` on the keyboard, It would be great if we could bring the nodes to the browser.

![https://i.imgur.com/Gsoil6j.png](https://i.imgur.com/Gsoil6j.png)

digitarald:
👋 I'm Harald, a PM from the VS Code team, and we recently announced VS Code for the Web at https://vscode.dev/, which provides a free, zero-install Microsoft Visual Studio Code experience running entirely in your browser. You can learn more [here](https://code.visualstudio.com/docs/editor/vscode-web).

We'd like to ensure that Foam can run in the web, and we have a guide for enabling extensions for the web [here](https://code.visualstudio.com/api/extension-guides/web-extensions). As a couple of highlights:

* In VS Code for the Web, both the UI and extension host run inside the browser.
* A web extension is structured like a regular extension, but with a different main file: it's defined by the `browser` property
* Access to workspace files needs to go through the VS Code file system API accessible at `vscode.workspace.fs`
* There are currently [three ways](https://code.visualstudio.com/api/extension-guides/web-extensions#test-your-web-extension) to test a web extension

Feel free to reach out to me if you have questions or if I can help somehow. Thank you

riccardoferretti:
Hi @digitarald, it's definitely on my radar to have Foam run in the web, will definitely reach out if I have any question in that regard

AlexW00:
Any update on the issue?

riccardoferretti:
I did a spike that got the build to work but there are still some issues to iron through, because of the libraries that would not be available in the browser (e.g. glob), or that need to be bridged

joshdover:
I'm interested in picking this work up, here's my rough plan. Please let me know if you have any feedback:
- [ ] https://github.com/foambubble/foam/pull/1005
- [ ] Identify any other modules that have dependencies on Node.js that need to be refactored (`glob`?)
- [ ] Add a bundled version of the extension source, using rollup and @rollup/plugin-typescript
- [ ] Setup the e2e suite to run the same tests against VSCode using the `--extensionDevelopmentKind=web` option

riccardoferretti:
I removed the dependency on `glob`, which was the main blocker here.
There are a few more modules, `os` `path` and `crypto`, which can be browserified.
There is a pending issue with `fs` in the preview, due to the async limitation of the plugins, to be properly thought about.

[...]

PR #1005, opened 2022-06-06
Replace usages of fs with vscode.workspace.fs

joshdover:
Related to https://github.com/foambubble/foam/issues/749

This PR removes the hard dependencies on the `fs` module which is not available in the browser environment. `FileDataStore` now requires a `readFile` function which in tests uses `fs` but in extension mode uses `vscode.workspace.fs.readFile`. This is my first contribution to this project, so any feedback on how this is done is welcome.

### DIFF
--- a/packages/foam-vscode/src/core/services/markdown-provider.ts
+++ b/packages/foam-vscode/src/core/services/markdown-provider.ts
@@ -8,7 +8,7 @@ import { isNone, isSome } from '../utils';
 import { Logger } from '../utils/log';
 import { URI } from '../model/uri';
 import { FoamWorkspace } from '../model/workspace';
-import { IDataStore, FileDataStore, IMatcher } from '../services/datastore';
+import { IDataStore, IMatcher } from '../services/datastore';
 import { IDisposable } from '../common/lifecycle';
 import { ResourceProvider } from '../model/provider';
 import { createMarkdownParser } from './markdown-parser';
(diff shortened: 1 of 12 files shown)


packages/foam-vscode/src/core/services/markdown-provider.ts
```diff
@@ -19,13 +19,13 @@ export class MarkdownResourceProvider implements ResourceProvider {
 
   constructor(
     private readonly matcher: IMatcher,
+    private readonly dataStore: IDataStore,
```
joshdover, line 22: I made this required so that an explicit choice needs to be made on how files should be read

packages/foam-vscode/src/core/utils/path.ts
```diff
@@ -147,21 +146,6 @@ export function relativeTo(path: string, basePath: string): string {
   return posix.relative(basePath, path);
 }
 
-/**
- * Asynchronously checks if there is an accessible file for a path.
- *
- * @param fsPath A filesystem-specific path.
- * @returns true if an accesible file exists, false otherwise.
- */
-export async function existsInFs(fsPath: string) {
```
joshdover, line 156: This function wasn't used, so I removed it

joshdover:
> On your point re adding a rule to avoid future use of `fs`, I am happy to have it
> Will wait for that then to merge the PR

ESLint rule has been added in b477f55 👍

joshdover:
@riccardoferretti May I have CI run on my PR as well before we merge it?

PR #1014, opened 2022-06-18
Support alias yaml property

### LINKED ISSUE #461: Page aliases support.
Many times when I write note about particular topic, there are different ways it can be referred in a different note. It would be nice to support the alias in the frontmatter like:

    ---
    title: some title
    alias: synonym to title, different title
    ---
and [[some title]] or [[synonym to title]] or [[different title]] would all link to the same page.

lingyv-li:
Adds "alias" yaml property that can be either a string or a list of string.

`title` property is not supported as it can be confusing with the existing Resource.title (which may come from multiple places).

Closes #461

### DIFF
--- a/packages/foam-vscode/src/features/link-completion.spec.ts
+++ b/packages/foam-vscode/src/features/link-completion.spec.ts
@@ -157,4 +157,26 @@ Content of section 2
       new Set(['Section 1', 'Section 2'])
     );
   });
+
+  it('should return page alias', async () => {
+    const { uri, content } = await createFile(`
+---
+alias: alias-a
+---
+[[
+`);
+    ws.set(parser.parse(uri, content));
+
+    const { doc } = await showInEditor(uri);
+    const provider = new SectionCompletionProvider(ws);
+
+    const links = await provider.provideCompletionItems(
+      doc,
+      new vscode.Position(4, 2)
+    );
+
+    expect(new Set(links.items.map(i => i.label))).toEqual(
+      new Set(['alias-a'])
+    );
+  });
 });
--- a/packages/foam-vscode/src/features/link-completion.spec.ts
+++ b/packages/foam-vscode/src/features/link-completion.spec.ts
@@ -159,24 +159,28 @@ Content of section 2
   });
 
   it('should return page alias', async () => {
-    const { uri, content } = await createFile(`
+    const { uri, content } = await createFile(
+      `
 ---
 alias: alias-a
 ---
 [[
-`);
+`,
+      ['new-note-with-alias.md']
+    );
     ws.set(parser.parse(uri, content));
 
     const { doc } = await showInEditor(uri);
-    const provider = new SectionCompletionProvider(ws);
+    const provider = new CompletionProvider(ws, graph);
 
     const links = await provider.provideCompletionItems(
       doc,
       new vscode.Position(4, 2)
     );
 
-    expect(new Set(links.items.map(i => i.label))).toEqual(
-      new Set(['alias-a'])
-    );
+    const aliasCompletionItem = links.items.find(i => i.label === 'alias-a');
+    expect(aliasCompletionItem).not.toBeNull();
+    expect(aliasCompletionItem.label).toBe('alias-a');
+    expect(aliasCompletionItem.insertText).toBe('new-note-with-alias|alias-a');
   });
 });
--- a/packages/foam-vscode/src/features/link-completion.ts
+++ b/packages/foam-vscode/src/features/link-completion.ts
@@ -185,7 +185,7 @@ export class CompletionProvider
       position.line,
       position.character
     );
-    const resources = this.ws.list().map(resource => {
+    const resources = this.ws.list().flatMap(resource => {
       const label = vscode.workspace.asRelativePath(toVsCodeUri(resource.uri));
       const item = new ResourceCompletionItem(
         label,
@@ -197,7 +197,20 @@ export class CompletionProvider
       item.range = replacementRange;
       item.command = COMPLETION_CURSOR_MOVE;
       item.commitCharacters = linkCommitCharacters;
-      return item;
+      const aliasItems = resource.aliases.map(a => {
+        const item = new ResourceCompletionItem(
+          a.title,
+          vscode.CompletionItemKind.Reference,
+          resource.uri
+        );
+        item.insertText = this.ws.getIdentifier(resource.uri) + '|' + a.title;
+        item.detail = `Alias of ${resource.uri.getName()}`;
+        item.range = replacementRange;
+        item.command = COMPLETION_CURSOR_MOVE;
+        item.commitCharacters = linkCommitCharacters;
+        return item;
+      });
+      return [item, ...aliasItems];
     });
     const placeholders = Array.from(this.graph.placeholders.values()).map(
       uri => {
--- a/packages/foam-vscode/src/features/link-completion.ts
+++ b/packages/foam-vscode/src/features/link-completion.ts
@@ -7,6 +7,7 @@ import { FoamFeature } from '../types';
 import { getNoteTooltip, mdDocSelector } from '../utils';
 import { fromVsCodeUri, toVsCodeUri } from '../utils/vsc-utils';
 
+export const aliasCommitCharacters = ['#'];
 export const linkCommitCharacters = ['#', '|'];
 export const sectionCommitCharacters = ['|'];
 
@@ -185,7 +186,7 @@ export class CompletionProvider
       position.line,
       position.character
     );
-    const resources = this.ws.list().flatMap(resource => {
+    const resources = this.ws.list().map(resource => {
       const label = vscode.workspace.asRelativePath(toVsCodeUri(resource.uri));
       const item = new ResourceCompletionItem(
         label,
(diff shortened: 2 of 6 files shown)


packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -21,6 +21,11 @@ export class FoamWorkspace implements IDisposable {
    */
   private _resources: Map<string, Resource> = new Map();
 
+  /**
+   * Resources by alias
+   */
+  private _resources_alias: Map<string, Resource> = new Map();
```
riccardoferretti, line 27: camel case

riccardoferretti: what happens if more than one resource declares the same alias?

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -199,6 +199,21 @@ export class CompletionProvider
       item.commitCharacters = linkCommitCharacters;
       return item;
     });
+    resources.push(
```
riccardoferretti, line 202: have you considered creating a `const aliases = ...`? It's more in line with the style of the function, it doesn't modify the `resources` variable (which is technically defined as `const`, so it can also be misleading) and still makes it easy to place them where we want in the final list of suggestions.

Also, have you considered looping through resources only once and returning both `resources` and `aliases`?

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -199,6 +199,21 @@ export class CompletionProvider
...
+      ...this.ws.list().flatMap(resource =>
+        resource.aliases.map(a => {
+          const item = new ResourceCompletionItem(
```
riccardoferretti, line 205: I need to double check later by playing with the feature, but will this insert the identifier of the resource, or the alias itself?
I would be concerned about having the alias there, because it introduces a bunch of implications (e.g. what happens if someone changes the alias list in the note).

So my question is whether the alias is just used to lookup the resource in the completion list, or whether it becomes an identifier for the note?

riccardoferretti (commented):
Thanks @lingyv-li - I left a few comments, will play with the feature later today to gather more comments

packages/foam-vscode/src/core/model/workspace.ts
lingyv-li: > what happens if more than one resource declares the same alias?
Good point!

If a resource uses an alias that conflicts with another resource (in `_resources`), it will show warning `Resource identifier is ambiguous`.

If two resources have the same alias, currently it will not show warnings (as they will be overwritten here). I can make it a `Map<string, Set<Resource>>` to fix this.

packages/foam-vscode/src/features/link-completion.ts
lingyv-li: I intended to write the alias as the identifier here, just like suggested in the original issue. The alias identifier will be included in the `listByIdentifier`.

It will be quite complex if we want to support refactoring/renaming aliases. But I think that is not a must-have.

My personal reason for this is that I have some notes written in another software (Logseq) that already uses aliases all over the place 😜

riccardoferretti: > It will be quite complex if we want to support refactoring/renaming aliases. But I think that is not a must-have.

We support refactoring of note names, and I would expect aliases to be supported in that.
I agree it's not trivial, that's why I have been hesitant to add this feature in the past.

Using the alias just to identify a note, but then using the identifier when committing the text, could be an option but I reckon the main reason why people use aliases is not to look up a link, but to make it easier to read the text with the link in it? In that case the feature would not add a lot of value.

I am curious to poll the community on this one, what are your thoughts?

lingyv-li: > We support refactoring of note names, and I would expect aliases to be supported in that.

IIUC, we currently support updating note filenames and sync to wikilinks. It's slightly tricker for alias because it's in the file content and should use the RenameProvider API. We need to parse to the exact alias as well (supporting comma separated string, trimming whitespaces etc). I don't think it's a must have because the alias is a list, users can easily add additional aliases. Doing a global search and replace is also easy.

For reference, both [Logseq](https://discuss.logseq.com/t/add-ability-to-rename-an-alias-and-all-of-its-instances/3983) and [Obsidian](https://forum.obsidian.md/t/automatically-update-links-when-aliases-are-renamed/9323) doesn't support renaming aliases yet.

> Using the alias just to identify a note, but then using the identifier when committing the text, could be an option but I reckon the main reason why people use aliases is not to look up a link, but to make it easier to read the text with the link in it? In that case the feature would not add a lot of value.

It's an option to generate `[[identifier|alias]]`, but it doesn't look as pretty as we don't have WYSIWYG in vscode.

riccardoferretti: I think the `[[identifier|alias]]` option is actually pretty good, as it addresses most points (except the WYSIWYG) I am concerned about, including one I forgot to mention before: `[[alias]]` is not a valid wikilink (it requires Foam metadata to work, and is very brittle).

So let's do what Obsidian does.
To improve UX we can evaluate having custom decorations for links that have aliases - e.g. this is what I do with my regular links to highlight the label:

image

lingyv-li: Good point, I'll do `[[identifier|alias]]` instead.

packages/foam-vscode/src/features/link-completion.spec.ts
```diff
@@ -157,4 +157,26 @@ Content of section 2
...
+    const provider = new SectionCompletionProvider(ws);
+
+    const links = await provider.provideCompletionItems(
+      doc,
+      new vscode.Position(4, 2)
+    );
+
+    expect(new Set(links.items.map(i => i.label))).toEqual(
+      new Set(['alias-a'])
+    );
```
riccardoferretti, line 180: should we add a check that the text inserted is the identifier of the note and not the alias?

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -197,7 +197,20 @@ export class CompletionProvider
       item.range = replacementRange;
       item.command = COMPLETION_CURSOR_MOVE;
       item.commitCharacters = linkCommitCharacters;
-      return item;
+      const aliasItems = resource.aliases.map(a => {
```
riccardoferretti, line 200: I wonder if we should just have
```
const resources = ...
const aliases = ...
...
return new vscode.CompletionList([...resources, ...aliases, ...placeholders]);
```

This way we are sure all resources are shown before the resources

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -197,7 +197,20 @@ export class CompletionProvider
...
+        const item = new ResourceCompletionItem(
+          a.title,
+          vscode.CompletionItemKind.Reference,
+          resource.uri
+        );
+        item.insertText = this.ws.getIdentifier(resource.uri) + '|' + a.title;
+        item.detail = `Alias of ${resource.uri.getName()}`;
```
riccardoferretti, line 207: Like we do for resources, I would show `vscode.workspace.asRelativePath(toVsCodeUri(resource.uri));`

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -197,7 +197,20 @@ export class CompletionProvider
...
+        item.range = replacementRange;
+        item.command = COMPLETION_CURSOR_MOVE;
+        item.commitCharacters = linkCommitCharacters;
```
riccardoferretti, line 210: what happens if someone commits with the `|` char?

riccardoferretti (commented):
Looks good, I have left a couple more pointers, I think we are getting there!

lingyv-li:
Hi @riccardoferretti I have addressed all open comments.

btw, it seems like the CI doesn't show failure when there are failed test: https://github.com/foambubble/foam/runs/6986244381?check_suite_focus=true

riccardoferretti (approved):
LGTM!

riccardoferretti:
@allcontributors add @lingyv-li for code

riccardoferretti:
> btw, it seems like the CI doesn't show failure when there are failed test:  https://github.com/foambubble/foam/runs/6986244381?check_suite_focus=true

mmm.. you are right, I wonder when this started, need to investigate

arden13:
I'm not trying to reopen, but it's not clear to me from the docs how to add aliases to a file. Any recommendations for examples?

riccardoferretti:
@arden13 there are 2 ways for adding aliases. Either at the link level `[[wikilink|alias]]` or at the note level, using the `alias` property:

```
---
alias: alias 1, alias 2, alias3 
---
Hello world
```

I realized that the documentation for the second approach is actually missing, will add that to the documentation

arden13:
Thanks! I figured it out in the meantime, I didn't realize the yaml properties had to be above the first heading.

Do the aliases work both as a comma delimited list as well as a hyphenated list?

```
---
alias: this, being, commas
---
```
vs
```
---
alias:
  - this
  - being
  - hyphenated 
---
```

riccardoferretti:
Yes they do :)

PR #1015, opened 2022-06-18
Bundle js release with esbuild

lingyv-li:
Following https://aka.ms/vscode-bundle-extension
Before this `yarn vscode:package-extension` doesn't work for me. Getting things like
>  ERROR  invalid relative path: extension/../../.all-contributorsrc
> error Command failed with exit code 1.

This produces a smaller .vsix package (from >10 MB that are mostly screenshots to ~720 KB).
Also the `yarn vscode:package-extension` becomes a lot faster.


riccardoferretti:
Thanks for this @lingyv-li - I totally missed the screenshots!

I want to ping @joshdover as he has been doing some work on the web extension and the bundler will definitely play a role there, let us know if you have any comment.
From the docs I have seen there should be no issues in using esbuild (and generally I am not a huge fan of webpack, even though that's what they use in the web-extension example), but a second pair of eyes will help.

joshdover:
This looks great. I've used `rollup` for other extensions, but `esbuild` should work too. I see we're currently using `--platform=node` here. Curious if we'll need a separate build for `--platform=browser` or if we'll be able to share the same build. @lingyv-li have you experimented with this?

riccardoferretti (approved):

PR #1016, opened 2022-06-18
proof of concept for disabling markdown features

BrighTide:
Hi ^^

Indentation is a big part of my note taking flow, and I'm on a quest to support it in my instance of foam.

Namely this involves removing the code-block markdown feature with https://www.npmjs.com/package/remark-disable-tokenizers so my markdown gets interpreted as I want even though it's indented

I've been able to do so by patching the source files, but it would be nice to not have to do that every time i install a foam update 😄

I've drafted up a setting that would enable one to disable remark features by providing a list of tokenisers to disable. Figured I might solicit comment before finishing it. Is this something you'd be interested in merging?

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -439,6 +439,7 @@
     "lodash": "^4.17.21",
     "markdown-it-regex": "^0.2.0",
     "micromatch": "^4.0.2",
+    "remark-disable-tokenizers": "^1.1.0",
     "remark-frontmatter": "^2.0.0",
     "remark-parse": "^8.0.2",
     "remark-wiki-link": "^0.0.4",
--- a/packages/foam-vscode/src/core/services/markdown-parser.ts
+++ b/packages/foam-vscode/src/core/services/markdown-parser.ts
@@ -4,6 +4,7 @@ import unified from 'unified';
 import markdownParse from 'remark-parse';
 import wikiLinkPlugin from 'remark-wiki-link';
 import frontmatterPlugin from 'remark-frontmatter';
+import remarkDisableBlocks from 'remark-disable-tokenizers';
 import { parse as parseYAML } from 'yaml';
 import visit from 'unist-util-visit';
 import detectNewline from 'detect-newline';
@@ -14,7 +15,7 @@ import { Range } from '../model/range';
 import { extractHashtags, extractTagsFromProp, isSome } from '../utils';
 import { Logger } from '../utils/log';
 import { URI } from '../model/uri';
-import { getDisabledMarkdownFeatures } from '../../settings';
+import { DisableMarkdownFeaturesSetting } from '../../settings';
 
 export interface ParserPlugin {
   name?: string;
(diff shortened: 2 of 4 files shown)


riccardoferretti:
Hi @BrighTide I am not sure I fully understand what you are trying to achieve, can you provide examples/screenshots of how things are and would change?

BrighTide:
Hi @riccardoferretti, I'm proposing to add a setting that allows users to disable certain markdown features in remark.

In this instance, I want to disable the "raw-block" feature so I can can get the following effect:

image

Note that my link is still identified as a link even though it's indented, which it wouldn't be if it was being identified as code.

To this end I've added the "remark-disable-tokenizers" plugin and am planning to wire it up to a setting that lets the user list markdown features they want to disable (in my case it will be "raw-block")

I started implementing this, but thought I'd check to see if you had any objections / desires that it be a certain way :)

riccardoferretti:
I think I understand. Semantically speaking I would consider that a "allow links in code blocks" setting from Foam's POV.
I would like to know how much interest there is in such feature, I am not very keen to add settings to enable specific use cases that are generally not needed (especially if they come from a "stretch" of markdown's semantics).

BrighTide:
That's quite understandable

So to be clear you don't like the general setting "disable markdown feature list" and allow the user to input an arbitrary list?

To elaborate a little more, I make heavy use of indentation and cold folding to organise my thoughts, I find this extemely helpful but markdown feature stop working once indented 

image

Also another thought, "allow links in code blocks" might be a little confusing, as what this option would functionally do is disable code blocks entirely

riccardoferretti:
Thing is most of those features (like the folding you are issues with) are managed by VS Code directly, not Foam.
We extract the metadata from the file content, but don't affect the main editor experience (except adding links), so there is not much we can do in this regard.

BrighTide:
Hi @riccardoferretti I think there's still some confusion regarding what I'm proposing 😄. I think I might leave this for now, and if there's ever any interest in this feature I'd be happy to discuss further.

chrisUsick:
While indentation isn’t part of my note taking flow, I’m happy to see
someone experimenting with extending foam to suit their needs. Since this
is completely additive, and opt-in behavior this would get my vote to merge
as it demonstrates foam’s ability to adapt to the community.
Of course, is should still fall within the foam project’s core
principles/goals/contributing guidelines/etc. For example, if it added
bloat to the project, incurred a performance penalty, or the code didn’t
follow the contribution guidelines it would be reasonable to not merge.

riccardoferretti:
Rereading this I will add $.02 

> there's still some confusion regarding what I'm proposing

possibly :) let's see if I got a better grasp on it now

> what this option would functionally do is disable code blocks entirely 

What I believe you mean here is to disable code blocks in the text editor, you don't want VS Code to read those as code blocks but just as indented text. This is not managed by Foam, but by VS Code directly, and I am not aware of ways to change it.
The only thing that Foam controls is the parsing of the links, so the workaround would be to enable them in code blocks, but as I mentioned before (and I have a feeling you agree) semantically this is not what you are after, and I don't love it as a solution because it feels link a bandaid (of the wrong type too).

Going back to your case, I think the easier solution here would be to use text files, I see VS Code does provide some basic folding support there, but of course you'd be missing other markdown features (e.g. heading, preview, formatting, ...).
The way Foam would be able to help there is by having some sort of support for text files (which it currently doesn't) so that you can still have your wikilinks in there.

filmerjarred:
Hi @riccardoferretti, @chrisUsick. Thank you for taking the time to get back to me!

Speaking personally, the big draw of using vscode to do PKM lays in the hackability. Being able to make 100's of little adjustments to make the whole system suit me is invaluable, I think this is why some people are fanatics about vim and org mode.

To @riccardoferretti, you are correct about what want (disabling the code-blocks markdown feature) 😊, but incorrect about my proposal for how to do this. 

The code blocks feature is managed by vscode, but also by the remark module. To get the behaviour I want, I currently:

1. Maintain my own fork of foam, where I add a line to disable the code-blocks feature in the remark-parse module in "packages\foam-vscode\src\core\services\markdown-parser.ts"
2. Patch the core vscode markdown module at "\Program Files\Microsoft VS Code\resources\app\extensions\markdown-basics\syntaxes\markdown.tmLanguage.json"

You can see below I've added a line to disable the code-blocks feature in remark:
image

This lets me use markdown features even with indendation
image

And lets mecollapse large blocks of notes like so:
image

This works for me for the moment. I'm happy to let this thread document the process for anyone who wants to do the same.

Issue #966, opened 2022-04-02
First class support for images and pdfs


alberto-lanfranco:
### Is your feature request related to a problem? Please describe.

In my workflow, pdfs and images are first class citizens in my note vault.
In Foam, at the moment, they are not properly supported in this way.

### Describe the solution you'd like

I would like to be able to link to them with wikilinks `[[]]` , and embed them with `![[]]`

### Describe alternatives you've considered

- images
  - linking and embedding are currently supported with `[]()` and `![]()`
- pdfs
  - linking is supported with `[]()` 

problems:
- this type of linking and embedding is not tracked in graph view
- syntax highlight is broken for linking to local files in this way (see attached screenshot)

### Screenshots or Videos

<img width="326" alt="image" src="https://user-images.githubusercontent.com/57365132/161382716-918e9dfd-692c-482d-a1ca-44b63f03381b.png">

femto-code:
+1
I already tried a PDF extension ([link](https://marketplace.visualstudio.com/items?itemName=tomoki1207.pdf)) but unfortunately it does not support linking a specific page in a PDF...

riccardoferretti:
I agree that this needs to be implemented for v1, adding to backlog. If someone wants to help please ping me

PR #1027, opened 2022-07-07
Support for attachments 

riccardoferretti:
Fixes https://github.com/foambubble/foam/issues/966

Images and PDF files are supported as attachments.
They can be:
- linked
- embedded (although PDF files can't be rendered)
- refactoring (e.g. changing the name of an image file will cause links to be updated)
- autocompletion

Also shows first use of a second `ResourceProvider`, which was a good way to test the design


riccardoferretti:
Also added a property to embed a note flat (as it was before) or within a container

PR #1049, opened 2022-08-05
Add FOAM_DATE_WEEK for ISO_8601 week number

dmurph:


riccardoferretti (commented):
This looks good, can you please add a couple of tests, also to document the expected behavior (e.g. jan 2 on a year where that would be week 53 of the previous year, or week 1 of the current year).

Speaking of which, I see that as being a possible issue for the workflow you are trying to implement (`${FOAM_DATE_YEAR}-${FOAM_DATE_WEEK}` might not return what you expect the first few days of the new year) - what do you think?

packages/foam-vscode/src/services/variable-resolver.ts
```diff
@@ -199,6 +200,28 @@ export class Resolver {
...
+                const thursday = date.getTime();
+
+                // Find January 1st
+                date.setMonth(0); // January
+                date.setDate(1);  // 1st
+                const janFirst = date.getTime();
+
+                // Round the amount of days to compensate for daylight saving time
+                const days = Math.round((thursday - janFirst) / 86400000); // 1 day = 86400000 ms
+                resolve(Math.floor(days / 7) + 1);
```
riccardoferretti, line 221: I think this should be in 2 digits format (e.g. `07`) to stay consistent with e.g. the date variable
(we should add a test for this)

riccardoferretti:
superseded by #1053

PR #1061, opened 2022-08-28
Clicking the placeholder tooltip asks for template for new note

### LINKED ISSUE #995: Allow users to select the foam template as creating a new note by clicking on the link in the preview popup
### Is your feature request related to a problem? Please describe.

Steps to produce:
1. Open a foam project which contains foam templates, then open a markdown file.
2. Add a new wiki link, ex. [[wiki-link| Wiki Link]], the associated markdown file was not created.
3. Hover on the wiki link.
4. Wait to see the preview popup, then click on `Create note for 'wiki-link'`.
5. The file was created without asking user to select the foam template.

### Describe the solution you'd like

As a user, I expect I can select the foam template after clicking on the link, similar to the feature `Create new note from template`.

### Describe alternatives you've considered

I have some suggestions:
1 Change behaviour on clicking on the preview popup link.
2. Keep it, but add a new link to allow create a new note from the foam template.
3. Add a setting to allow users to choose the behaviour they want.
 

### Screenshots or Videos

_No response_

Dominic-DallOsto:
Fixes #995

Happy to work on this once we decide the best way of going about it.

At the moment this is just a minimal example with a lot of copy-pasting from `create-note-from-template.ts` (should the template selection logic be migrated to `templates.ts`, since at least some is already duplicated, and now a lot more?).

Current behaviour
* clicking the placeholder tooltip text - you are asked which template you want to use, then the note is created
* Holding Ctrl and hovering over the placeholder - the new note is already created with the default template before clicking. This seems undesirabled?

For me at least, the majority of new notes I create are using a template so some option that makes this quickly accessible would be great. I think Ctrl+Alt+Click would be a bit painful in this respect?


riccardoferretti:
Thanks for looking into this @Dominic-DallOsto 
I like the idea of optionally providing the user with the choice of template to use for a placeholder.
Currently the `new-note.md` template would be used, I think that is fine as a default behavior.

> clicking the placeholder tooltip text - you are asked which template you want to use, then the note is created

Simply clicking I reckon should just use the `new-note.md` template, which is the current behavior

> Holding Ctrl and hovering over the placeholder - the new note is already created with the default template before clicking. This seems undesirabled?

I wonder if more than ctrl/cmd we should use alt/option?
Regardless of the modifier, this is the place where I would give the user the option to select the template to use

Thoughts?

Dominic-DallOsto:
I think a nice workflow would be something like this
* clicking the link (by holding Ctrl) creates a new note with the default template (works* at the moment)
* there's a tooltip when hovering over the link - something like "create new note" (works at the moment)
* there's a second tooltip when hovering over the link - something like "create new note with custom template" (new to add)

I don't think we can use Alt as a click modifier? At least for me it's used to place multiple cursors

Essentially, this would boil down to adding a Hover provider to placeholder wikilinks for the "create new note with custom template". We could also add a setting to specify the default template, or continue to use the existing default.
* Additionally, the fact that just hovering over the placeholder link and holding Ctrl already creates a new note doesn't seem like desired behaviour. Is this because of a Go To Definition provider?

Does that make sense?

riccardoferretti:
> there's a second tooltip when hovering over the link - something like "create new note with custom template" (new to add)

To me is a trade-off of clarity vs clutter, and I believe in this specific case it can make sense to add it - but I'd have to see it in action to give e final answer to that

> I don't think we can use Alt as a click modifier? At least for me it's used to place multiple cursors

Haven't tried this, for me as long as we have a modifier I think it's fine, it just needs to be in line with the existing UX of VS Code 👍 

> Essentially, this would boil down to adding a Hover provider to placeholder wikilinks for the "create new note with custom template"

Interesting approach, I like it. Important I think is that the hover doesn't interfere with the "Also referenced in ..." content, nor the link hover (which it wouldn't). 
So basically the only way to create using a template would be by clicking on this hover. (if we implement that as a command I guess it could also be triggered by a keyboard shortcut)

> We could also add a setting to specify the default template, or continue to use the existing default

I am happy to simply use the default `new-note.md` template. In the future we might add a `new-note-from-placeholder.md` template if we think further customization is necessary.

> the fact that just hovering over the placeholder link and holding Ctrl already creates a new note doesn't seem like desired behaviour. Is this because of a Go To Definition provider?

This should not happen in Foam, are you sure you are not also using Markdown Notes? I have seen people report similar issues with that extension

Dominic-DallOsto:
Yep, Markdown Notes was causing that issue for me. Now things work.

image

What do you think?

I think the implementation could be refactored - I'm not sure the best place to modify the logic. At the moment I modified the `open-resource` command by introducing an extra parameter, but not sure this is the best way / might cause a problem if the command isn't built with `OPEN_COMMAND.asURI`?

Dominic-DallOsto:
I updated the hover provider tests - hopefully that works now. Maybe the one testing this functionality specifically will need to be updated still.

riccardoferretti (commented):
Thanks @Dominic-DallOsto, I really like the approach you have taken in the UI, simple and very effective.

For the code, I am just about to introduce a `create-note` command, which is probably what you will want to use here, and better integrates with the concept of "asking for the template". It shouldn't impact too much your PR, so I am holding the merge until that code is pushed so that we can leverage it. (that should happen today or tomorrow)

riccardoferretti:
FYI #1076

riccardoferretti:
The PR has been merged, thanks for the wait!
The work there should make yours much simpler here (forgive the conflict), let me know if the API works well for you

Dominic-DallOsto:
I rebased to master and updated the code.

I'm not sure if I'm missing something, but I can't get the new command to fit the workflow. The hover text needs to include a link to a command that will ask for the template, then create a note using it. But the new `create-note` function needs me to already provide the template. If I ask the user for the template when creating the link, the template pop up appears when hovering over the placeholder link (which makes sense, but isn't what we want).

Ideally what I need is a function like "ask user for template then create note", or some way to achieve that which currently is through hacking `open-resource.ts`  and `templates.ts`.

Or am I missing a simple way of wrapping `open-note` to do what I want?

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -108,8 +108,19 @@ export class HoverProvider implements vscode.HoverProvider {
         : this.workspace.get(targetUri).title;
     }
 
+    // If placeholder, offer to create a new note from template (compared to default link provider - not from template)
+    const command = OPEN_COMMAND.asURI(targetUri, true);
```
riccardoferretti, line 112: Have you considered using the `CREATE_NOTE` command here? It would be "more in line" with the purpose of the action (somewhat related: I do think the `OPEN_COMMAND` does a bit too much in its current implementation, that was a design flaw back when I wrote it)

packages/foam-vscode/src/services/templates.ts
```diff
@@ -246,15 +334,32 @@ export const NoteFactory = {
...
    * @param wikilinkPlaceholder the placeholder value from the wikilink. (eg. `[[Hello Joe]]` -> `Hello Joe`)
    * @param filepathFallbackURI the URI to use if the template does not specify the `filepath` metadata attribute. This is configurable by the caller for backwards compatibility purposes.
+   * @param chooseTemplate whether the user should be prompted to choose a template for the newly created note
    */
-  createForPlaceholderWikilink: (
+  createForPlaceholderWikilink: async (
     wikilinkPlaceholder: string,
-    filepathFallbackURI: URI
+    filepathFallbackURI: URI,
+    chooseTemplate: boolean
```
riccardoferretti, line 342: I was thinking here we might use the same pattern we have for `onFileExists`, and have an `async getTemplate` function that the client can pass in. The default implementation just returns the default template, but we can pass in your `askUserForTemplate` implementation when needed. wdyt?

riccardoferretti: (or maybe this can be handled at the `CREATE_NOTE` level, I guess we'd have to see it to get a good sense of it)

packages/foam-vscode/src/services/templates.ts
```diff
@@ -105,6 +113,86 @@ export type OnFileExistStrategy =
   | 'ask'
   | ((filePath: URI) => Promise<URI | undefined>);
 
+async function askUserForTemplate() {
```
riccardoferretti, line 116: I think it makes sense to move this function here, good call.
I would remove the old one and have this one being the only one in use

riccardoferretti: (same comment applies to the other functions here)

packages/foam-vscode/src/features/commands/open-resource.ts
```diff
@@ -20,7 +20,7 @@ const feature: FoamFeature = {
     context.subscriptions.push(
       vscode.commands.registerCommand(
         OPEN_COMMAND.command,
-        async (params: { uri: URI }) => {
+        async (params: { uri: URI; chooseTemplate: boolean }) => {
```
riccardoferretti, line 23: A `chooseTemplate` param in the `OPEN_COMMAND` doesn't feel right to me.
I know that I put too much responsibility in this command, that's the reason why I introduced the `CREATE_NOTE` command

riccardoferretti (commented):
Thanks for updating the PR, I left some comments based on the new `CREATE_NOTE` command that I just introduced.
Let me know if they make sense, thanks!

packages/foam-vscode/src/features/commands/open-resource.ts
Dominic-DallOsto: Yeah, logically it doesn't make a lot of sense - I was just hacking to get something to work. I can move this to the `CREATE_NOTE` command - I guess that makes more sense

packages/foam-vscode/src/services/templates.ts
Dominic-DallOsto: So the template can either be a template file (in which case we use it), a function (in which case we run the function to get a template), or nothing (then we choose the default template)?

I guess it'd work, but maybe this would be easier / make more sense at the `CREATE_NOTE` level like you say. Here we could just accept a template or choose the default one if it's not provided. Then `CREATE_NOTE` could have a way of asking the user for a template if it isn't provided?

packages/foam-vscode/src/features/hover-provider.ts
Dominic-DallOsto: Yeah, I think that makes more sense (ie. that's the point of the link here - to create a new note).

The create note command would just have to be modified so I can call it here in such a way that it asks the user for a template when run. I guess the command would either need an extra option to specify when it should ask the user for a template, or to also accept a function as the template argument, which would be called to get a template. Does that make sense / do you prefer either approach?

Dominic-DallOsto: It didn't work for me to pass a callback into `CREATE_NOTE_COMMAND.asURI` - it doesn't survive the `JSON.stringify`. So I guess an extra option is the best/only way?

riccardoferretti: We should look into the `asUri` function, there must be an issue somewhere there. I can take a look later tomorrow or the day after, but if you see the problem beforehand go ahead and fix it

packages/foam-vscode/src/services/templates.ts
riccardoferretti: That sounds about right. I am wondering whether there will be a problem with the `asUri` function, so maybe we need a string (like "ask") that is interpreted in the command as prompting the user.
The default behavior should be to use the default template, to ask the user the proper value needs to be provided.
We might need to change `templatePath` as it's becoming more than just the path. Maybe we can have a `template` property, which:
- if `ask` asks for the template
- otherwise will interpret the name as the path to the template

it should work?

Dominic-DallOsto:
Ok, it works now. I edited the parameters for `CREATE_NOTE` though. If you prefer, I can move it to `asUri`. Or fold it into `templatePath` like you say.

packages/foam-vscode/src/services/templates.ts
Dominic-DallOsto: Maybe would be weird if one wanted to create a note called `ask`? But otherwise should work the same

riccardoferretti: I guess you mean if you wanted a template called "ask"? I am ok with the tradeoff

For the `template` I am really thinking of having this priority list (out of scope for this PR, but helps you understand the thinking):

- not present: don't use template
- 'ask': ask the user
- <template-name>: if it matches the name of a template (from its metadata), use that
- <template-path>: if it matches the path of a file/template, use that
- otherwise don't use template (equivalent to not present)

I understand the risk of overlapping in the namespace, but the overlap amongst the options is quite small and "workaroundable" IMO, so I am ok with that.

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -108,8 +111,36 @@ export class HoverProvider implements vscode.HoverProvider {
...
+    }
+    const target = fromVsCodeUri(basedir)
+      .resolve(targetUri, true)
+      .changeExtension('', '.md');
+    const args = {
+      text: target.getName(),
+      notePath: target.path,
+      askForTemplate: true,
+    };
+    const command = CREATE_NOTE_COMMAND.asURI(args);
```
riccardoferretti, line 132: This all looks very good to me 👍

packages/foam-vscode/src/services/templates.ts
```diff
@@ -246,17 +343,24 @@ export const NoteFactory = {
...
    * @param wikilinkPlaceholder the placeholder value from the wikilink. (eg. `[[Hello Joe]]` -> `Hello Joe`)
    * @param filepathFallbackURI the URI to use if the template does not specify the `filepath` metadata attribute. This is configurable by the caller for backwards compatibility purposes.
+   * @param templateURI URI of the template to use. If undefined, use the default template.
    */
-  createForPlaceholderWikilink: (
+  createForPlaceholderWikilink: async (
     wikilinkPlaceholder: string,
-    filepathFallbackURI: URI
+    filepathFallbackURI: URI,
+    templateURI?: URI
```
riccardoferretti, line 351: Looks good

riccardoferretti (approved):
This looks very good to me.
Also your `askForTemplate` approach sounds pretty sensible to me, so no need to change it at this stage.

packages/foam-vscode/src/services/templates.ts
riccardoferretti: After looking at the code, I like your approach. 

I might update it in the future, with a proper deprecation/migration path (which I would need to do anyways for the `templatePath` prop), but it looks pretty solid now.

Dominic-DallOsto:
Ok, sorry for the delay. Tests are passing now on my end at least

riccardoferretti:
@allcontributors add @Dominic-DallOsto for code

riccardoferretti:
Amazing, thx!

PR #1065, opened 2022-09-02
add ignore root folder setting

### LINKED ISSUE #1063: Exclude whole repo/Folder when using foam in a multi-root workspace
### Is your feature request related to a problem? Please describe.

i want to have my notes repo and my work repo open simultaneously, but i dont want foam to index my work repo's docs (random jpgs, docs, etc)

### Describe the solution you'd like

notes-and-work.code-workspace:
```json
{
	"folders": [
		{
			"name": "notes",
			"path": ".."
		},
		{
			"name": "WORK",
			"path": "../../WORK"
		}
	],
	"settings": {
           "foam.files.ignore": [
             "WORK/**/*"
            ]
        }
```

or some such

### Describe alternatives you've considered

my current workaround is to manually exclude every top-level folder in the work repo, and make sure my notes folder doesnt have any folders with the same name, which sucks

### Screenshots or Videos

_No response_

asg0451:
Fixes #1063

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -283,6 +283,13 @@
           ],
           "description": "Specifies the list of globs that will be ignored by Foam (e.g. they will not be considered when creating the graph). To ignore the all the content of a given folder, use `<folderName>/**/*`"
         },
+        "foam.rootFolders.ignore": {
+          "type": [
+            "array"
+          ],
+          "default": [],
+          "description": "Specifies the list of folders that will be ignored by Foam (e.g. they will not be considered when creating the graph). This is useful in multi-root workspaces."
+        },
         "foam.logging.level": {
           "type": "string",
           "default": "info",
--- a/packages/foam-vscode/src/extension.ts
+++ b/packages/foam-vscode/src/extension.ts
@@ -7,7 +7,10 @@ import { Logger } from './core/utils/log';
 
 import { features } from './features';
 import { VsCodeOutputLogger, exposeLogger } from './services/logging';
-import { getIgnoredFilesSetting } from './settings';
+import {
+  getIgnoredFilesSetting,
+  getIgnoredRootFoldersSetting,
+} from './settings';
 import { fromVsCodeUri, toVsCodeUri } from './utils/vsc-utils';
 import { AttachmentResourceProvider } from './core/services/attachment-provider';
 import { VsCodeWatcher } from './services/watcher';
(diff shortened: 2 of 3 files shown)


riccardoferretti:
Thanks for taking a stab at this @asg0451 

I understand the approach taken, but I am concerned that it "splits" the ignoring of files/folders into two different settings, and it only provides an on/off switch to supporting multiple workspace folders.

Have you considered a way to "incorporate" this into the behavior of `foam.files.ignore`?

In my originally (although limited) thinking about this issue, I was wondering whether in case there is more than one workspace folder, the glob would (not) include the folder name. For example, given:

```
notes/
  family/
    family.md
  business/
    business.md
```

and

```
work/
  company/
    file.md
  side-project/
    project.md
```

I would expect something like:

1. if just the `notes` folder is open, use `foam.files.ignore: ["family/**"]`
2. if a workspace with both `work` the `notes` folders is open, use `foam.files.ignore: ["notes/family/**", "work/company/**"]`

An alternative, which I don't know if I like tbh, is to have a map of folder to ignore patterns:

```
foam.files.ignore: {
  "notes": ["family/**"],
   "work": ["company/**"]
}
```

But honestly the array approach seems more sensible to me.

Let me know your thoughts, I admittedly haven't given this a proper deep dive, so I might be missing some edge cases

asg0451:
hey @riccardoferretti, i understand your thinking there. my thought was that changing the meaning of the `foam.files.ignore` setting's elements depending on multirootness would be surprising in a bad way. though it might be clear if set in a named workspace file, users may set it elsewhere, or not even use named workspace files.

i also don't love the map idea, as it doesn't make sense for regular workspaces.

what do you think?

riccardoferretti:
Yup, I can see the possible confusion, although I wonder if practically it would be clearer than the current situation - here is what I mean:

When the workspace has a single folder, you don't see the folder in the explorer, so intuitively you wouldn't use in your glob (which is also the behavior today, and the reason why on simple projects we haven't had issues).

image

When the workspace has more than a folder, then they show up in the explorer, so it does "feel" more right to include it in the glob

image

So the root folder "becomes" part of the explorer, and in the same way I can see how it could "become" part of the glob.

What do you think?

riccardoferretti:
Hi @asg0451 are you still looking into this?

asg0451:
Not at the moment, sorry. I had an even bigger issue with multiroot
workspaces and switched to just having 2 windows 💀

once I have time to resolve that, I can look back into this. if you just
want to tap it in tho I won't mind 🤷‍♂️

riccardoferretti:
What other problem did you have? can you file an issue for that?

asg0451:
it wasn't to do with foam at all. vscode just started taking forever to start up in multiroot mode.

riccardoferretti:
I followed my approach in #1083, happy to hear comments

riccardoferretti:
replaced by #1083

asg0451:
looks good to me man, sorry i didnt follow through on it. i'm still off of multiroot, but maybe i'll try it again with this change 👍

Issue #1095, opened 2022-10-27
Link Preview with image


lucify123:
### Is your feature request related to a problem? Please describe.

Link Preview feature is really great.But it can not display images.

### Describe the solution you'd like

It will be much better if we can preview images when link previewing.
The image is intuitive and must be important when it is at the link previewing position.

### Describe alternatives you've considered

Compensation behavior: jump into links, from the preview page we can find the images.

### Screenshots or Videos

![preview no img](https://user-images.githubusercontent.com/19730021/198299376-8df53469-09ce-4a2e-9c52-ff764d600b85.png)
![微信截图_20221027213556](https://user-images.githubusercontent.com/19730021/198299383-7f751e6c-e3e5-4e5a-ba7b-eb61ce6f1ff8.png)

PR #1098, opened 2022-11-01
Fixed regex for image placeholder in hover

riccardoferretti:
The regex that was replacing the image was no longer working.
It's been replaced with a more specific regex.
To keep compatibility with past expected behavior we are still not showing the image, but just a reference to it.
We could change that in the future (PR welcome) but for now I consider enough to remove the broken image from the preview. Fixing the image would probably require a bit more work, to properly locate the file and find a relative path from the source (or maybe an absolute would work?)


riccardoferretti:
will partially fix #1095

PR #1099, opened 2022-11-03
Fix broken markup.

jonathanpberger:
Markdown was improperly formatting a line as a table (because of a "|" pipe being interpreted as a table delimiter instead of a logical OR?)

### DIFF
--- a/docs/user/features/tags.md
+++ b/docs/user/features/tags.md
@@ -11,7 +11,7 @@ You can add tags to your notes to categorize or link notes together.
 There are two ways of creating a tag:
 
 - Adding a `#tag` anywhere in the text of the note, for example: #my-tag1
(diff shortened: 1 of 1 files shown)


sn3akiwhizper:
I was looking at this as well, and the "|" pipe within the wikilink is meant as an alias for the file name and was, AFAIK, recently implemented in PR #1059. However, the "click the link to open note" functionality wasn't working for me on this option, and I am still unclear if the alias must be defined in the target file's front matter or if it is handled internally within Foam's code.

sn3akiwhizper:
Ah, after looking closer, the ordering of the link title and alias are backward so instead of `[[note property|note-properties]]` it should be `[[note-properties|note property]]`

PR #1105, opened 2022-11-12
Support tags with leading '.' and '#' within tag

msp301:
Tags like '#.NET' and '#C#' are nice for their brevity over their longer written forms.


riccardoferretti:
I am not completely against this, but would want to see more interest before moving forward it with it

PR #1107, opened 2022-11-15
Fix #932: Exclude custom folders from Graph

### LINKED ISSUE #932: Exclude _specific_ tags/notes from the graph
It'd be really nice if we could exclude notes with certain tags excluded from the graph. My use case for this would be the daily notes. My daily note template has the `#daily` tag, and my graph would be much cleaner if I could exclude all my daily notes.

Another option is to exclude a whole filter, which would achieve the same goal for me but it'd be much less flexible.

theAkito:
Resolves #932.

Created this draft with marked locations, on which I would apply the core part of the patch.

I require assistance on how to proceed & what other locations I need, for example, for this to be user configurable.

### DIFF
--- a/packages/foam-vscode/src/core/model/graph.ts
+++ b/packages/foam-vscode/src/core/model/graph.ts
@@ -106,6 +106,7 @@ export class FoamGraph implements IDisposable {
     this.placeholders.clear();
 
     for (const resource of this.workspace.resources()) {
+      //TODO: Workspace locations for graph.
       for (const link of resource.links) {
         try {
           const targetUri = this.workspace.resolveLink(resource, link);
--- a/packages/foam-vscode/src/services/editor.ts
+++ b/packages/foam-vscode/src/services/editor.ts
@@ -143,7 +143,7 @@ export const createMatcherAndDataStore = async (
   excludePatterns: Map<string, string[]>;
 }> => {
   const excludePatterns = new Map<string, string[]>();
-  workspace.workspaceFolders.forEach(f => excludePatterns.set(f.name, []));
+  workspace.workspaceFolders.forEach(f => excludePatterns.set(f.name, [])); //TODO: Excluded locations.
 
   for (const exclude of excludes) {
     const tokens = exclude.split('/');
@@ -162,6 +162,7 @@ export const createMatcherAndDataStore = async (
   const listFiles = async () => {
     let files: Uri[] = [];
     for (const folder of workspace.workspaceFolders) {
+      //TODO: Filter not excluded locations.
       const uris = await workspace.findFiles(
         new RelativePattern(folder.uri.path, '**/*'),
         new RelativePattern(


riccardoferretti:
Hi @theAkito and thanks for looking into this.
The way I am thinking about this is a bit more complicated, because my goal would be to not add many configuration items for as many use cases, but rather provide the building blocks upon which users can customize their workflow.

In that regard, I have been wondering whether to achieve the goal of this issue we could provide optional parameters to the `show-graph` command, so that people can show only a subset of it. This would also allow to have multiple graph configurations, based upon the parameters to the command (which can be bound to different keys).

More in general, I am thinking of introducing the concept of "note filter" that can be used across Foam to modify the default behaviour of various components, such as the panels, commands, ...

Imagine for example the following:

```
{
  "key": "cmd+m",
  "command": "foam-vscode.show-graph",
  "args": {
    "filter": {
      "and": [
        { "include": "/projects/*" },
        { "tag": "#a" },
        { "tag": "#b" },
        {
          "or": [
            { "type": "house-project" },
            { "type": "car-project" }
          ],
        }
      ]
    }
  }
}
```

Does that make sense?

theAkito:
> Does that make sense?

Absolutely.

> More in general, I am thinking of introducing the concept of "note filter" that can be used across Foam to modify the default behaviour of various components, such as the panels, commands, ...


The example you show has only an include flag, if I understand correctly. Do you want to base exclusions implicitly by explicit inclusions or do you think it makes sense to have something like `include **/*` and `exclude bulk/*`?



> I should clarify that this is still in brainstorming mode and the reason why I am sharing it here is to see if/how it plays with the issue at hand.

Okay. I agree with your draft of the implementation.

Could you proceed to the next stage and apply it to the actual code? I would like you to let me gain insight on that, so I fully understand the design of the implementation, which allows me to complete this issue.

riccardoferretti:
> The example you show has only an include flag, if I understand correctly. Do you want to base exclusions implicitly by explicit inclusions or do you think it makes sense to have something like include **/* and exclude bulk/*?

Oops, I forgot to mention also the `not` operator, and possibly and `exclude` filter (which can be implemented as `{ not: {include: '/path/**}}` but maybe it's still clearer to have it explicitly)

> Could you proceed to the next stage and apply it to the actual code? I would like you to let me gain insight on that, so I fully understand the design of the implementation, which allows me to complete this issue.

I do have a super draft design/implementation (but I am no expert in this type of query design, so I don't claim my approach is particularly clever here). Basically the idea is to use that query to build a filter function that all resources are filtered through. Here is the only code I have written:

```
interface ResourceFilterOp {
  and?: ResourceFilter[];
  or?: ResourceFilter[];
  not?: ResourceFilter;
}

/**
 * A set of params to filter resources.
 * If multiple attributes are specified, the filter will
 * check that they are ALL satisfied.
 * To achieve an OR-like behavior, use the OR operator instead.
 */
interface ResourceFilterParam {
  /**
   * A glob of the notes to include
   */
  include?: string;

  /**
   * A tag
   */
  tag?: string;

  /**
   * A note type
   */
  type?: string;
}

type ResourceFilter = ResourceFilterOp | ResourceFilterParam;

const hello: ResourceFilter = {
  tag: '#a',
  type: 'test',
};

const _: ResourceFilter = {
  and: [
    { tag: '#a', type: 'test' },
    { tag: '#b' },
    {
      or: [
        { type: 'daily-note' },
        { type: 'weekly-note' },
        { include: '/projects/**' },
      ],
      not: {
        type: 'root',
      },
    },
  ],
};

const operators = {
  or: (fns: ((Resource) => boolean)[]) => (resource: Resource) => {
    for (const fn of fns) {
      if (fn(resource)) {
        return true;
      }
      return false;
    }
  },
  and: (fns: ((Resource) => boolean)[]) => (resource: Resource) => {
    for (const fn of fns) {
      if (!fn(resource)) {
        return false;
      }
      return true;
    }
  },
  not: (fn: (Resource) => boolean) => (resource: Resource) => {
    return !fn(resource);
  },
};

const filters: { [key: string]: (string) => (Resource) => boolean } = {
  type: (type: string) => (resource: Resource) => resource.type === type,
  tag: (tag: string) => (resource: Resource) =>
    resource.tags.map(t => t.label).includes(tag),
  include: (glob: string) => (resource: Resource) => true, // TODO
  exclude: (glob: string) => (resource: Resource) => true, // TODO
};

const FILTER_KEYS = Object.keys(filters);

const makeFilter = (
  operator: string,
  param: string
): ((Resource) => boolean) => {
  if (FILTER_KEYS.includes(operator)) {
    return filters[operator](param);
  }
};
```

theAkito:
> Basically the idea is to use that query to build a filter function that all resources are filtered through. Here is the only code I have written:

I see. So, I could leverage the existing filtering to create an additional filter, which applies the exclude functionality.

If that is correct, I'll look into it and try to make it work, as a first step.

riccardoferretti:
> If that is correct, I'll look into it and try to make it work, as a first step.

Yes that's correct, basically for me the goal there is to produce a function `const filterFn = (r: Resource) => boolean`, which can then be used to filter the workspace resources (imagine a `workspace.resources.filter(filterFn)`).

Let me know if you have any question!

theAkito:
Looking at this again, it's not entirely clear to me how I would compose & integrate the new filtering.

I was thinking of starting out with your query draft, though I'm not sure where in the project it would make sense to place it. Then, I am not sure, where it should be implemented.

I guess the filter draft you created belongs into `packages/foam-vscode/src/services` and it should be implemented in `packages/foam-vscode/src/core/model/graph.ts`?

riccardoferretti:
I was originally thinking of having it in `services`, as I am not sure (but I haven't excluded) that it belongs to the model - feels to me more like a "service" on the model.

To me the filter works on top of resources: it works on `Resource`s and produces `Resource`s
The closest model would be `FoamWorkspace`, not `FoamGraph`.

To me it could work in one of these two ways, the first being my preferred unless there are reasons to go for the second:
1. `foam.workspace.list().filter(filterFn)` - you get a bunch of resources and you return a subset of those. I would keep the filtering outside of workspace and create a new module
2. `foam.workspace.filter(filterObj)`, tbh I am not sure about it, but as a first pass unless it's necessary I would avoid the coupling with `FoamWorkspace`, although I wonder if that will be necessary to later produce a coherent graph with the filtered subset of resources (but we can think about it after the first draft imo)

> packages/foam-vscode/src/features/panels/dataviz.ts would probably need to make those optional parameters available

In commands this is achieved in two ways:
1. we can define a keyboard shortcut
```
{
  "key": "cmd+m",
  "command": "foam-vscode.show-graph",
  "args": {
    "filter": {
      "and": [
        { "include": "/projects/*" },
        { "tag": "#a" },
        { "tag": "#b" },
        {
          "or": [
            { "type": "house-project" },
            { "type": "car-project" }
          ],
        }
      ]
    }
  }
}
```
2. you can have a setting which would be the default params for the command, if no others are defined

To start with I prefer the first approach, as I am not a big fan of having too many configuration settings.

To summarize, this is my advice:
- start in `services/filter.ts`, implement the logic there
- the service is used independently of `FoamWorkspace` and `FoamGraph`, it simply `(resources: Resource[]) =>.Resource[]` - if we want we can then use those resources to create a new workspace/graph
- in the case of the dataviz, the filter is defined as a parameter of the `foam-vscode.show-graph` command

Ok, this is a lot, let me know if it's clear or you have any question! :)

theAkito:
Thank you, this already helps tons!



> 1. we can define a keyboard shortcut

Are there shortcuts already available? I found none.
If this is the first shortcut, I would need to look up how to add one properly to the project.



> * the service is used independently of `FoamWorkspace` and `FoamGraph`, it simply `(resources: Resource[]) =>.Resource[]` - if we want we can then use those resources to create a new workspace/graph
> 
> * in the case of the dataviz, the filter is defined as a parameter of the `foam-vscode.show-graph` command

Does that mean, the shortcut is modified by the user to have additional excluded directories and, whatever method is used to retrieve the Resources, just automatically retrieves the filtered list, which again means, that no modification in `FoamWorkspace` & `FoamGraph` is needed?



> * in the case of the dataviz, the filter is defined as a parameter of the `foam-vscode.show-graph` command

There are no parameters yet, are there?
Or are the parameters including the already existing search/filtering option?

If there are none, I would also need to check out how to properly add some.

riccardoferretti:
> Are there shortcuts already available? I found none.

A good starting point to see how these shortcuts work is https://github.com/foambubble/foam/blob/master/docs/user/features/commands.md
Currently Foam doesn't define any shortcut upfront, it's the user that needs to define them in their settings for the situations they care about.

> Does that mean, the shortcut is modified by the user to have [...]

It means that if I want e.g. to see the graph only for a subset of the notes, I will create a keyboard shortcut for the `foam-vscode.show-graph` command that contains as argument the filter I want. The default behavior of commands doesn't change (that is, the change is backward compatible, and providing argument is a more "advanced" feature built on top of the existing functionality).

> [...] that no modification in FoamWorkspace & FoamGraph is needed?

No changes are needed in those two classes. We might need them to then do something with the filtered list of resources, but the classes themselves shouldn't need manipulation. 
Using "show graph" as an example: you can see that in [generateGraphData](https://github.com/foambubble/foam/blob/7f8c58084b06d3ded80755661436587b20bc0b0f/packages/foam-vscode/src/features/panels/dataviz.ts#L66) we need the `FoamGraph` to see the connections. I reckon what that means is that, if a filter is passed to the command, the command will:
1. compute the filtered resources from the "main" `FoamWorkspace` (e.g. `foam.workspace.list().filter(filterFn)`)
2. use those resources to compute a new temporary `FoamWorkspace`, which is used to compute a new temporary `FoamGraph`
3. use the graph to generate the graph data for the webview

> There are no parameters yet, are there?

That command has no parameters.
You can look at the [create note command](https://github.com/foambubble/foam/blob/master/packages/foam-vscode/src/features/commands/create-note.ts) for how to approach the implementation.


I hope this helps, let me know if anything is unclear

riccardoferretti:
https://github.com/foambubble/foam/pull/1158

riccardoferretti:
Closing as this will be implemented using note filters

PR #1118, opened 2022-12-05
add note root dir

jiangbo0216:
when I use foam, I wanna create note in my specify folder whenever which folder I in or which file I open.

this is a raw implementation, need more  improvement

eg:
1. config name
2. file name
3. commit format

from a foam amateur who wanna improve foam，please give some advisor  @riccardoferretti


riccardoferretti:
Hi @jiangbo0216 and thanks for taking a look at this.
The approach we want to take is about supporting variables in the configuration so that the user can specify where new notes should be created.
Imagine having something like `$WORKSPACE_ROOT` or `$CURRENT_DIR`.
This way we can reuse the same command and make it more flexible based on the user's needs.

jiangbo0216:
Straightforward commands are better，so I like a new command for workspace.
what I can do to complete this feature, I can give some gif or test for this feature.

our goal is to make foam better

one scene is that I want to take some note when I work in another project, that why I want this feature

riccardoferretti:
This will be superseded by https://github.com/foambubble/foam/pull/1142

PR #1125, opened 2022-12-22
Feat #903: Added support for emoji tags

### LINKED ISSUE #903: Allow using emoji in tags

badsketch:
Capture
Resolves #903; This should allow for most emojis to be supported with pure emoji tags, text + emoji tags, and nested emoji tags. I opted for `\p{Emoji_Presentation}` over `\p{Emoji}` because (TIL) numbers are emojis and the latter would permit pure number tags `#123` through, which we don't want.

This has weird behavior with keycap sequences such as 1️⃣, 9️⃣, as these are encoded as a digit, a variation selector, and enclosing number cap. So currently, a note with `#example1️⃣` and a note with `#example1` would result in sharing the same tag. I thought maybe this would be an edge case, but in case we do want full support, I can look into adding that, but it would probably mean adding something harder to read to the existing regex like `(\u00a9|\u00ae|[\u2000-\u3300]|\ud83d[\ud000-\udfff])`. 

Also, let me know if the tests are sufficient. I could add an e2e test to `tags.test.ts`, but wasn't sure since those test more general tag functionality than the type of tags. 

Appreciate any feedback!

### DIFF
--- a/packages/foam-vscode/src/core/utils/hashtags.ts
+++ b/packages/foam-vscode/src/core/utils/hashtags.ts
@@ -1,6 +1,6 @@
 import { isSome } from './core';
-const HASHTAG_REGEX = /(?<=^|\s)#([0-9]*[\p{L}/_-][\p{L}\p{N}/_-]*)/gmu;
-const WORD_REGEX = /(?<=^|\s)([0-9]*[\p{L}/_-][\p{L}\p{N}/_-]*)/gmu;
+const HASHTAG_REGEX = /(?<=^|\s)#([0-9]*[\p{L}\p{Emoji_Presentation}/_-][\p{L}\p{Emoji_Presentation}\p{N}/_-]*)/gmu;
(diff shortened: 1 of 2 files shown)


riccardoferretti (approved):
Thanks for the amazing work @badsketch

Everything you say makes sense, the test coverage looks good, it looks great!

riccardoferretti:
Actually can you fix the lint errors? :)

badsketch:
Sure thing, pushed 👍

riccardoferretti:
@allcontributors add @badsketch for code

PR #1127, opened 2022-12-30
[WIP] Enable support for vscode.dev

pderaaij:
I started on enabling Foam for https://vscode.dev, the online version of VSCode. I decided to work in public to make this a broader effort. It's not an easy task and could use any help and knowledge.

Right now I have been able to get the webpack compilation going and enable a local test environment. It's not loading Foam currently. So tackling that one to see what remaining tasks are left.


pderaaij:
Okay, I have it working in the local online environment right now. Don't have time now to look what's working and what's not. But from first sight things seem to work. Feel free to check out the branch and play around.

pderaaij:
I did some further research, and this is what I found:

- Tag explorer is not loading the tags. 
   - Looks something is different with the way `this` is being handled. For example, when adding a tag to a note it does call the `refresh()` method correctly, but the resource does not contain any tags. The parsing of the file is done correctly. I haven't been able to figure out what's going on here. 
- Graph is not being displayed. 
   - Need to add some static files to the webpack config
- Need to do some testing with renaming and moving 

For production we need to look if all polyfills are required and to see if we can reduce the size of the bundle.

riccardoferretti:
Thanks for looking into this @pderaaij! 

I imagine you are familiar with this issue regarding this task, but to be sure and for documentation here it is: https://github.com/foambubble/foam/issues/749

Here are my notes from the latest spike I did on this some time ago:

- [x] fix packages `os`, `path`, `crypto`
- [ ] YAML module causes issues

Another question I have from your approach regards the use of webpack, I would love not to have two build/bundle systems in parallel, do you have a plan around that?
In the end a web extension is also a regular extension, so I guess we'd migrate everything to webpack? I need to check what the DX around that is like.

Beside this, I am gonna play with the branch and give more feedback.

Thanks!

riccardoferretti:
I have been playing with it, here is what I found:
- the `problemSolvers` require the `amodio.tsl-problem-matcher` extension (we need to note/document that)
- the `gray-matter` module is causing issues - simply returning `content` in `getNoteTooltip` helped me remove a bunch of warnings (helpful to get things working, but we'll need to find a solution)
- I haven't been able to get the debugger to work (not sure if I am missing something)
- I noticed that webpack produces a single `extension.js` file - today we produce a `js` file per `ts` module, which I personally find easier to navigate. I was trying to configure webpack to do that but without success

The combo of debugger not working and not being able to navigate the code kinda blocked me, so I couldn't find out a lot about the tags explorer issue.
To be noted that also other features were not working for me, like sections (`[[wikilnk#section]]`), backlinks panel and orphan panel. My guess is that the cause is the same for all these.

Thoughts?

pderaaij:
Thanks for looking into it! I'll have a look at debugging tomorrow. Looks like a trick with sourcemaps will help enable debugging.

riccardoferretti:
I found the cause of the tags and other issues, the way files were being read was not properly decoding.

Here is the fix.
In `editor.ts`, replace the `readFile` function with the following:

```
  const decoder = new TextDecoder('utf-8');
  const readFile = async (uri: URI) => {
    const content = await workspace.fs.readFile(toVsCodeUri(uri));
    return decoder.decode(content);
  };
```

pderaaij:
@riccardoferretti could it be that I'm missing something? I replaced the function and it's still not working for me. Did a couple of rebuilds and recompile, but still no luck.

pderaaij:
As a next step I am trying to get the end2end test working with the web extension. For some reason I am running into some polyfill issues in the compiled test bundle. I am debugging what's going on here and will report back when I have more information.

pderaaij:
Little progress update. I am not able to get the tests running for now. After a lot of fiddling around and testing I am able to find the source of my troubles. 

The tests point to `src/test/suite.ts`. This is the file that configures `jest` and run it on CLI. However, jest pulls in all kinds of depedencies that rely on functions that are not available in a browser. The examples all use mocha, which can run in a browser.

So.. I'm a bit lost how to navigate further. I haven't been able to find a way to run jest in a browser accepted manner. If anyone has an idea I'm all for it!

riccardoferretti:
I think it's ok if the tests themselves do not run in the browser, the integration tests are still interacting with the VS Code web extension, I consider that to be enough.
Technically speaking the unit tests are not running in the same environment, but I wouldn't consider that a blocker at this point.

riccardoferretti:
To me there are few open questions:

- [ ] getting `gray-matter` module to work (or find a replacement)
I believe this is the only functional delta we have atm?
- [ ] webpack vs tsdx/esbuild
 We shouldn't have both, either everything works on one or on the other.
 I don't really have a strong preference between them, but we need at least function parity to migrate and/or understand any delta.
- [ ] support for debugging (which I want to thoroughly test, I think we are getting there)



A note regarding a previous comment:

> I noticed that webpack produces a single extension.js file - today we produce a js file per ts module, which I personally find easier to navigate. I was trying to configure webpack to do that but without success

Nevermind this comment, as it's actually recommended to bundle the extension into a single file, so we should probably do it no matter what.

package.json
```diff
@@ -21,6 +21,7 @@
     "watch": "lerna run watch --concurrency 20 --stream"
   },
   "devDependencies": {
+    "@types/estree": "^0.0.51",
```
riccardoferretti, line 24: what is this dependency for?

packages/foam-vscode/package.json
```diff
@@ -493,26 +500,40 @@
...
     "tslib": "^2.0.0",
     "typescript": "^3.9.5",
     "vscode-test": "^1.3.0",
-    "wait-for-expect": "^3.0.2"
+    "wait-for-expect": "^3.0.2",
+    "webpack": "^5.38.1",
+    "webpack-cli": "^4.7.0"
   },
   "dependencies": {
+    "assert": "^2.0.0",
```
riccardoferretti, line 513: is this necessary because of `snippetParser.test.ts`?

packages/foam-vscode/src/services/editor.ts
```diff
@@ -175,8 +175,11 @@ export const createMatcherAndDataStore = async (
     return files.map(fromVsCodeUri);
   };
 
-  const readFile = async (uri: URI) =>
-    (await workspace.fs.readFile(toVsCodeUri(uri))).toString();
+  const decoder = new TextDecoder('utf-8');
```
riccardoferretti, line 178: I am planning to add this to master, so that this PR is focused on configuration changes required for the switch to web extension
(I consider the current implementation of this fn to be a bug that happens to work by chance)

packages/foam-vscode/web/webpack.config.js
```diff
@@ -0,0 +1,114 @@
...
+
+//@ts-check
+/** @typedef {import('webpack').Configuration} WebpackConfig **/
+
+const path = require('path');
+const webpack = require('webpack');
+
+/** @type WebpackConfig */
+const webExtensionConfig = {
+	mode: 'none', // this leaves the source code as close as possible to the original (when packaging we set this to 'production')
```
riccardoferretti, line 17: minor styling note: I see you use tabs, but the Foam codebase uses 2-spaces for indentation

tsconfig.base.json
```diff
@@ -9,6 +9,8 @@
     "paths": {
       "foam-core": ["./packages/foam-core/src"],
       "foam-vscode": ["./packages/foam-vscode/src"]
-    }
+    },
+    "typeRoots": ["node_modules/@types"],
+    "sourceMap": true
```
riccardoferretti, line 14: What are these for?
is it specific to the move to web extension or did you uncover some issue with the current configuration?

riccardoferretti (commented):
I left a few comments to better understand the scope of the changes that are strictly due to the migration to web extension VS other changes that can/should be done regardless but were exposed by the work you are doing.

Once we have identified those "extra" changes I will include them in master, so that the PR can focus on only what's needed to to migrate to the web extension

pderaaij:
So, this is becoming an interesting project. I'm still trying to get the end-2-end tests working for the browser. The setup of `vscode-test-web` is that the testsuite is a `webpack` packaged version of the test suite that runs in the browsers. Our test suite is based on `jest` which isn't designed for running in the browser. The samples of vscode all use `mocha` to run their test suite. I've tried to get both `mocha` and `jest` working, but that causes all kinds of problem. 

On the plus side, it seems that most polyfills I used were caused due to `jest` becoming part of the bundle.I expect to be able to clean up the webpack config a lot, when we got this sorted.

Options I see are:

- Do not run end-2-end tests for the web version
- Develop a specific test suite for the web version
- Migrate from `jest` to `mocha` for the entire project

@riccardoferretti curious to your thoughts here.

In #749 a member of the VSCode offered assistance, I am going to use that message to hook him in on this PR. Perhaps there might be some other ideas.

riccardoferretti:
Thanks for the thorough summary.
The way I was thinking about it is that e2e could still run in node (as a process), but they would nonetheless interact with a VS Code instance running the web extension.
I reckon this should be possible, but to be checked.

Hi @digitarald , see the comments above, it would be really helpful to get a bit of guidance here, what are your thoughts?
Specifically:
- how have extensions using jest for testing migrated to web extension?
- does it make any sense to have the setup with the e2e test suite running in node+jest with a VS Code instance running the web extension?
- any thought on webpack vs tsdx/esbuild/rollup?

And of course please share anything else that you think might be helpful

JakeGinnivan:
Rather than mocha, moving towards vite and vitest may be a better option. vitest has API compatibility with jest which makes the migration pretty easy. Though I'm not sure if it can run in vscode web.

pderaaij:
Working on it in #1290

PR #1131, opened 2023-01-03
Fix orphan, placeholder, tag-explorer counts

badsketch:
I think there's a slight bug with change https://github.com/foambubble/foam/commit/7a2756eb1db155c4d4b301d0b0e577a9eb2986f7, where the counts are out of sync with the actual number of entries

To reproduce in new project:
1. create note1
2. create note2 and add a placeholder [[note3]]
3. note that orphans panel says `Orphans (2)` even though it only has note1 and placeholders panel says `Placeholders (0)` with the note3 entry
Screenshot 2023-01-03 102024

I believe we just need to tweak the subscription logic so the panel title is set after we do the calculations for each panel.

### DIFF
--- a/packages/foam-vscode/src/features/panels/tags-explorer.ts
+++ b/packages/foam-vscode/src/features/panels/tags-explorer.ts
@@ -25,10 +25,10 @@ const feature: FoamFeature = {
     context.subscriptions.push(
       treeView,
       foam.tags.onDidUpdate(() => {
+        provider.refresh();
         treeView.title = baseTitle + ` (${foam.tags.tags.size})`;
       })
     );
-    foam.tags.onDidUpdate(() => provider.refresh());
   },
 };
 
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/src/features/panels/tags-explorer.ts
```diff
@@ -25,10 +25,10 @@ const feature: FoamFeature = {
     context.subscriptions.push(
       treeView,
       foam.tags.onDidUpdate(() => {
+        provider.refresh();
         treeView.title = baseTitle + ` (${foam.tags.tags.size})`;
       })
     );
-    foam.tags.onDidUpdate(() => provider.refresh());
```
badsketch, line 31: I actually didn't run into issues with tag count becoming out of sync during my testing, but I still made this change as I believe there could be a race condition. The title could be wrong if  `provider.refresh()` doesn't complete before title is generated. Thoughts?

riccardoferretti: There is no race condition here because we don't use the provider to compute the number, but the `foam.tags` object directly (also this is the reason why you didn't run into issues in the first place)

riccardoferretti (approved):
Thanks for looking into this, your change makes sense, it's good to go

Issue #1149, opened 2023-01-28
Support displaying wikilinks using target's note metadata like "link_label" instead of name if alias not present


yoctozepto:
### Is your feature request related to a problem? Please describe.

Currently, wikilinks render as their names unless an alias is given. The target's title is used only as the link title and not as the label. To be able to present the same link canonically in all places, one has to either keep the canonical representation in the filename or keep aliasing everywhere (and break the DRY principle).

### Describe the solution you'd like

As a user, I would like to make wikilink render using target note's metadata like ``link_label`` instead of the name if no alias is given.

### Describe alternatives you've considered

I considered aliasing everywhere, but it violates the DRY principle.

riccardoferretti:
Wikilinks in Foam use the filename (which might or not match the title of the note).
If you want a different text in the source document you can either use an ``[[link|alias]]` or a [reference definition](https://github.com/foambubble/foam/blob/master/docs/user/features/link-reference-definitions.md).

We have an [`alias` property](https://github.com/foambubble/foam/pull/1014) in the note that can be used to help with autocompletion (I admit this could do with some documentation), but there are [no plans](https://github.com/foambubble/foam/pull/1014#discussion_r902525686) to change the principle of having the wikilink to be a standard one (that is, one that matches the name of the file).

yoctozepto:
Dear @riccardoferretti, I think I was misunderstood. My proposal is nothing like the quoted one. I mean to add additional capabilities to the rendering side of things, not how the wikilinks are built and what reference system they use.

yoctozepto:
I proposed an early pull request showing what I mean to achieve. It works for me already. :-)

PR #1150, opened 2023-01-28
Add support for link_label property

yoctozepto:
link_label provides the capability to set the canonical way of displaying a wikilink to the note instead of defaulting to the link name.

Implements #1149

### DIFF
--- a/packages/foam-vscode/src/features/preview/wikilink-navigation.ts
+++ b/packages/foam-vscode/src/features/preview/wikilink-navigation.ts
@@ -34,7 +34,11 @@ export const markdownItWikilinkNavigation = (
           toVsCodeUri(resource.uri)
         )}${formattedSection}`;
         const title = `${resource.title}${formattedSection}`;
-        return `<a class='foam-note-link' title='${title}' href='/${link}' data-href='/${link}'>${label}</a>`;
+        const finalLabel =
+          isEmpty(alias) && !isEmpty(resource.properties.link_label)
+            ? `${resource.properties.link_label}${formattedSection}`
+            : label;
+        return `<a class='foam-note-link' title='${title}' href='/${link}' data-href='/${link}'>${finalLabel}</a>`;
       } catch (e) {
         Logger.error(
           `Error while creating link for [[${wikilink}]] in Preview panel`,


yoctozepto:
Open to discussion. It's missing proper documentation as well.

yoctozepto:
(and missing relevant tests)

riccardoferretti:
Thanks for looking into this, tbh I feel I don't fully understand the use case, can you give me a specific example?

yoctozepto:
One simple use case is to have url-friendly note filenames like `my-awesome-note` while being able to *always* display `My Awesome Note` instead (DRY). More sophisticated use cases involve acronyms (usually preferred at filename level but the functionality allows also for a vice-versa situation) and generally name shortening. In principle, this little change gives a lot of freedom to customise the default link display behaviour per-note. It has almost no performance penalty because all metadata is read already.

mmcnl:
I am just coming to Foam and very much like it so far (thank you @riccardoferretti for all your work on it). However I keep getting tempted to return to Obsidian because of its better support for formatting the wiki text. This PR sounds like it would be a good step in the direction of improved visual presentation; I'd be in support for this in Foam.

mmcnl:
Sure! As related to this specific example, I'd love for there to be a way, given a note named `jane-smith.md`, with some content, eg

```
# Jane Smith

text about Jane...
```

 to be able to reference that note from inside `other-note.md` in manner like this:
```
my link to [[jane-smith]]
```
and have it actually be rendered in the markdown preview like 



my link to [Jane Smith](jane-smith.md) 


as if the raw text were actually
```
my link to [[jane-smith|Jane Smith]]
```
Maybe the simplest way would be like how the Foam Graph will use the top-level heading (`#`) as the title in the graph rather than just the note filename. Another option could be parsing the YAML front-matter, eg
```
---
alias: Jane Smith
---
# Jane Smith

text about Jane...
```
so that Foam knows to default to using `Jane Smith` as the alias for a link to `[[jane-smith]]`. Of course one should still be able to override a particular instance with a different alias, eg `[[jane-smith|my friend Jane Smith]]` and have that render like

my link to [my friend Jane Smith](jane-smith.md)

riccardoferretti:
I love the idea of using the title of the note, will get that out by the weekend

mmcnl:
Sounds good!

riccardoferretti:
Done in 0.25.5

mmcnl:
Looks great! That will allow me to simplify my notes. Thank you @riccardoferretti!

PR #1160, opened 2023-02-10
Updated a bunch of dependencies

riccardoferretti:
It all started because I wanted to update the `vscode` engine version to support workspace trust.

That needed a more recent version of typescript.

That was causing some issues with linting, so I updated that.

That was causing issues with `tsdx`, which unfortunately is being discontinued, so I migrated to `dts` https://github.com/weiran-zsd/dts-cli 

That meant updating a bunch of other dependencies like jest, eslint, ...

At what point I was like: "well let's update as many as we can"

...

Ended up updating most dependencies (`unified` and family are a notable exception)

So I feel good about it, it builds, tests are passing...
...what could possibly go wrong


riccardoferretti:
[many commits later]

To get CI to pass I updated node version

which required to update lerna version
and required me to update the node version in the github action

which required to update the github action version

which required to update the ubuntu version

At which point the CI passed but the caching action was not working, so that was updated
And, why not, let's also update the windows and macos version

macOS was successful, there was an issue with Windows related to the glob library
so I updated the test code and killed the library

PR #1162, opened 2023-02-12
Fix Previews of Embedded Sections

Skakerman:
I've only recently started exploring Foam and when I was porting over some of my notes I noticed that when a note had a section of another note embedded in it the preview panel wouldn't behave as expected.

e.g.,
```md
# A Note

Here's a section from another note ![[another-note#My Section]]
```

I took a look around and it seems to be because when `foam.preview.embedNoteInContainer` is `true` (which is the default value), the embedding logic attempts to split the embedded note after it has already been rendered to HTML which breaks the mapping of the content to the section ranges.

It may be a naïve solution but I was able to fix it by moving the section-splitting before the HTML rendering for embedded notes. I'm still new to Foam and I've never done VS Code extension development before so I'm open to feedback!

| Broken section embedding |
| --- |
| foam-embed-section-broken |

| Fixed section embedding |
| --- |
| foam-embed-section-fixed |

### DIFF
--- a/packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
+++ b/packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
@@ -125,10 +125,12 @@ This is the third section of note E
       () => {
         const md = markdownItWikilinkEmbed(MarkdownIt(), ws);
 
-        const res = md.render(`This is the root node. ![[note-e-container#Section 3]]`);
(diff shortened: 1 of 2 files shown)


riccardoferretti (approved):
Looks great, thanks for contributing!

riccardoferretti:
@allcontributors add @Skakerman for code

Issue #1182, opened 2023-04-05
Tag expansion fires inside Markdown headings


jimgraham:
### Describe the bug

When writing a markdown heading prefixed with `#`, the tag expansion bubble opens up

### Steps to Reproduce the Bug or Issue

1. create a tag such as `decision-log`
2. in a separate file, type a markdown header like `## Must DO`
3. the tag-expansion bubble will appear with the `decision-log` tag

### Expected behavior

I expect that tag expansion is ignored in headings

### Screenshots or Videos

<img width="734" alt="tag-expansion" src="https://user-images.githubusercontent.com/430293/230211642-42f20477-c271-4366-82be-1810dd32de02.png">

### Operating System Version

OSX 13.4

### Visual Studio Code Version

1.77

### Additional context

my suspicion is that it occurs because the `TAG_REGEX` https://github.com/foambubble/foam/blob/d0ee71be1b95935fee6987b8f737d0c06a380f3c/packages/foam-vscode/src/features/tag-completion.ts#L8

is far too generous. Something that excludes `# text`. I assume tags can't have spaces, so `\S+` should work
```
TAG_REGEX = /#(\S+)/
```

I could add a test similar to https://github.com/foambubble/foam/blob/122db206956a36d145bea54041559a3b53e067a7/packages/foam-vscode/src/features/tag-completion.spec.ts#L83

jimgraham:
Possible fix in #1183

riccardoferretti:
Fixed in 0.21.2

PR #1183, opened 2023-04-05
Relax the TAG_REGEX to miss markdown headings

jimgraham:
Addresses #1182 

The `TAG_REGEX` was extremely greedy, and would include matches to markdown headings such as `## Heading 2` even though that could not be a tag.

Relax the regex to match on non-space characters only.

### DIFF
--- a/packages/foam-vscode/src/features/tag-completion.spec.ts
+++ b/packages/foam-vscode/src/features/tag-completion.spec.ts
@@ -93,4 +93,18 @@ describe('Tag Completion', () => {
     expect(foamTags.tags.get('primary')).toBeTruthy();
     expect(tags).toBeNull();
   });
+
+  it('should not provide suggestions when inside a markdown heading #1182', async () => {
+    const { uri } = await createFile('# primary heading 1');
+    const { doc } = await showInEditor(uri);
+    const provider = new TagCompletionProvider(foamTags);
+
+    const tags = await provider.provideCompletionItems(
+      doc,
+      new vscode.Position(0, 7)
+    );
+
+    expect(foamTags.tags.get('primary')).toBeTruthy();
+    expect(tags).toBeNull();
+  });
 });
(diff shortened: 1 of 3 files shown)


packages/foam-vscode/src/features/tag-completion.spec.ts
```diff
@@ -93,4 +93,18 @@ describe('Tag Completion', () => {
     expect(foamTags.tags.get('primary')).toBeTruthy();
     expect(tags).toBeNull();
   });
+
+  it('should not provide suggestions when inside a markdown heading', async () => {
```
riccardoferretti, line 97:
```suggestion
  it('should not provide suggestions when inside a markdown heading #1182', async () => {
```

riccardoferretti: Adding a reference to the issue solved by the test

packages/foam-vscode/src/features/tag-completion.ts
```diff
@@ -5,7 +5,7 @@ import { FoamFeature } from '../types';
 import { mdDocSelector } from '../utils';
 import { SECTION_REGEX } from './link-completion';
 
-export const TAG_REGEX = /#(.*)/;
+export const TAG_REGEX = /#(\S+)/;
```
riccardoferretti, line 8: in `hashtag.ts` there is the actual regex used to parse hashtags, could you try using that? Would love to consolidate the two

riccardoferretti (commented):
Thanks for reporting and fixing the issue, I have added a couple of minor things, than this is ready to go!

packages/foam-vscode/src/features/tag-completion.ts
jimgraham: updated in [7695819](https://github.com/foambubble/foam/pull/1183/commits/7695819a2ddc75093801980cb337a4143bc50bba)

jimgraham:
Thanks for the feedback. I've made the change and squashed my commits. 

I'm not sure about linting. `husky` is broken for me on a new install, so I did `git commit -n` ..

riccardoferretti (approved):

riccardoferretti:
The PR looks good to me. I see it's still in draft, is there anything that I am missing? Otherwise make it ready and I will merge it

riccardoferretti:
@allcontributors add @jimgraham for code

PR #1192, opened 2023-04-14
Allow for `#` alone to trigger tag completion

### LINKED ISSUE #1189: Tag completition doesn't work anymore after update
### Describe the bug

After the [v0.21.2](https://github.com/foambubble/foam/releases/tag/v0.21.2) when i start to add a tag to a note (#something) the extension doesn't show the auto completition hint as it did before.
I also try to install the [v0.21.3](https://github.com/foambubble/foam/releases/tag/v0.21.3) version but there is no change in that behavior.

_No response_

Create a new file with markdown syntax. Start adding a title and after try to add a tag that is present in the tag explorer.

Example of file content:

```
# Title

#<your-tag-here>
```
### Expected behavior

I expect that a tag selection windows is shown to aid the tag insertion. (previous operating mode)

### Screenshots or Videos

_No response_

### Operating System Version

OS Name	Microsoft Windows 10 Pro (Version 10.0.19045 Build 19045)

1.77.3

### Additional context

_No response_

jimgraham:
Closes #1189 

In #1183, I reused [HASHTAG_REGEX](https://github.com/foambubble/foam/blob/83a90177b921064c16e754a053f445ce79d6e9af/packages/foam-vscode/src/core/utils/hashtags.ts#L2-L3) to validate the tag line when the `CompletionProvider` was triggered.

I wanted to prevent this:

```markdown
# This is a Markdown header
```

but using the `HASHTAG_REGEX` had the side effect of requiring an _additional_ character to trigger the completion provider.

```markdown

1. #p <-- triggers completion
2. #  <-- does not trigger
3. #_ (space) <-- does not trigger
```
both 1. and 2. should have triggered.

To fix, I use a slightly different regex that uses a negative lookahead to ensure that the `#` is not followed by a space. I also added spec cases to cover this situation.

### DIFF
--- a/packages/foam-vscode/src/features/tag-completion.spec.ts
+++ b/packages/foam-vscode/src/features/tag-completion.spec.ts
@@ -66,6 +66,36 @@ describe('Tag Completion', () => {
     expect(tags).toBeNull();
   });
 
+  it('should provide multiple suggestions when typing #, issue #1189', async () => {
+    const { uri } = await createFile(`# Title
+
+#`);
+    const { doc } = await showInEditor(uri);
+    const provider = new TagCompletionProvider(foamTags);
+
+    const tags = await provider.provideCompletionItems(
+      doc,
+      new vscode.Position(2, 1)
+    );
+    expect(tags.items.length).toEqual(3);
+  });
+
+  it('should not provide a suggestion when typing `# `, issue #1189', async () => {
+    const { uri } = await createFile(`# Title
+
+# `);
+    const { doc } = await showInEditor(uri);
+    const provider = new TagCompletionProvider(foamTags);
+
+    const tags = await provider.provideCompletionItems(
+      doc,
+      new vscode.Position(2, 2)
+    );
+
+    expect(foamTags.tags.get('primary')).toBeTruthy();
+    expect(tags).toBeNull();
+  });
+
   it('should provide a suggestion when typing #prim', async () => {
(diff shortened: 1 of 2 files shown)


jimgraham:
I will record videos for mac osx and windows tomorrow ..

jimgraham:
Updated video showing this in action


https://user-images.githubusercontent.com/430293/232229298-0d72b8db-7dd3-4393-b16e-56208e44d16c.mp4

riccardoferretti (approved):
Thanks for adding the various cases, this looks good to me

jimgraham:
do you want me to squash commits?

riccardoferretti:
I found an odd edge case while testing locally.
 
tag-autocompletion

jimgraham:
> I found an odd edge case while testing locally.
> 
> It seems to only occur when there is already a tag in the same line. At that point the using the `#` will trigger autocompletion even after a space

Will fix.

jimgraham:
> I found an odd edge case while testing locally.
> It seems to only occur when there is already a tag in the same line. At that point the using the `#` will trigger autocompletion even after a space

Fixed in [83b7971](https://github.com/foambubble/foam/pull/1192/commits/83b7971c84a8ace778a9c79cb8173990150564c1)

riccardoferretti:
Released 0.22.1 to include this fix right away

Issue #1145, opened 2023-01-17
Clicking on placeholder from list would bring to location in note


wrgrant:
### Is your feature request related to a problem? Please describe.

Currently when you click on a placeholder entry in the sidebar, it creates a new note in the root of the vault.

To me this is a strange UX.

I would rather be dumped into the note where the placeholder is called out so I can get a context for why that placeholder exists and if there is any extra context or links that make sense to add there.

With the current behavior, I'm faced with a blank file that I don't know what to do with. That provides no context.

If I want to get to the placeholder, I need to do a VSCode content search to find where the placeholder was defined.
This is rather clunky.

### Describe the solution you'd like

Expected behavior would be to click on a placeholder entry in the sidebar and land at the point in the existing note where the placeholder was defined. Using the exact same behavior as when you click on a tag for a file from the Tags Explorer pane.

### Describe alternatives you've considered

Searching via VSCode content search.

badsketch:
![Screenshot 2023-01-17 231244](https://user-images.githubusercontent.com/8953212/213082728-0d48c6ec-1243-4a29-acf4-a5392e650abe.png)
I like how in certain file lists in vscode such as the source control tab, the context actions include jumping to the source, so what you're saying makes sense to me.

One thing we'd need to think about is placeholders that are referenced in multiple places. Currently, two separate files with the same placeholder name produces one placeholder, so I'm wondering the best way to display all sources.

riccardoferretti:
Currently:
- the placeholder panel doesn't show where placeholders are coming from (whereas the tags panel does it for each tag)
- clicking on the placeholder item triggers the flow of placeholder creation (which currently creates the note in the root)

We are about to release a setting to create a new note either at the root or in the current folder, this will affect the flow for placeholders.

Idea:
- I agree it would be helpful to expand on a placeholder to see all the places where it's referenced (like the tags panel)
- clicking on one of those notes would navigate to the note (same behavior as in the tags panel)
- this means that clicking on it will no longer create the placeholder
  - which is not necessarily a problem, so the above could simply be a solution. If you want to create the note, you can then do it from the wikilink in the editor

Otherwise to still enable the creation of the placeholder note from the panel, we have a few options:
1. do something similar to what we did in the tags explorer, where the first item of the children is "Create note for placeholder". Clicking on this would follow the standard flow of creation, as if triggered by cmd+clicking on the link in the first child note for the placeholder
2. add a `+` context action to the placeholder, to achieve the same as above
[...]

PR #1196, opened 2023-04-15
Resource tree items improvements

riccardoferretti:
A lot of TreeViews in Foam have a similar structure, they refer to resource or ranges within a resource.

In this PR the implementation of these differente views have been consolidated, to offer a consistent UX:
- resource items
  - use the title of the resource
  - hover will preview the resource
  - clicking will navigate to the resource
  - can be expanded when it has ranges within it
  - shows the number of ranges within the resource
- a range item (e.g. a backlink within a resource, a placeholder, ...) 
  - in the tree provides some context
  - on hover will show the full line
  - clicking on it takes the user to the range

I have also ported the Placeholder view to this new pattern, also inspired by the comments from #1145.


wrgrant:
Nice update!

PR #1226, opened 2023-05-07
Placeholder and Orphan panels using the Folder hierarchy

riccardoferretti:
Minor refactoring to consolidate the tree view class hierarchy


riccardoferretti:
Scope creep: I also updated the backlinks panel to follow the same pattern as the other tree views

PR #1230, opened 2023-05-18
Connections panel

riccardoferretti:
This PR replaces the Backlinks panel with a more complete Connections panel.

The connection panel shows both backlinks and forward links.
It allows the possibility to filter the list to only show backlinks or forward links.

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -56,34 +56,34 @@
     "views": {
       "explorer": [
         {
-          "id": "foam-vscode.backlinks",
-          "name": "Backlinks",
+          "id": "foam-vscode.connections",
+          "name": "Connections",
           "icon": "$(references)",
-          "contextualTitle": "Backlinks"
+          "contextualTitle": "Foam"
         },
         {
           "id": "foam-vscode.tags-explorer",
           "name": "Tag Explorer",
           "icon": "$(tag)",
-          "contextualTitle": "Tags Explorer"
+          "contextualTitle": "Foam"
         },
         {
           "id": "foam-vscode.notes-explorer",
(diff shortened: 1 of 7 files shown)


redactedscribe:
Nice addition; I've appreciated all Foam updates so far. The only immediate issue I had was confusion over was what the difference between _Links_ and _Connections_ was since they can be synonyms for each other. After reading this PR, I now realise that _Links_ is short for _Forward Links_ in this case and not the same as a general connection (which is a link in either direction). Perhaps consider being explicit and renaming _Links_ to _Forward Links_? I don't see much benefit in brevity if it can be misconstrued, however, it's probable that the confusion I experienced wouldn't have lasted once actually using the feature (I just happened to check it out immediately after updating the extension). Thanks.

riccardoferretti:
Thanks for the feedback, I am gonna fix it up today

martigr:
It is very useful to see both the forward and backward links.

But would it be possible to specify `excludes`? Now I have the problem that a note is no longer listed under orphans if this note contains images. The images are now listed as forward links in the connections view.

riccardoferretti:
The changes in the connections panel didn't impact on the definition of orphan note. 
My guess is that this change happened when we added support for attachments. 

I can see the case for not considering attachments as connections in the context of orphan notes, feel free to open a new issue for that.

manvibharat:
Hey I am just adding a point here, the connections panel just shows the connections that are there but not whole content. I can view the content on hover on each connection. Is there a possibility of showing whole content of the connections. That is, allowing the hovered content to be seen in place in the connections panel.  One more possibility (feature) is to put all the connections to the page at the bottom of the same page itself.  Thanks for the work you have done.

Issue #1129, opened 2023-01-02
Configure Foam's default file extension(s)


riccardoferretti:
### Is your feature request related to a problem? Please describe.

Currently Foam supports only `.md` files as notes, and will always create `.md` files when clicking on a link (well, unless differently instructed by the template being used)

### Describe the solution you'd like

As a user, I expect to be able to configure the list of extensions to be considered notes, e.g. `txt`, `mdx`, `qmd`, ...

There are really two issues here:

1. support for multiple file extensions to be included as notes - this would need to be a setting, similar to what we have for `foam.files.attachmentExtensions`

2. support for the creation of non-`.md` files. Here I can see two approaches: 
  a. use note templates - this requires no further changes and is the current workaround
  b. use the first extension from the setting above as the default extension for new notes

### Describe alternatives you've considered

No workarounds are available at the moment for including non-`.md` files in the graph.

To create non-`.md` files the current workaround would be to use the `filepath` property in templates.

riccardoferretti:
This is now fixed as of 0.25.0

PR #1235, opened 2023-05-22
Support for multiple file extensions and default extension

### LINKED ISSUE #924: Support for Quarto
We've started to use [Quarto](https://quarto.org) to create dynamic documents – weaving together narrative and code in plain text. With plans for a [Quarto vscode extension](https://github.com/quarto-dev/quarto-cli/issues/149) it would be amazing to have Foam support for `.qmd` files in preparation for potentially tighter integration.

Interested to hear thoughts on this.

riccardoferretti:
Fixes #924 and #1129

Introducing the `foam.files.noteExtensions` property, which tells Foam what extensions to treat as (markdown) text notes.

The first extension from the list will be treated as the default extension.

The default extension is the one assumed when a `[[wikilink]]` doesn't include an extension, and is also the one used when creating notes from a `[[placeholder]]`

Still debating whether the default extension should be an explicit property or the first one from the list of extensions. Leaving for the first in terms of clarity, but I like the second as it enforces consistency (i.e. you can't have a default extension that is not part of the allowed extensions) and avoids extra configuration

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -395,15 +395,15 @@
           "default": "pdf mp3 webm wav m4a mp4 avi mov rtf txt doc docx pages xls xlsx numbers ppt pptm pptx",
           "description": "Space separated list of file extensions that will be considered attachments"
         },
-        "foam.files.noteExtensions": {
+        "foam.files.notesExtensions": {
           "type": "string",
-          "default": "md",
-          "description": "Space separated list of file extensions that will be considered markdown notes"
+          "default": "",
+          "description": "Space separated list of file extensions that will be considered markdown notes (e.g. 'mdx md markdown')"
         },
         "foam.files.defaultNoteExtension": {
           "type": "string",
           "default": "md",
-          "description": "The default note extension"
+          "description": "The default extension for new notes"
         },
         "foam.files.newNotePath": {
           "type": "string",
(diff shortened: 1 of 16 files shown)


redactedscribe:
> Still debating whether the default extension should be an explicit property or the first one from the list of extensions.

Unless you read some docs, it may not be apparent that the value for the new setting acts in this way. A separate setting for the default makes it self-explanatory, but so do unambiguous tooltips. I generally prefer a label or variable to express its intent or function clearly as to not need any further explanation.

> Leaving for the first in terms of clarity, but I like the second as it enforces consistency (i.e. you can't have a default extension that is not part of the allowed extensions) and avoids extra configuration

If I've understood your sentence correctly, I'd say avoiding extra configuration isn't an weighty argument because the majority of users will likely not change either of these settings (especially if given sane defaults), but for those who do, I wouldn't consider it cumbersome to configure two separately.

If the hypothetical "`noteDefaultExtension`" were to be leveraged in the future for some additional purpose, a discrete settings could be more user-friendly, if not, then the first extension defined in `noteExtensions` is sufficient (given a good tooltip).

The problem I see if two settings are used is that reading `noteExtensions` doesn't imply that there exists a second setting, or main setting, "`noteDefaultExtension`" to define. Maybe "`noteOtherExtensions`"?

riccardoferretti:
Thanks for the feedback @redactedscribe, some interesting points. I will take it into account

redactedscribe:
Not a problem @riccardoferretti. Always happy to add a second opinion. My criticisms mainly come from a usability standpoint since I am more of a designer than a coder, yet I hope they do hold relevance. Thanks for the continued development.

Issue #1252, opened 2023-07-02
Connections/backlinks panel is empty when using a custom editor like Markdown Editor


64bits:
### Describe the bug

This has been a bug for a while, and I had an [interim fix](https://github.com/foambubble/foam/commit/5213547d2eb968def95eedf517b88ae2576fb8dd) but my fix seems to have stopped working in the latest version.

Essentially, I expect 'All Links' to work regardless of what type of editor I am using, and should not be restricted to plain text.

### Steps to Reproduce the Bug or Issue

1. Install https://marketplace.visualstudio.com/items?itemName=zaaack.markdown-editor
2. Open a markdown file that is supposed to have backlinks (linked to from other files) in this editor
3. Note that backlinks do not show under 'All Links', but will if you switch to plain text editing

### Expected behavior

The backlinks should show up, the issue seems to be related to triggering the refresh appropriately.

### Screenshots or Videos

https://github.com/foambubble/foam/assets/12042989/a5642d9b-0f52-45d6-b5d6-52ac2cab750d

### Operating System Version

macOS

### Visual Studio Code Version

1.78.0

riccardoferretti:
Addressed by ef5439c4bd23632b2e3f4dd2fb7c69f95dfaa834

PR #1253, opened 2023-07-04
Fixed an issue with third party editors and connection panel

64bits:
This fixes https://github.com/foambubble/foam/issues/1252.

Screenshot 2023-07-04 at 12 55 53 PM


riccardoferretti (commented):
Thanks for looking into this.
I checked the code, and I feel it might break under various cases, from when there is more than a root folder, to having notes in subfolders.
Maybe we could check whether we can get the URI from the active tab, can you take a look?

PR #1267, opened 2023-08-04
Update note-templates.md, current example doesn't work

tcheneau:
Fix typo (backticks instead of simple quotes)


riccardoferretti (approved):
thanks for fixing this!

riccardoferretti:
@allcontributors add @tcheneau for docs

PR #1271, opened 2023-08-12
Update Jest to v29

nicholas-l:
This updates to the latest Jest version.

Note: I haven't been able run tests (either master or this branch) locally as I believe it is due to Arm issues.


riccardoferretti (commented):
Thanks for this. 

I tested it locally and the only issue I have found is that if I run the tests from the root via `yarn test`, the command ends (successfully) before the tests are fully run. This doesn't happen with Jest 27.
This doesn't happen in CI as the `--stream` flag is being passed, and I tested locally that by passing the flag is issue is resolved too. 
I have also verified that instead when running `yarn test` from within the `foam-vscode` package the behavior is correct (that is, the tests all run before the command returns)

Adding the `--stream` flag to the root package.json `test` script seems to fix the problem:

```
    "test": "yarn workspace foam-vscode test --stream",
```

nicholas-l:
What does that `--stream` flag do? I had assumed that it would be used in `packages/foam-vscode/src/test/run-tests.ts` but it doesn't seem to.

riccardoferretti:
The issue seems to be in the way `yarn` delegates the command to the package.
I assume as much given that the issue only appears when running the command from the top level package, and not from within the package itself (the only difference between the two is the yarn delegation)

I can't find anything specific on the yarn documentation for the `--stream` flag though.

Keep in mind that we have a very old version of `yarn` (still v1), which should be updated to v3 (or removed altogether).

For the sake of this PR I am happy by simply adding `--stream` to the top-level script as mentioned in my previous comment

riccardoferretti (approved):

riccardoferretti:
@allcontributors add @nicholas-l for code

Issue #879, opened 2021-12-16
Support different embedding styles


riccardoferretti:
Currently embedding includes the note/section content inlined.

It would interesting to support different "embed styles" to customize the embedding behavior with: 
- inline (include the full section, title included, in the note)
- content (include the full section, excluding the title, in the note)
- card/panel (add the embed inside a card or panel)

This could either be set by a workspace setting, or with some flag on the embed itself.

With a setting this would be pretty straightforward, with a flag on the embed we'd have to figure out a syntax that is falls back somewhat gracefully in other tools (e.g. Obsidian), but that adds the semantic we care about it Foam. (e.g. `card![[my-note#section]]`)

The interesting thing about this approach is that it would also allow more styling to be applied in the future.

riccardoferretti:
Hi @SteveDadoly thanks for your thoughts above.

I like the idea of describing both content and layout in the "modifier ", I think it's a good way to conceptualize the mapping.
To keep things simple I would avoid to have "special syntax", but instead make the list "flat". The simple way to achieve it is by replacing `|` with `-` (which IMO feels more natural and less "scripty"). See below for what I mean.

So, assuming:
- the content modifiers `full` (I think it works better than `all`) and `content` 
  - I am excluding for now the "first block" and "exclude sub sections" options; the design supports them (which is what matters) but I am prefer to start small
- and the style modifiers `inline` and `card`

we'd end up with:

`full-inline![[wikilink#section]]`
`full-card![[wikilink#section]]`
`content-inline![[wikilink#section]]`
`content-card![[wikilink#section]]`

For each of those dimensions (content and style) we can have a default (which is what comes from the Foam setting).
E.g. with `full` and `inline` as defaults, the above list would turn into:

`![[wikilink#section]]`
`card![[wikilink#section]]`
`content![[wikilink#section]]`
`content-card![[wikilink#section]]`

Which seems pretty reasonable to me.

A couple of extra questions:
- which settings is more appropriate as default between `card` and `inline`, thoughts?
- would it make sense to have `content-card![[wikilink#section]] == card-content![[wikilink#section]]`.. does it really matter if the content modifier is before or after the style modifier?

SteveDadoly:
Hi @riccardoferretti,

Thank you for your thoughtful reply. Agree with all of your points above.  Best to keep it simple.   

From my perspective:

- Would default to the style that matches the target document (inline)
- The order of modifiers does not matter to me.

badsketch:
[...]

PR #1273, opened 2023-08-18
Feat #879: Different embedding styles prep

badsketch:
Addresses #879
This PR does not change functionality. It's mainly implementation changes under the hood in preparation for when we do add other embedding styles.

raw

preview

I tried to keep in mind that we want to make this extensible for future variations of style and content scoping, so I tried to use a strategy design pattern where we have extractors and formatters. Hoping it's not overkill and not too hard to read. 😅OOP isn't my strong suit, but I had fun! Hoping for feedback.

Is this what you had in mind? Could you also clarify what you think would be the correct rolloutstrategy? I noticed if I add a `deprecationMessage` to `preview.embedNoteInContainer` then it disappears from the settings UI unless you explicitly have it set. At the moment, the new  `preview.embedStyle` doesn't do anything. Wondering at what point should I hot swap the two. 
If we're going to maintain both at the same time, (before `preview.embedNoteInContainer` gets removed), which setting should take precedence?


packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -123,4 +118,102 @@ function withLinksRelativeToWorkspaceRoot(
...
+    }
+    noteText = withLinksRelativeToWorkspaceRoot(
+      noteText,
+      this.parser,
+      this.workspace
+    );
+    return noteText;
+  }
+}
+
```
badsketch, line 151: A `class ContentExtractor implements EmbedNoteExtractor...` would go here in the future

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -47,25 +47,20 @@ export const markdownItWikilinkEmbed = (
         let content = `Embed for [[${wikilink}]]`;
         switch (includedNote.type) {
           case 'note': {
-            let noteText = readFileSync(includedNote.uri.toFsPath()).toString();
-            const section = Resource.findSection(
+            const noteStyle = getFoamVsCodeConfig(
```
badsketch, line 50: In the next phase, I'd update the regex to capture the new syntax. Then here we'd have logic to determine `noteStyle` and `noteContent`. For now, `noteStyle` pulls from the existing config setting, and there is no `noteContent`, it just defaults to "full".

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -123,4 +118,102 @@ function withLinksRelativeToWorkspaceRoot(
...
+
+interface EmbedNoteFormatter {
+  format(content: string): string;
+}
+
+class CardFormatter implements EmbedNoteFormatter {
+  md: markdownit;
+  constructor(md: markdownit) {
+    this.md = md;
+  }
```
riccardoferretti, line 160:
```suggestion
  constructor(private md: markdownit) {  }
```

packages/foam-vscode/src/features/preview/wikilink-embed.ts
riccardoferretti: That's correct, although from the configuration we should be getting both the style and the content, is that planned for this PR?

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -47,25 +47,20 @@ export const markdownItWikilinkEmbed = (
...
         switch (includedNote.type) {
           case 'note': {
-            let noteText = readFileSync(includedNote.uri.toFsPath()).toString();
-            const section = Resource.findSection(
+            const noteStyle = getFoamVsCodeConfig(
+              CONFIG_EMBED_NOTE_IN_CONTAINER
+            )
+              ? 'card'
+              : 'inline';
+            const noteEmbedder = new NoteEmbedder(
```
riccardoferretti, line 55: I wonder if `NoteEmbedder` helps encapsulate a behavior, and compare that with the additional "complexity" of the indirection - what do you think?

riccardoferretti (commented):
Thanks for taking a stab at this!
For what regards the design, in Foam we tend to be more on the functional side than OOP (with blurred lines, especially because VS Code API is mostly OO).
I like the separation of the extraction from the rendering.
For me an interface with a single method is a strong candidate for being replaced by a function type (e.g. see `FoamFeature` interface, it [used to be an object](https://github.com/foambubble/foam/pull/1224), in the end I decided that a simple fn was enough for all the use I needed for it, and made the code more readable).

I left some comments, let me know your thoughts

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: So I'm a little fuzzy on release strategy. The new `preview.embedStyle` isn't used anywhere at the moment, but I can change it so that this wikilink-embed reads it and it overrides `preview.embedNoteInContainer` if both are set. I can also add a deprecation flag to  `preview.embedNoteInContainer` so it no longer appears in settings UI. In the next PR, I can remove it altogether `package.json` as well as this file. wdyt?

badsketch:
> For me an interface with a single method is a strong candidate for being replaced by a function type (e.g. see `FoamFeature` interface, it [used to be an object](https://github.com/foambubble/foam/pull/1224), in the end I decided that a simple fn was enough for all the use I needed for it, and made the code more readable).

Great idea! I think it'll make it a lot more succinct

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -123,4 +118,102 @@ function withLinksRelativeToWorkspaceRoot(
...
+    this.md = md;
+  }
+  format(content: string) {
+    return `<div class="embed-container-note">${this.md.render(content)}</div>`;
+  }
+}
+
+class InlineFormatter implements EmbedNoteFormatter {
+  format(content: string) {
+    return content;
```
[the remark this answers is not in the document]
badsketch: I think the switch statement for note, attachment, and image all return content and outside the switch statement `content` gets rendered here https://github.com/foambubble/foam/pull/1273/files#diff-42b32f07212d469a3715b0a74b0f67ff192e592d0724fcfce358130debe8dcfaL84

It does feels weird for InlineFormatter to be doing nothing, but the Extractor kind of already does everything and I couldn't find something in 
```javascript
  extract(note: Resource) {
    let noteText = readFileSync(note.uri.toFsPath()).toString();
    const section = Resource.findSection(note, note.uri.fragment);
    if (isSome(section)) {
      const rows = noteText.split('\n');
      noteText = rows
        .slice(section.range.start.line, section.range.end.line)
        .join('\n');
    }
    noteText = withLinksRelativeToWorkspaceRoot(
      noteText,
      this.parser,
      this.workspace
    );
    return noteText;   // already "formatted"
  }
}
```
that could belong in formatter.

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: A little confused by "encapsulate a behavior". Is that a type of implementation pattern?

badsketch:
> > For me an interface with a single method is a strong candidate for being replaced by a function type (e.g. see `FoamFeature` interface, it [used to be an object](https://github.com/foambubble/foam/pull/1224), in the end I decided that a simple fn was enough for all the use I needed for it, and made the code more readable).
> 
> Great idea! I think it'll make it a lot more succinct

Hmm, gave this shot, but I realized that with a function type I don't have a way to enforce classes to implement that function. And NoteEmbedder expects the interfaces to be defined so that it can use polymorphism to `extract()` and `format()`. 
```
class NoteEmbedder {
  extractor: EmbedNoteExtractor; // can't be a function type
  formatter: EmbedNoteFormatter; // can't be a function type
.
.
 generateEmbedding () {
  content = extractor.extract(stuff)
  return formatter.format(content)
}
```

Could I be missing something?

packages/foam-vscode/src/features/preview/wikilink-embed.ts
riccardoferretti: The release strategy you outlined is what I had in mind and I think works pretty well.

I am not aware of a built-in way in VS Code to do that, did you come across something or had something in mind?

packages/foam-vscode/src/features/preview/wikilink-embed.ts
riccardoferretti: no no, what I mean is basically that currently in `NoteEmbedder` lives the code that was in this function (that's what I mean by "encapsulating": we don't see the actual code and rely on this object to "do the work").
This indirection removes complexity by removing some code here (and putting it in `NoteEmbedder`) but also adds complexity because understanding the code requires an extra step (that is, check the implementation of `NoteEmbedder`).
My original comment was whether what's gained offsets what's lost, I wonder if we could in fact remove the `NoteEmbedder` and simply have the code here, wdyt?

packages/foam-vscode/src/features/preview/wikilink-embed.ts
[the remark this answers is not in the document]
riccardoferretti: `CardFormatter` renders the md, so I feel the behavior should be consistent: do we expect the formatter to return an html string or a md string?

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: Do you mean for the deprecation flag? https://code.visualstudio.com/api/references/contribution-points#contributes.configuration:~:text=deprecationMessage%20/%20markdownDeprecationMessage

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: Gotcha, I see what you mean. That's a good point. For https://github.com/foambubble/foam/pull/1273/commits/e05c1c282e59e995e8c707f8787de8157e73eb05 I changed the interfaces to function types and replaced all the classes with functions. Is this what you had in mind? 

I feel like the class-based approach would be preferable after we add more variations, so maybe we can go with the functional approach and reconsider it as an option in the future? I say that because with the functional approach, it seems dependencies get added for all formatters and extractors. Like if extractor A requires `parser: ResourceParser`, but extractor B doesn't,  the type still needs to include it, and extractor B has to have it in its signature:
```
export type EmbedNoteExtractor = (
  note: Resource,
  parser: ResourceParser,
) => string;

function extractorA(note: Resource, parser: resourceParser) {}
function extractorB(note: Resource, parser: resourceParser) {} // doesn't need the parser, but generateNoteEmbedding() expects a type EmbedNoteExtractor 
```

packages/foam-vscode/src/features/preview/wikilink-embed.ts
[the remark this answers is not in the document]
badsketch: So the issue is that card style needs an additional`md.render(content)` step. 
Existing code returns an md string: 
```js
  content = getFoamVsCodeConfig(CONFIG_EMBED_NOTE_IN_CONTAINER)
    ? `<div class="embed-container-note">${md.render(noteText)}</div>`
    : noteText;
```
So we can have both return an html string: 
```js
function cardFormatter(content: string, md: markdownit): string {
  return md.render(
    `<div class="embed-container-note">${md.render(content)}</div>`
  );
}

function inlineFormatter(content: string, md: markdownit): string {
  return md.render(content);
}
```
or we can have both return an md string
```js
function cardFormatter(content: string, md: markdownit): string {
    return `<div class="embed-container-note">${md.render(content)}</div>`
}

function inlineFormatter(content: string, md: markdownit): string {
  return content;
}
```

packages/foam-vscode/src/features/preview/wikilink-embed.ts
riccardoferretti: Yes, that's great, I was already thinking of a small utility to deal with deprecation, but that won't be necessary 👍

packages/foam-vscode/src/features/preview/wikilink-embed.ts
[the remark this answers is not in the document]
riccardoferretti: Oh I see what you mean, I had forgotten about the double rendering.
I don't have a strong preference between the two options, so long as we are consistent, happy to go with what you prefer.

packages/foam-vscode/src/features/preview/wikilink-embed.ts
riccardoferretti: How about moving the two lines of `generateNoteEmbedding` in the caller and remove the fn altogether?

Your comment about the dependencies is correct. And in some of my functions this situation happens. For example in the feature activate function I used as example before, sometimes the `Foam` object is not required, but I pass it anyways.
Depending on the number of dependencies and how related they are to the method, I am ok with the compromise.

In this specific case of the extractor and the formatter, so far the parameters being passed IMO read well with the function signature, so I don't see yellow/red flags. 
This is something that can be reassesed in the future shall the need arise. 

---

As an aside, the functional way of mimicking the constructor would be using a factory, e.g.

```
function createExtractor(extraDep: any) : (note: Resource, parser: resourceParser) => string {
  return (note: Resource, parser: resourceParser) => {
    // do something with all extraDep, note and parser
  }
}

extractorC = createExtractor(extra)

extractorC(note, parser)
```

riccardoferretti:
I think I replied to all your questions, let me know if the part around the embedder is still unclear though!

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: sweet!

packages/foam-vscode/src/features/preview/wikilink-embed.ts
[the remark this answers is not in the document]
badsketch: cool, I decided to go with producing an html and having the rest of the cases like embedding images or attachments also produce the html as well

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: Great suggestion, done! Also really appreciate the snippet

packages/foam-vscode/package.json
```diff
@@ -514,7 +514,20 @@
...
-          "description": "Wrap embedded notes in a container when displayed in preview panel"
+          "description": "Wrap embedded notes in a container when displayed in preview panel",
+          "deprecationMessage": "*DEPRECATED* use foam.preview.embedNoteType instead."
+        },
+        "foam.preview.embedNoteType": {
+          "type": "string",
+          "default": "full-card",
+          "enum": [
+            "full-inline",
+            "full-card"
```
badsketch, line 525: decided to save `content-inline` and `content-card` enum options for when we actually add their implementation since it would be confusing if the options didn't do anything.

badsketch:
Appreciate your patience in working through this with me! I think the code/design is in a good spot based on what we discussed 👍 

Lastly:
I decided to change `preview.embedStyle` to `preview.embedNoteType` since in the code I use "style" to refer to the format type (`inline`, `card`), and the naming was starting to get confusing to me if the setting has the word "style" in it. I thought it might be a clearer separation of identity for our variables if scope = full/content, style = inline/card, and `embedNoteType` as the combination of the two. Thoughts?

Is it possible for you to retrigger the `CI /Build and Test`? It's failing on `wikilink-embed.spec.ts`, but it succeeds on my machine. Also, I'm noticing `refactor.spec.ts` fail on my machine, but succeed in CI. Have you noticed this before? Or might I have messed something up 😓

packages/foam-vscode/package.json
riccardoferretti: Yes, if we are not adding the implementation we shouldn't have the options 👍

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -45,27 +46,31 @@ export const markdownItWikilinkEmbed = (
...
-            const section = Resource.findSection(
-              includedNote,
-              includedNote.uri.fragment
-            );
-            if (isSome(section)) {
-              const rows = noteText.split('\n');
-              noteText = rows
-                .slice(section.range.start.line, section.range.end.line)
-                .join('\n');
+            let { noteScope, noteStyle } = retrieveNoteConfig();
```
riccardoferretti, line 53: why `let`?

riccardoferretti (commented):
Yes it looks really good!
I left a couple of minor comments, but we are there.

riccardoferretti:
I re-triggered the tests, it looks like something is getting escaped in the html, so I reckon it will fail again.

(re the refactor tests, yes sometimes they fail for me too locally, haven't figured out how to remove those spurious fails)

packages/foam-vscode/src/features/preview/wikilink-embed.ts
badsketch: woops, thanks

badsketch:
You might be able to tell from the commit history I've somewhat lost it 😓. I've just been getting the most bewildering results in terms of reproducibility, which makes it really frustrating to debug, and it requires I push to really see if my changes are effective. I'm starting to suspect it's the async nature of `wikilink-embed.spec.ts`, but knowing my luck it could just be another red herring. Might take a bit longer as I try to fix this. If you do notice anything glaringly wrong, though, please let me know!

riccardoferretti:
I took a look at the tests and it seems like they are using the `card` formatter instead of the `inline`.
And I wonder if the problem is in `retrieveNoteConfig`, probably we shouldn't give priority to the old setting `CONFIG_EMBED_NOTE_IN_CONTAINER`.

(2m later - I did a quick test locally by commenting the overwrite and the tests passed)

badsketch:
Yeah! Based on what you said, I think I got it figured out. First mistake was setting the wrong config name https://github.com/foambubble/foam/pull/1273/commits/4190c438a1378b89e059f393ab9a27a9483b0845, and secondly, I needed to initialize `CONFIG_EMBED_NOTE_IN_CONTAINER`, by wrapping the tests in an additional `withModifiedFoamConfiguration()` otherwise all had the setting set to true, and the flag has higher precedence in `retrieveNoteConfig()`

Should finally be in a good review state. Again, thanks for the troubleshooting help! Looking forward to finally doing the implementation 😅

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -123,4 +121,60 @@ function withLinksRelativeToWorkspaceRoot(
...
+export function retrieveNoteConfig(): {
+  noteScope: string;
+  noteStyle: string;
+} {
+  let config = getFoamVsCodeConfig<string>(CONFIG_EMBED_NOTE_TYPE); // ex. full-inline
+  let [noteScope, noteStyle] = config.split('-');
+
+  // **DEPRECATED** setting to be removed
+  // for now it overrides the above to preserve user settings if they have it set
+  if (getFoamVsCodeConfig<boolean>(CONFIG_EMBED_NOTE_IN_CONTAINER, false)) {
```
riccardoferretti, line 133: From this I assume that if someone has both settings, the old setting will overwrite the new one?
I think that the new setting should always have the precedence, and we should only read the previous one if the new setting has not been used, wdyt?

riccardoferretti (approved):
On second thought, I don't think that comment is very important, as we are planning to sunset the old config pretty quickly, so all looks good to me!

PR #1281, opened 2023-09-01
Feat #879: Note Embedding Syntax

badsketch:
Addresses #879 
- deprecates `foam.preview.noteEmbedInContainer`
- adds syntax discussed. I added comments in the code, but basically the regex plugin for markdown-it only supports one group, which is currently just the wikilink. Therefore I need to capture the entire wikilink item (all of `content-card![[note-a]]`), and then have another regex that does the parsing and grouping to extract the modifier and the wikilink. This leads to a bit of messy regexes, but I added tests to try to confirm their validity. 
- is not order agnostic, so `full-card![[note]]` would match both modifiers, whereas `card-full![[note]]` wouldn't recognize the `card` portion and just parse the `full` part. 🫤 I could make it recognize both, but it would require expanding the existing regex, which could get ugly. Then again, the existing behavior might confuse users. 

Let me know if this is what you had in mind!

## Testing

Given we have `content-inline` as the user setting

given

Case 1: no type modifiers should just default to the user setting
1-b
1-res

Case 2: adding a type modifier `content` should not have an effect since it's already the default

4-b
4-res

Case 3: adding one modifier `card` should override `inline` resulting in `card-inline` style
2-b
2-res

Case 4: adding two modifiers `full-card` should override both default style and scope resulting in `full-card` style
3-b
3-res


riccardoferretti (commented):
This looks good to me! Only thing I would ask is to add documentation for the feature in the docs folder

badsketch:
Updated docs, so lemme know your thoughts on wording, formatting, etc.!

riccardoferretti (approved):
This looks amazing!

PR #1286, opened 2023-09-15
Fix embedded notes not generating proper reference links

badsketch:
I noticed when generating reference links for a note that includes embedded notes, the reference links have a syntax error.
before

This is caused by an off-by-one error when extracting the label after gathering the links for a note since embedded notes have an extra `!`. 

After the fix:
after

Appreciate any feedback!

### DIFF
--- a/packages/foam-vscode/src/core/services/markdown-provider.ts
+++ b/packages/foam-vscode/src/core/services/markdown-provider.ts
@@ -137,12 +137,22 @@ export function createMarkdownReferences(
         relativeUri = relativeUri.changeExtension('*', '');
       }
 
+      let label = link.rawText;
+      if (link.rawText.indexOf('[[') > -1) {
+        if (link.isEmbed) {
+          // ![[embedded-note]]
+          label = link.rawText.substring(3, link.rawText.length - 2);
+        } else {
+          // [[standard-note]]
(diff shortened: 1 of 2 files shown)


packages/foam-vscode/src/core/services/markdown-provider.ts
```diff
@@ -137,12 +137,22 @@ export function createMarkdownReferences(
...
+      if (link.rawText.indexOf('[[') > -1) {
+        if (link.isEmbed) {
+          // ![[embedded-note]]
+          label = link.rawText.substring(3, link.rawText.length - 2);
+        } else {
+          // [[standard-note]]
+          label = link.rawText.substring(2, link.rawText.length - 2);
+        }
+      } else {
+        label = link.rawText;
```
riccardoferretti, line 150: I believe this can be removed
```suggestion
```

riccardoferretti (approved):
I have left a very minor comment, this looks good otherwise.

I don't consider it a blocking comment, so in any case will merge tomorrow

packages/foam-vscode/src/core/services/markdown-provider.ts
badsketch: nice! updated, lemme know if it's too "busy"

riccardoferretti (approved):
Looks good!

PR #1290, opened 2023-09-23
[WIP] Enable web extension for Foam

pderaaij:
I am making a fresh attempt on getting Foam to work on the browser. I'll close #1127 later on, but am keeping it for reference. So far, so good. But long way to go.


pderaaij:
Update:

The extension is running quite well in the online environment at first sight. To be sure - and for regression, I'd like to run the integration tests for web as well. This is where the pain begins. 

Last time I got stuck in a long loop to get Mocha working, so this time I went for `vitest`. However, this is not an easy ride as well. Vitest works well for unit tests, but integration tests are the problem. For reference, we need to run the test suite programmatically. We need to start the vscode environment and then run the test suite (see: `run-tests/ts`). Support for programmatically is limited in vitest (https://vitest.dev/advanced/api.html). However, this only works in `ESM` mode. Vscode,  due to electron, runs only `commonjs` (https://github.com/microsoft/vscode/issues/130367).

There seems to be some ways to get commonjs working in vscode, but I am wondering if we want all these changes just to get testing for web working.

For now, I feel `vitest` is not the way forward. However, I did learn a thing or two. I am going to have a fresh look at getting `mocha` working. But, fresh ideas are welcome.

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -194,7 +193,7 @@ function contentExtractor(
   parser: ResourceParser,
   workspace: FoamWorkspace
 ): string {
-  let noteText = readFileSync(note.uri.toFsPath()).toString();
+  let noteText = vsWorkspace.fs.readFile(toVsCodeUri(note.uri)).toString();
```
riccardoferretti, line 196: I think `vsWorkspace.fs.readFile` returns a promise, this was to me the only open question as to how to transition in the code

riccardoferretti (commented):
Thanks for giving it a second stab! I am happy to see that the code is basically ready for the transition, except for the `readFileSync` which I still think needs a solution (have you checked if the `toString` works?)

Unfortunately I am not super versed in mocha nor vitest, so can't help much on that front. The commonjs requirement is a bit of a limitation, and I agree that is probably better to play within that unless a solid solution can be found otherwise.

pderaaij:
I've been working on getting the tests to work with Mocha. Although it seemed I was making progress I ran into a big problem. Our assertion library is jest's `expect`. A library that's not easily replaced by `chai` or a other browser compatible assertion library. So, my conclusion is that if we really want to have tests for both extension and browser we would need to overhaul the current testing library, or maintain two libraries. Both are not preferable.

My suggestion is to enable the brwoser extension and only add a test suite that checks if Foam can be activated in the web browser. A pattern that I have seen at more VSCode extensions. This way we at least confirm the webpack bundle is okay and works. Specific checks for working functionality in the browser would not exist. Only perhaps if it is a recurring event in which we would write a specific regression test.

@riccardoferretti what do you think?

riccardoferretti:
Thanks @pderaaij for the investigation.

> My suggestion is to enable the brwoser extension and only add a test suite that checks if Foam can be activated in the web browser. A pattern that I have seen at more VSCode extensions [...] Specific checks for working functionality in the browser would not exist. Only perhaps if it is a recurring event in which we would write a specific regression test.

Mmmm.. can you share a couple of VS Code extensions that use that approach? 
I fear it might be a little brittle, but maybe I am missing something?
I guess testing the activation would validate that imports and packages are valid, but wouldn't cover the actual runtime.. then when an issue arises we could cover it with an explicit test (which requires a new assertion library, see comment below).
I am trying to figure out how big the window for uncaught bugs is with this approach.. 

> Our assertion library is jest's expect. A library that's not easily replaced by chai or a other browser compatible assertion library. So, my conclusion is that if we really want to have tests for both extension and browser we would need to overhaul the current testing library, or maintain two libraries. Both are not preferable.

I am not against migrating over time to another library that would be more web friendly. Basically all new and edited tests would use the new library until we reach a point of switch over.
I found some articles online of people that have done such thing, so at least we know it's possible.

pderaaij:
Annoyingly enough I can't find these examples anymore in my history and a quick GitHub search leaves me blank as well. I have researched so many repo's I might lost track how often I have seen this pattern. I do agree it feels quite brittle.

We could first rewire the testing framework. Or we just publish the web extension and rewrite the tests based on incoming reports. But that's not the greatest experience. In any case, it will require some work.

riccardoferretti:
Ok, this is something to think about, I am still debating if it's "good enough" (it might be).
Can you get the PR in a place where all tests pass using jest? That feels like the first thing to ensure regardless.

I am open to your suggestion about using the legacy framework (jest) for the regular testing, and only have some smoke tests for the web extension.

packages/foam-vscode/src/features/preview/wikilink-embed.ts
tegkhanna: Hi, is this still open? If yes can we get around by writing a new `markdown-it` plugin to handle the async operation?
Their documentation has some [guidelines](https://github.com/markdown-it/markdown-it/blob/master/docs/development.md#i-need-async-rule-how-to-do-it). Though I don't follow the wording aptly but I can give it a quick stab by forking `markdownItRegex`. WDYT?

riccardoferretti: This is still open, you can find all the info here: https://github.com/foambubble/foam/pull/1295

flancian:
Ahoy there! I come with few skills but willing to contribute help to try to merge this if I can provide any. I also know several other people who are interested in getting Foam to work on VS Code web :)

What's left to be done? Could the extension be merged as-is without interfering with code quality/stability for regular (non-browser-based) VS Code users, maybe with a warning/alpha disclaimer?

pderaaij:
@riccardoferretti this would be enough to make the web version work. However, note-embedding will not work correctly. I might be able to get it working via: https://github.com/microsoft/vscode/issues/174080#issuecomment-1885857624.

Additionally, I have not been able to get a simple end-2-end test working. My idea was to simply test if the extension was loaded. But I can't get `mocha` to play along. 

Nevertheless, we can publish the extension for web with these comments and start to collect the feedback.

pderaaij:
It took some time, but I got an end-2-end test working. It is a simple check that asserts if the foam extension is loaded.

image

@riccardoferretti I've added an extra CI step to build and verify the web version is working. If you see this differently, let me know.

My goal would be to get this PR merged. The next step is to attempt to change the markdown preview.

packages/foam-vscode/web/webpack.config.js
```diff
@@ -0,0 +1,77 @@
...
+    filename: '[name].js',
+    path: path.join(__dirname, './../out/web'),
+    libraryTarget: 'commonjs',
+    devtoolModuleFilenameTemplate: '../../../[resource-path]',
+  },
+  resolve: {
+    mainFields: ['browser', 'module', 'main'],
+    extensions: ['.ts', '.js'], // support ts-files and js-files
+    alias: {
+      // Overwrite wikilink embedding as it will not work in the current setup on web
```
riccardoferretti, line 24: 2 comments:
1. I wonder if we can handle it from within `wikilink-embed.ts` so that it's more explicit (I don't love webpack nor its black magic ;) )
2. if approach n.1 is not good, I would actually have `/src/web/features/preview/wikilink-embed.ts` -> `src/features/preview/wikilink-embed-for-web-extension.ts` (and add a comment on both files to explain what's happening)

packages/foam-vscode/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "lib": ["ES2019", "es2020.string", "DOM"],
     "sourceMap": true,
     "strict": false,
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "skipLibCheck": true
```
riccardoferretti, line 11: what's the goal of this setting?

packages/foam-vscode/src/web/features/preview/wikilink-embed.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { ResourceParser } from 'packages/foam-vscode/src/core/model/note';
+import { FoamWorkspace } from 'packages/foam-vscode/src/core/model/workspace';
+
+export const markdownItWikilinkEmbed = (
+  md: markdownit,
+  workspace: FoamWorkspace,
+  parser: ResourceParser
+) => {
+  return md;
```
riccardoferretti, line 11: I would actually return something along the lines of:
```
return `<div class="foam-no-embeds-warning">Embeds are not supported in web extension: ${wikilink}</div>`;
```

`foam-no-embeds-warning` could be along the lines of `foam-cyclic-link-warning`

This is gonna save us from a bunch of bug reports about the missing feature ;)

packages/foam-vscode/src/test/web/web-extension.webtest.ts
```diff
@@ -0,0 +1,11 @@
+/*global suite:readonly*/
+
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+suite('Foam Web Extension Test Suite', () => {
+  vscode.window.showInformationMessage('Start all tests.');
+
+  test('Foam extension is loaded', () => {
+    assert.ok(vscode.extensions.getExtension('foam.foam-vscode'));
```
riccardoferretti, line 9: I like the idea of doing just a quick smoke check here, making sure that:
- the extension is present
- the API is exposed
- a few commands are available

I think that should be enough to get started

packages/foam-vscode/src/test/web/index.ts
riccardoferretti, line 1: Feels like for consistency this file should be `packagaes/foam-vscode/src/test/run-tests-for-web-extension.ts` ?

packages/foam-vscode/src/services/editor.ts
```diff
@@ -1,6 +1,5 @@
 import { isEmpty } from 'lodash';
 import { asAbsoluteUri, URI } from '../core/model/uri';
-import { TextEncoder } from 'util';
```
riccardoferretti, line 3: doesn't remove these import affect the native extension?

package.json
```diff
@@ -17,6 +17,7 @@
     "clean": "lerna run clean",
     "build": "lerna run build",
     "test": "yarn workspace foam-vscode test --stream",
+    "test:web": "yarn workspace foam-vscode test:web",
```
riccardoferretti, line 20: I think the top level `test` should call both test targets on the package, so we have a consistent API at this level

riccardoferretti: Would that also mean that you don't need the extra step in the CI? I am not totally opposed to it (we could call the task on the package if anything)

riccardoferretti (commented):
Thanks @pderaaij I have left a few comments, let me know what you think

packages/foam-vscode/web/webpack.config.js
pderaaij: First won't be possible. As the file contains an import of `fs` it will try to pack that. By overwriting the file on compile time, I am preventing these issues.

packages/foam-vscode/tsconfig.json
pderaaij: Preventing `tsc` to check the types within the project. I've added this to prevent the collisions between `mocha` and `jest`. We use `mocha` for the e2e tests and `jest` for the vscode extension. They are incompatible with each other, but for our use case that's not a problem. However, `tsc` scans the entire `node_modules` folder and detects the collisions.

packages/foam-vscode/src/services/editor.ts
pderaaij: Not required anymore, as it is default part from bode > 16.

packages/foam-vscode/src/web/features/preview/wikilink-embed.ts
pderaaij: Screenshot 2024-07-11 at 20 18 57

packages/foam-vscode/src/test/web/web-extension.webtest.ts
pderaaij: I am stuck at expanding the tests. See https://github.com/microsoft/vscode-test-web/issues/130.

riccardoferretti:
Hi @pderaaij I was trying to test this locally, but I am getting the following error in the browser 

image

Sequence of steps:
- checkout the branch
- `yarn reset`
- `cd packages/foam-vscode`
- `yarn compile-web`
- `yarn package-web`
- `yarn open-in-browser`
- Open foam-template
- go to `getting-started.md` (so far it is working as the links showing up in the "All Links" panel are correct)
- hover with the mouse on a link (interesting that the navigation still work)

Also, the `Foam: Show Graph` command doesn't work (seems related to the lookup of the webview).

Can you confirm the issues?

pderaaij:
Yes, the Buffer one I found out as well. It is locally in my stage, but yet to commit. Will do so. 

It is a webpack config issue by the way. Easily resolved.

pderaaij:
Got the tests working:

image

Functionality should be working now. Let me know what you think @riccardoferretti

riccardoferretti:
First of all, amazing work @pderaaij! This is not a simple change!

There is one thing that I am really hesitating on, and that's mocha. I normally wouldn't like to have 2 "competing" libraries, in this specific case the fact that it's for testing and the extra magic on the TS front makes me even more wary.

So, I am looking at the code and am trying to understand where jest falls short.
Why can't `packages/foam-vscode/src/test/web/index.ts` use jest? I understand the limitations on the assertion library, and in that regard I would suggest we don't need an assertion library for two tests - if we can run them via jest.
Can you help me understand?

Thanks!

pderaaij:
As far as I understand the problem is that Jest is an unit testing library. It is designed to run within the nodejs environment, but not the browser. 

The test suite runs in the browser. The vscode-web-test extension starts a local VS code web environment in which Foam is loaded. After activation the test suite runs in the browser. For this we need a browser library. `` mocha`. Is the suggested library by the vscode docs. 

I know it isn't ideal, but it is the pattern I saw in many repositories.

riccardoferretti:
I see what you mean.
I was testing the extension and I am getting an error when showing the node graph, are you having the same issue?

pderaaij:
@riccardoferretti Quite sure it was working, but can't understand how it could be working. I worked on a concept to make the graph working again. Could you verify if this works for your? run `npm run compile-web` before running `npm run open-in-web`

riccardoferretti:
Thanks @pderaaij - I think things are really coming together and I was dealing with the odd sensation that there is something that rubs me the wrong way.

After much thinking I think I have come to the source of the issue: I feel too many bits are impacted, and between webpack and mocha, and the extra TS settings, I am very hesitant.
The work you have done is **amazing**, and I think I found a solution that I am comfortable with. 
I have created a PR (#1395) that leverages the great parts of this one, but without introducing webpack/mocha/etc with some compromise (e.g. no web tests).

More info on the PR itself, please let me know what you think, and thanks again for making this possible!

Issue #1116, opened 2022-11-30
Reference-style links are not clickable in preview view


thiskurt:
### Describe the bug

When using a reference-style link instead of a wikilink or classic markdown link the link is not accessible in the preview mode, when hovering over the link it reports as 'link to non-existent resource' and clicking on it will not take you to the correct file.

![referencelink_non-existing](https://user-images.githubusercontent.com/15367288/204778809-67028d9a-0905-494e-a33a-11d4eaf6b36a.png)

This only applies in the preview window, in the regular editor the link is still accessible as expected.

![referencelink_editview](https://user-images.githubusercontent.com/15367288/204778828-0d063e79-5e1a-45b1-a39a-c4dc9ffd29aa.png)

### Steps to Reproduce the Bug or Issue

1. Create two files
2. Link to the second file from the first file using a reference-style markdown link
3. Show the preview of the page and hover over the link/try to click it to notice it is not accessible that way

### Expected behavior

The link in the preview should behave the same as in the editor and the same way wikilinks and classic links behave at the moment.

### Operating System Version

Windows

### Visual Studio Code Version

1.73.1

pderaaij:
Hi,

I was looking into this issue to see if I could find a fix. However, when looking at your example I am wondering if we would really want to mix the wikilink syntax with reference-style links. Looking at your example I would wonder if using an alias isn't the appropriate solution here? Like `[[Page 2.md|SecondPage]]`.

Perhaps you can explain more about your use case?

riccardoferretti:
Both approaches are valid and theoretically already supported.
I believe the workaround described can work, but we should nonetheless fix this issue.

The reason this is not resolving in the same way is that in preview we don't have access to the "source" file, therefore we cannot fully resolve the link (with our current approach).

This probably requires a bit of thinking, we have a couple of issues with preview rendering that are due to the context provided and the fact that the work needs to be done synchronously.

badsketch:
> The link is not resolving in the same way in preview because in that context we don't have access to the "source" file, therefore we cannot fully resolve the link (given our current approach).

Any tips on how to get started on this 🙂? I see what you mean by not having access to the source file. Is there a place that's an example of getting the current open file? Was thinking it would have to be passed into `preview` when its' initialized with the other features in the main `extension.activate()`, but that part seems to do bootstrapping for all files, rather than the current one. Any advice appreciated!

riccardoferretti:
[...]

PR #1295, opened 2023-10-11
Change reading file from native fs to vscode alternative

pderaaij:
This PR changes the lasst occurence of reading files natively from `fs`. This is required for using Foam on the web version of vscode, but also follows best practices.

### DIFF
--- a/packages/foam-vscode/src/features/preview/wikilink-embed.ts
+++ b/packages/foam-vscode/src/features/preview/wikilink-embed.ts
@@ -177,7 +177,11 @@ function fullExtractor(
   parser: ResourceParser,
   workspace: FoamWorkspace
 ): string {
-  let noteText = readFileSync(note.uri.toFsPath()).toString();
+  let noteText;
+
+  vsWorkspace.fs
+    .readFile(toVsCodeUri(note.uri))
+    .then(f => (noteText = f.toString()));
   const section = Resource.findSection(note, note.uri.fragment);
   if (isSome(section)) {
     const rows = noteText.split('\n');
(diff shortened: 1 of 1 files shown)


packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -177,7 +177,11 @@ function fullExtractor(
   parser: ResourceParser,
   workspace: FoamWorkspace
 ): string {
-  let noteText = readFileSync(note.uri.toFsPath()).toString();
+  let noteText;
+
+  vsWorkspace.fs
```
riccardoferretti, line 182: Are you sure this will work?
I am not sure this would work, as this call returns a promise, and whether `noteText` is actually set when needed (e.g. line `192`) becomes a matter of timing.
This is the reason why originally we didn't migrate this part of the code from `readFileSync`.

Am I missing something?

riccardoferretti (commented):
Thanks for pulling out this change, I believe it's great to have the code itself ready for the transition, so that the rest is just a matter of build and testing.

I have a question about the approach though, see my comment in the diff.

packages/foam-vscode/src/features/preview/wikilink-embed.ts
pderaaij: It is not working. Need to refine it, but needed a test run on thr CI as I had some issues locally. Forgot to add wip to the PR ;)

pderaaij:
I am not sure how we can get this working. `markdown-it` is synchronous by design. On that same note, `vascode` is asynchronous intentional. Two things that doesn't work well together. I am not sure how to continue. The only hunch I have is to change its implementation from a regex replace to a custom renderer. But that would require quite some research.

Any ideas moving forward are welcome. This is a blocking issue to get Foam working properly on the web (#749).

riccardoferretti:
Yup, that limitation of `markdown-it` is quite something. Here are some thoughts I had about working around it, let me know what you think: https://github.com/foambubble/foam/issues/1116#issuecomment-1757236795

pderaaij:
Not eligible any more. We'll be taking a different approach.

PR #1302, opened 2023-10-26
extension descriptions now reflect .vscode dir

dcarosone:
including that image pasting is now supported natively

### DIFF
--- a/docs/user/getting-started/recommended-extensions.md
+++ b/docs/user/getting-started/recommended-extensions.md
@@ -5,13 +5,13 @@ These extensions defined in `.vscode/extensions.json` are automatically installe
 This list is subject to change.
 
 - [Foam for VSCode](https://marketplace.visualstudio.com/items?itemName=foam.foam-vscode) (alpha)
-- [Markdown All In One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one)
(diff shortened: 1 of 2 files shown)


dcarosone:
But I'm unclear why we no longer recommend Markdown All-in-One?
I've moved it below the break, unless there's a reason it's no longer recommended at all?  PR marked draft until this is clarified.

riccardoferretti:
My bad, markdown all in one is still recommended

dcarosone:
> My bad, markdown all in one is still recommended

Ok then it should still be in the `.vscode/extensions.json` ?

riccardoferretti:
Yup, I am fixing that shortly. I also need to figure out why it's not copying the directory in the right place. Will be done in the next couple of hours

riccardoferretti:
Thanks for pointing it out, I had missed that I had removed markdown AIO from the list of extensions, I was thinking of markdown-notes when I removed that line, thinking it was still on the list although we no longer use it

dcarosone:
> I also need to figure out why it's not copying the directory in the right place.

In the GHA? Wasn't that because of my typo (extra s)? With that fixed, the additional mkdir can also be removed.

riccardoferretti (approved):

riccardoferretti:
Yup the issue was around that. All fixed now, thanks!

PR #1311, opened 2023-12-10
Fix #1298. Wikilink definitions links not working.

### LINKED ISSUE #1298: Wikilinks activated: special characters make the links to not work
### Describe the bug

Creating a link with special characters when the wikilinks are activated, makes the link to stop working at saving the file (which is the time where the wikilinks are created).

_No response_

**Example 1:**
1. I create a link: [[Select a choice: Multiple ones]]
2. I press my keys to enter/create the note.
3. I can see the Select a choice: Multiple ones.md file is created.
4. I save the file containing the link.
5. It does some magic, so it adds this at the end of the file: `[Select a choice: Multiple ones]: <Select a choice: Multiple ones> "Select a choice: Multiple ones"`.

6. Now, the link has the "this link is not linked to any note" color.
7. If I click on the link now, it creates a file called Multiple ones.md

**Example 2:**

1. I create a link [[Música]]
2. I enter the link, and it creates the file Música.md
3. I save the original file containing the link.
4. Magic happens and this is appended: [Música]: Música "Música"
5. The link is in the "this link points to a created note" state, but when I click on it, it goes to the bottom of the same file, where this [Música]: Música "Música" thingy is...

Note: it could be related to `:` in the title changed to `_` in the file name. Example: `P2 - Alignment_ Excerpts in emails having links.md` instead of `P2 - Alignment: Excerpts in emails having links.md`

### Expected behavior

The link should take you to the right note. 
The note filename should be the same as the link (for example, if the link has `:` don't replace it with `_`). 
If the wikilink is defined in the file, clicking the link should still take you to the note, not to the definition at the bottom of the file.

### Screenshots or Videos

_No response_

### Operating System Version

macOs: Sonoma 14.0

Version: 1.83.1 (Universal)

### Additional context

I write in Spanish, so my file names commonly use `ñ` and tildes (`á`).

MABruni:
This is a fix for #1298.

The issue was that link creation was being done literally, and some special characters can break URLs, for example:
- Create placeholder note [[test:1]]
- Click on the link to create the note and save.
- **Update wikilink definitions**: This creates a link that is interpreted as linking to a note called 1 instead of test:1
- The original note changes to a placeholder note because now it links to a note called "1" that doesn't exist.

The solution involves using encodeURIComponent() to properly encode special characters in the generated links. This ensures that links work with several unsafe special characters that were breaking links before (:, ?, ^, {, }, ~, %).


riccardoferretti (commented):
That looks great, thanks @MABruni! Could you add a test for it?

MABruni:
Thank you, @riccardoferretti! 

I have added two tests, one to check that encoding is handled correctly and the original filename is preserved and another one to check that links generated resolve correctly. Both of them test wikilinks containing the most common symbols that can break a URL. I also had to modify some previous tests to handle the encoded links. 

I tried to be consistent with previous tests and I checked that all of them pass; but let me know if there is anything that I may have overlooked!

riccardoferretti (approved):
Looking good thx!

riccardoferretti:
@allcontributors add @MABruni for code

PR #1322, opened 2024-01-12
Fix #1321. Link references escaping HTML.

### LINKED ISSUE #1321: Link references escape html in v0.25.6
### Describe the bug

The link reference definitions have suddenly started using html escaping for slashes which breaks links.

E.g. 
````
I want to link to [[another page]] but this produces the reference below.

[//begin]
[another page]: ..%2F..%2Fanother%20page.md "another page"
[//end]
````

This is broken in v0.25.6. Downgrading to v0.25.5 fixes the issue.

I also don't see why #1311 decided to replace the (valid in markdown) spaces in angel brackets with the less readable %20 but unlike replacing the path separators (i.e. /) with %2F this doesn't break the links.

_No response_

1. Create some page
2. Create a page in a different directory linking to the first
3. Update the link reference definition

### Expected behavior

The definition should look like this:
````
[//begin]
[another page]: ../../another%20page.md "another page"
[//end]
````
or this
````
[//begin]
[another page]: <../../another page.md> "another page"
[//end]
````

### Screenshots or Videos

_No response_

### Operating System Version

Windows

1.85.1

### Additional context

Foam v0.25.6, downgrade to v0.25.5 fixes issue

MABruni:
This is a fix for #1321.

EncodeURI was targeting the entire URL. This PR makes the following changes:

- Only filenames are targeted by encodeURI now. This preserves especial characters in the base path. Notes with names that contain especial characters are encoded and link correctly inside VSCode.
- Blank spaces are not encoded now. Notes that contain them are back to using angel brackets.

Feedback would be greatly appreciated. Thank you!


MABruni:
The base path should be folder names, so I did not cover that case (for example, windows doesn't allow using special characters in folders). But, I could change the logic to encode everything except the forward slashes, that should take care of special characters everywhere and preserve link functionality across HTML and markdown.

If the filename has spaces in it the behavior is the same as before #1311. The wikilink url is enclosed in angle brackets and links correctly. For example, if a link is created to a page titled 'another page' it will appear like `<docs/features/another page>`. Honestly, I prefer having consistency and using percent encoding for everything, but I agree that using angle brackets makes it more readable. Both options work, so I'd be happy to go with your preferred one. 

With regard to #1298, percent encoding is considered escaping special characters while preserving link functionality, and it is my understanding that percent encoding gives broader compatibility (works on HTML and markdown) and is the standard for urls. That is why I decided on that approach.

riccardoferretti (approved):

PR #1344, opened 2024-03-17
Rename placeholder on note creation

### LINKED ISSUE #1327: Creating a note from a template via wikilink doesn't work if the template has a custom filepath
### Describe the bug

When using the tooltip to create a note from a template, and the template has a `filepath` specified, the wikilink is not linked up. I assume this is because the filepath no longer matches the text in the wikilink.

_No response_

- Create a template that changes the filepath of notes created with it
- Create a wikilink, and use the hover tooltip "Create note from template" option to create the note
- Note is created with the correct filename, but the wikilink is not updated with the right name

Example Template:
```md
---
foam_template:
  name: 'Example Template'
  description: 'An example for reproducing a bug'
  filepath: '$FOAM_DATE_DATE-$FOAM_SLUG.md'
---
```

### Expected behavior

The text of the wikilink should be updated to match the new notes title. 
For example:
- Given a wikilink `[[test]]`, and a template that creates a file named `2024-01-26-test.md`
- I would expect that the wikilink gets updated to either `[[2024-01-26-test]]`, `(test)[2024-01-26-test]` (probably not accurate, since it creates a markdown link instead of a wikilink), or create a link reference definitioon at the bottom of the page

### Screenshots or Videos

https://github.com/foambubble/foam/assets/7729878/5828f8ba-8a21-4b02-8efc-6201016bbfb5

### Operating System Version

macOS 14.3

1.85.2 (Universal)

### Additional context

_No response_

riccardoferretti:
This PR fixes #1327

Passing to the create-note command a source link, in which case it will check whether, once the note has been created (possibly using a template that changed the name of the original placeholder), the name of the new note matches the originating link - and if not updates it.

I made a conscious decision here to only update the source link, and not all placeholders that match the source link. 
Happy to have a conversation about that, for now it felt better this way.

Oh yeah, and I was playing with some logging stuff (minor)

### DIFF
--- a/packages/foam-vscode/src/core/model/foam.ts
+++ b/packages/foam-vscode/src/core/model/foam.ts
@@ -5,7 +5,7 @@ import { FoamGraph } from './graph';
 import { ResourceParser } from './note';
 import { ResourceProvider } from './provider';
 import { FoamTags } from './tags';
-import { Logger } from '../utils/log';
+import { Logger, withTiming, withTimingAsync } from '../utils/log';
 
 export interface Services {
   dataStore: IDataStore;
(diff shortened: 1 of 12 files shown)


ashtuchkin:
The editor is super slow after this change for me (specifically v0.25.9), looks like it's recalculating the orphans and other stuff after every change, see this video:
https://github.com/foambubble/foam/assets/627997/0f30182f-65c4-4d47-9811-ae665fc4a93f

I've checked that this doesn't happen in 0.25.8.

riccardoferretti:
v0.25.10 should fix this

ashtuchkin:
Thank you!

Issue #1353, opened 2024-04-29
Option to use internal standard Markdown Links instead of WikiLinks


PhilB-RLB:
### Is your feature request related to a problem? Please describe.

The feature is not related to a problem.

### Describe the solution you'd like

I use Foam to manage my markdown files library and then I use Hugo to generate a static site to browse my PKM. As you may know Hugo isn't compatible with WikiLinks.

So for those who use Foam and a static site generator as Hugo it would be great to have to possibility, via option in configuration, to use internal standard Markdown Links instead of WikiLinks. This has been done in this [foam fork](https://github.com/theowenyoung/foam). But this fork is not complete on only some feature proposed by Foam are available.

Would it be possible to implement this fork into the Foam code and let the user choose between WikiLinks or Standard Internal Markdown Links ?

### Describe alternatives you've considered

No alternative considered.

riccardoferretti:
That's an interesting idea, basically a setting that would expand wikilink autocompletion into a standard md link instead.
I am open to it, I don't believe it should be a crazy amount of work, I am open to someone looking into it and get my support to get it done.

PhilB-RLB:
:+1:

stereobooster:
duplicate or related: https://github.com/foambubble/foam/issues/1319, https://github.com/foambubble/foam/issues/791

n8layman:
I also like this idea because it would make it easier to use foam with Quarto.

riccardoferretti:
This has been released some time back, both as a configuration option, or via command to convert a specific link.

Issue #1319, opened 2024-01-11
Link manipulations


KillyMXI:
### Is your feature request related to a problem? Please describe.

Wikilinks are convenient where they are supported, but if I need to keep the document compatible - I'd like to have an option to convert them to equivalent Markdown links.
Currently there seems to be no way to do it automatically and brackets autocompletion fights me when I try to do it manually.

### Describe the solution you'd like

Introduce commands/refactorings to change between link types. Refactorings for individual links, commands for whole document.

For example, Obsidian plugin that does the same thing: <https://github.com/mii-key/obsidian-links>
(Not all of it might be necessary in context of Foam)

### Describe alternatives you've considered

- Separate vscode extension
  - I can't find any vscode extension that would do the same thing;
  - If there were such extension - it most likely introduces its own wikilink support, that might lead to conflicts;
- Text replace
  - Manually writing regex every time I need this feature - is a chore that should be avoidable;
- Text replace rules extensions
  - Extension example: <https://github.com/bhughes339/vscode-replacerules>
  - Should allow to write replace rules once;
  - General purpose find-and-replace, not aware of any context (can't use document title for example);
  - Might not support all scenarios (runs on full document, not individual links).

riccardoferretti:
Although I wouldn't say it's a priority for Foam, I like the idea, and given the model Foam has of the system, it don't expect it to be super complicated to implement. If anyone wants to look into it I will be happy to provide pointers

riccardoferretti:
The core implementation of this has been done (see #1525) and will be released in the next version (today).
For now the only available commands are for the conversion of wikilink to mdlink and viceversa for the link under the cursor (and it's used by the config to automatically convert completed links).
In the future we might add commands to do the update for a whole file or the whole repo, should be trivial at this point but will leave it to avoid bloat.

riccardoferretti:
Will close this for now as the current implementation feels appropriate

Issue #791, opened 2021-10-19
Support linking using full path relative to repo root 


ais:
Links that start with `/` leave placeholders in the graph. Users in our org use a web tool to create files, and links in this tool are expected to be absolute. We would like to use the graph to get an overview of our knowledge and improve the structure of our documentation.

<details>

![image](https://user-images.githubusercontent.com/5240292/137931864-ff69a325-328e-451f-a075-0772e48abb04.png)

</details>

riccardoferretti:
related conversation: https://github.com/foambubble/foam/issues/806#issuecomment-961326199

SHiLLySiT:
I'd also like to see this work in Foam, since out-of-the-box paths that start with `/` in a Markdown link are interpreted as paths relative to the project root.

From the [VS Code docs](https://code.visualstudio.com/docs/languages/markdown#_path-completions):
> Paths starting with / are resolved relative to the current workspace root, while paths staring with ./ or without any prefix are resolved relative to the current file.

As a work around, Foam correctly links a file instead of adding a placeholder in the graph if you add `..` in the path to navigate up folder(s). 

Current, undesired behavior:
![image](https://github.com/foambubble/foam/assets/950984/4f890239-1c60-48d1-b88c-64a163d865b1)

Workaround:
![image](https://github.com/foambubble/foam/assets/950984/e1508066-b03d-4c30-988e-809fe3c31a0f)

stereobooster:
IMO links should work the same as in VSCode and Github e.g. `[Readme.md](/Readme.md)` should be supported. Related [Portable markdown links](https://stereobooster.com/posts/portable-markdown-links/)

riccardoferretti:
Thanks for the comment and pointer @stereobooster and @SHiLLySiT - this is actually a good reference point.

I agree Foam should follow the VS Code convention, which is also compatible with @stereobooster 's portable md links.

wdhongtw:
I have a minimal working version. The extension can resolve absolute path correctly and the graph function also works.

- in `FoamWorkspace.find`, try to join workspace folder with the absolute path.
- inject workspace folder during setup process.

I would like to ask

1. whether is this change a good way to implement this feature? and
2. should we consider multi-folder workspace scenario? if so, how?

@riccardoferretti 

If the idea looks good, maybe I can help with this recently. 😊

Following are the minimal patch for the PoC.

```diff
commit 23bfece00e3f33c8cd3bc690ea86071b625ab744
[...]

PR #1365, opened 2024-06-21
generate copy without wikilinks

hereistheusername:
# Abstract

As mentioned in #1353 , #1319, and #791, many people got difficulties when working with other tools that don't support wikilinks. To tackle this problem, I add a function to convert link format.

# Further Explanation

## Implementation

Refactoring the `MarkdownLink.createUpdateLinkEdit`, the `convertLinkFormat` function resolves links based the note containing it and its workspace and combine components properly.

## User Interface

4 commands:
1. foam-vscode.convert-wikilink-to-markdownlink-inplace
2. foam-vscode.convert-markdownlink-to-wikilink-inplace
3. foam-vscode.convert-wikilink-to-markdownlink-in-copy
4. foam-vscode.convert-markdownlink-to-wikilink-in-copy

to meets users' different needs.

Thanks to @riccardoferretti help.


packages/foam-vscode/src/core/janitor/generate-mdlinks.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return '';
+  }
+  public getSection(): string {
+    if (this.section_m) {
+      return this.section_m;
+    }
+    return '';
+  }
+  public toMarkdownLink(
```
riccardoferretti, line 54: You probably can reuse the mechanisms available in `MarkdownLink` to achieve the same in a more integrated way

riccardoferretti (commented):
Thanks for looking into this. 
At a high level, have you looked at `MarkdownLink.createUpdateLinkEdit`? it should do most of what you need in terms of refactoring.

Can you also tell me your thinking about having this as a command vs a formatting setting (kinda like using wikilink definitions)?
I would like to understand the feature from the POV of the user: is this about having a documenta that is always md compatible (in which case the transformation would happen on save), or is it about being able to export the knowledge base for e.g. a static site generator (in which case do we need a command for that?), or about exporting a single file (which is what you are implementing IIUC) - thoughts?

hereistheusername:
> Thanks for looking into this. At a high level, have you looked at `MarkdownLink.createUpdateLinkEdit`? it should do most of what you need in terms of refactoring.
> 
> Can you also tell me your thinking about having this as a command vs a formatting setting (kinda like using wikilink definitions)? I would like to understand the feature from the POV of the user: is this about having a documenta that is always md compatible (in which case the transformation would happen on save), or is it about being able to export the knowledge base for e.g. a static site generator (in which case do we need a command for that?), or about exporting a single file (which is what you are implementing IIUC) - thoughts?

Thank you for the quick review. Currently, I ran into a bug that it cannot resolve the wikilink to embeded images URI through `workspace.resolveLink(resource, link)`. I am working on it and I'll also check the `MarkdownLink.createUpdateLinkEdit`.

The reason for doing this is wikilinks are more flexible, which leaves the workspace to manage the resource path. For example, I usually start with a single draft, and with more resources are added in the idea gets mature. At some point, I would like to re-organize or archive a group of ideas. If files are linked by wikilinks, I can move them freely. So, I think keeping linkages as wikilink would be benefit for PKM. However, if choosing format, there could be chance that links get broken, and it will need efforts to fix them.

In my imagination, users are recommanded to use wikilinks and only need to generate a standalone copy when they want to "export" their ideas. After that, those copies can be wiped out, by manually or scripts. Single file is not neccessary. Providing options to export a list of selected files or a whole directory would be helpful.

packages/foam-vscode/src/features/commands/generate-standalone-note.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const fileUri = vscode.Uri.file(
+    `${
+      basePath ? basePath + '/' : ''
+    }${resource.uri.getName()}.standalone${resource.uri.getExtension()}`
+  );
+  const encoder = new TextEncoder();
+  await vscode.workspace.fs.writeFile(fileUri, encoder.encode(text));
+  await window.showTextDocument(fileUri);
+  await commands.executeCommand('foam-vscode.update-wikilink-definitions');
```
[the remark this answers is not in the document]
hereistheusername: Since wikilinks are wiped out, executing it can empty definitions safely.

packages/foam-vscode/src/core/janitor/convert-links-format.ts
```diff
@@ -0,0 +1,103 @@
...
+ * exception.
+ * @param link
+ * @param targetFormat
+ * @param workspace
+ * @param note
+ * @returns LinkReplace { newText: string; range: Range; }
+ */
+export function convertLinkFormat(
+  link: ResourceLink,
+  targetFormat: string,
```
riccardoferretti, line 27:
```suggestion
  targetFormat: 'wikilink' | 'link',
```

packages/foam-vscode/src/core/janitor/convert-links-format.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    /* if it's originally an embeded note, the markdown link shouldn't be embeded */
+    if (embed && targetRes.type === 'note') {
+      embed = '';
+    }
+    return {
+      newText: `${embed}[${alias}](${url})`,
+      range: link.range,
+    };
+  }
```
riccardoferretti, line 99: I haven't tested it locally, but I wonder if you could be able to replace this whole section with
```
const newLink = {...link, type: targetType}
const edit = MarkdownLink.createUpdateLinkEdit(newLink)
```

packages/foam-vscode/src/features/commands/convert-links-format-in-note.ts
```diff
@@ -0,0 +1,179 @@
...
+) {
+  const foam = await foamPromise;
+
+  /* 
+  commands:
+  foam-vscode.convert-wikilink-to-markdownlink-inplace
+  foam-vscode.convert-markdownlink-to-wikilink-inplace
+  foam-vscode.convert-wikilink-to-markdownlink-in-copy
+  foam-vscode.convert-markdownlink-to-wikilink-in-copy
+   */
```
riccardoferretti, line 44: I'd suggest having only one command, something like `foam-vscode.convert-link-style`, with a `linkStyle` property (which defaults to `markdown-link`). Running the command on a file with the links already in that style would result in no change, otherwise the changes are applied.

packages/foam-vscode/src/features/commands/convert-links-format-in-note.ts
```diff
@@ -0,0 +1,179 @@
...
+    textReplaceArr.forEach(edit => {
+      editorBuilder.replace(edit.range, edit.newText);
+    });
+  });
+}
+
+/**
+ * convert links based on its workspace and the note containing it.
+ * Changes happen in a copy
+ * 1. prepare a copy file, and makt it the activeTextEditor
```
riccardoferretti, line 143: makt?

packages/foam-vscode/src/features/commands/convert-links-format-in-note.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (!isMdEditor(editor) || !fMatcher.isMatch(fromVsCodeUri(doc.uri))) {
+    return;
+  }
+  // const eol = getEditorEOL();
+  let text = doc.getText();
+
+  const resource = fParser.parse(fromVsCodeUri(doc.uri), text);
+
+  const textReplaceArr = resource.links
+    .filter(link => link.type === convertOption.from)
```
riccardoferretti, line 120: (note: in my proposed approach, this check would be `link.type !== convertOption.to`)

riccardoferretti (commented):
PR is looking good, I left a few comments. Thanks for the explanation around the rationale, we are aligned and this feels like it's pushing things in the right direction

packages/foam-vscode/src/core/janitor/convert-links-format.ts
hereistheusername: accepted

packages/foam-vscode/src/core/janitor/convert-links-format.ts
hereistheusername: I think it's impossible. The first line of `MarkdownLink.createUpdateLinkEdit` is

```typescript
const { target, section, alias } = MarkdownLink.analyzeLink(link);
```

which extracts three components via regex from link.rawText based on link.type. If the pattern doesn't match, three of them may not be extracted correctly.


Moreover, when dealing with conversion from wikilink to link, we have to retrieve the relative path to target, which is usually a basename in wikilink's rawText. Therefore, we must have `workspace` and `note: Resource | URI`.

packages/foam-vscode/src/features/commands/convert-links-format-in-note.ts
hereistheusername: done

packages/foam-vscode/src/core/janitor/convert-links-format.ts
riccardoferretti: Yeah, good point. Basically instead in`createUpdateLinkEdit` we need to add `type?: 'wikilink' | 'link'` to the `delta` parameter, and take it from there. Does that sound good?

packages/foam-vscode/src/features/commands/convert-links-format-in-note.ts
```diff
@@ -0,0 +1,189 @@
+import { commands, ExtensionContext, window } from 'vscode';
+import { isMdEditor } from '../../utils';
+import { Foam } from '../../core/model/foam';
+import { FoamWorkspace } from '../../core/model/workspace';
+import { fromVsCodeUri, toVsCodeRange } from '../../utils/vsc-utils';
+import { ResourceParser } from '../../core/model/note';
+import { IMatcher } from '../../core/services/datastore';
+import { convertLinkFormat } from '../../core/janitor';
+const vscode = require('vscode'); /* cannot import workspace from above statement and not sure what happened */
```
riccardoferretti, line 9: can you try again to import `workspace` from vscode, or use `import * from vscode` and use that?
I am not introducing the `require`

riccardoferretti (commented):
Thanks @hereistheusername - it all looks good and I think it's good to go, except for the `require` use - can you please see if the proposed solutions work? I am not keen on introducing `require`

packages/foam-vscode/src/features/commands/convert-links-format-in-note.ts
hereistheusername: Done.

riccardoferretti (approved):
Great, thx!

riccardoferretti:
@allcontributors add @hereistheusername for code

Issue #1277, opened 2023-08-26
When using workspaces, wikilink are not working


PeterJ504:
### Describe the bug

When using workspaces, the wikilink is pointing to the incorrect directory.  Instead of `c:\directory\filename.md`, it's pointing to `c:\directory\directory\filename.md`.

## Create test files

 `file1.md`
   ```markdown
   # file 1
   ## Link
   [[file 2]]
   ## Display
   ![[file 2]]
   ```

   and `file2.md`

   ```markdown
   # file 2
   This is file 2
```

Display and links both work correctly.

## Create workspace and add a new folder
```
{
  "folders": [
    {
      "path": "."
    },
    {
      "path": "../folder2"
    }
  ],
  "settings": {}
}
```

Display works correctly but the link is broken.

### Expected behavior

As  a user, I expected the links to work the same way in a workspace,

### Screenshots or Videos

![screen_3](https://github.com/foambubble/foam/assets/16738926/ad6c5c1d-b239-4cc1-9975-93aa93d2f651)

### Operating System Version

Windows

### Visual Studio Code Version

Version: 1.81.1

riccardoferretti:
Thanks for reporting this. 

As a possible workaround, can you try having the workspace file in the parent directory:

{
  "folders": [
    {
      "path": "./folder1"
    },
    {
      "path": "./folder2"
    }
  ],
  "settings": {}
}


Will look more into it when I am at my laptop

PeterJ504:
I get the same error.

riccardoferretti:
When I try to reproduce your example, the issue I find is that the file is called `file2.md` whereas the wikilink is `[[file 2]]`. The name of the file and the wikilink do not match.

You should use `[[file2]]` 
Or name your file `file 2.md`

PeterJ504:
I apologize for screwing up the the name when I typed it into the report but the even when I change the name as you suggest, the problem still occurs.  In the screenshot and the arrow that I highlighted, you can clearly see that the name is being picked up incorrectly. 

Do you want to reopen this report or shall I create a new one?

riccardoferretti:
Can you please attach a zip file with a small reproducible example?

PeterJ504:
[test.zip](https://github.com/foambubble/foam/files/12704107/test.zip)
Go into folder1 and open the project using the workspace file.
Open file1.md and then try to click on the links to file2 and file3.

riccardoferretti:
[...]

PR #1372, opened 2024-07-19
Exclude workspace when linking to a file

### LINKED ISSUE #1332: Wikilink are not working in preview mode when using workspace
### Describe the bug

Wikilink are not working in preview mode when using workspace. and this issue cannot be reproduced when open folder directly. and cannot be reproduced in source editor mode.

hope my description is clear, sorry for my English.

https://github.com/chaleaoch/template_foam

Attached a .gif for your information.
gif

### Expected behavior

could jump to correct path.

### Screenshots or Videos

_No response_

### Operating System Version

windows

1.86.1

### Additional context

_No response_

pderaaij:
This fixes #1332 and #1277.

`vscode.workspace.asRelativePath` needs to be told to exclude the workspace, as we already have in the resource location.

### DIFF
--- a/packages/foam-vscode/src/features/preview/wikilink-navigation.ts
+++ b/packages/foam-vscode/src/features/preview/wikilink-navigation.ts
@@ -42,12 +42,18 @@ export const markdownItWikilinkNavigation = (
           return getPlaceholderLink(label);
         }
 
-        const link = `${vscode.workspace.asRelativePath(
+        const resourceLabel = isEmpty(alias)
+          ? `${resource.title}${formattedSection}`
+          : alias;
+        const resourceLink = `/${vscode.workspace.asRelativePath(
           toVsCodeUri(resource.uri),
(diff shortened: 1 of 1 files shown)


riccardoferretti:
Thanks @pderaaij - a couple of things:
- can you add a test?
- does this work well when NOT in a workspace?
- is this the only place where such change would be required?

pderaaij:
Will add a test. Focussed on #1290 first.

pderaaij:
@riccardoferretti I am bit lost on the best way to add a test for this situation. The end-2-end test suite runs a single-root workspace. I can't see a way how to manipulate this on runtime. The alternative is to create an end-2-end test suite for a multi-root workspace, but that might be a bit too much. Do you have any thoughts on how to handle this?

riccardoferretti (approved):
Fair point, manual testing will do for now

Issue #1376, opened 2024-07-27
🆕 Add option to create new notes from Explorer


oviniciusfeitosa:
## Feature Description
Currently, it's not possible to create new notes directly from the Foam Explorer. It would be great to add this option.

## Proposal
Add the option "foam.files.showNewNoteInExplorer": true to the configuration file to allow creating new notes directly from the Explorer.

## Benefits
- Facilitates quick note creation.
- Improves the user experience when navigating the Explorer.

## Context
This feature would make Foam even more efficient for managing knowledge and notes.

Thank you for your attention!

PR #1377, opened 2024-07-27
🆕 Add option to create new notes from Explorer 

oviniciusfeitosa:
References: #1376


pderaaij:
I think this won't be context aware. Meaning, it will not create a note on that location you trigger that action. That might be weird UX.. is there a way that we can forward the location in which the user triggered this call?

Issue #1375, opened 2024-07-24
[RESEARCH] Performance of Foam in large projects


pderaaij:
### Describe the bug

Various users reported performance issues with large notes or projects with many notes. This issue serves as a collection of those reports. It acts as the documentation of ongoing research on performance.

### Steps to Reproduce the Bug or Issue

..

### Expected behavior

We want to optimise the performance of Foam, even in large projects.

### Operating System Version

All OS

### Visual Studio Code Version

Latest at least

pderaaij:
For now, I am doing some tests and research with https://github.com/github/docs. A relatively large project with many markdown files that I can use for research.

pderaaij:
Just came across https://github.com/rcvd/interconnected-markdown/tree/main/Markdown. This contains not only many notes, but also highly linked. Great for researching the performance of Foam

pderaaij:
Some initial investigation points to the function `listByIdentifier` in `workspace.ts` This function is used in both `find` as in `wikilink-diagnostics.ts`. 

The function is defined as:

```js
  public listByIdentifier(identifier: string): Resource[] {
    const needle = normalize('/' + identifier);
    const mdNeedle =
      getExtension(needle) !== this.defaultExtension
        ? needle + this.defaultExtension
        : undefined;
    const resources: Resource[] = [];
    for (const key of this._resources.keys()) {
      if (key.endsWith(mdNeedle) || key.endsWith(needle)) {
        resources.push(this._resources.get(normalize(key)));
      }
    }
    return resources.sort(Resource.sortByPath);
  }
```

For my test repo the`this._resources` is a Map of 10,000 entries. This function is called for every processing of a wikilink. Both on boot and graph update. I am thinking that the for loop is too inefficient for large projects. Whether it is many notes or many links. 

I will do some experiments with optimising the for loop in this area and see if that boosts performance.

DrakeWhu:
I am interested in this. I have a graph of aproximately 3k notes and 8k links. I also have some Python scripts I am using to make community detection over the graph. It results in around 50~100 or more communities. On the visualization, everything shows no problem.

[...]

PR #1391, opened 2024-09-02
[WIP] improve graph performance

pderaaij:
Based on some performance findings found #1375 I started to experiment with some improvements. I've noticed the function`listByIdentifier` being called alot by other functions. In this function, there is a `foreach` loop running over all resources. A function which works fine for small notebooks, but large notebooks seems to be hit exponentionally. 

In this concept I've added a new state object that lists all resources for an identifier. In this case, I choose the identifier to be the file-name of the note. Using that identifier, I list all resource paths with that same identifier. Using this I can remove the `foreach` loop and just use the identifier to list a small set of resources. 

For the 10000 linked notes folder from this repo (https://github.com/rcvd/interconnected-markdown/tree/main/Markdown) this reduces the startup time to:

```
Foam Logging: info
[info - 21:39:06] Starting Foam
[info - 21:39:07] Loading from directories:
[info - 21:39:07] - /Users/paulderaaij/projects/interconnected-markdown/Markdown/10000s/10000
[info - 21:39:07]   Include: **/*
[info - 21:39:07]   Exclude: **/.foam/**,**/.vscode/**/*,**/_layouts/**/*,**/_site/**/*,**/node_modules/**/*,**/.git,**/.svn,**/.hg,**/CVS,**/.DS_Store,**/Thumbs.db
[info - 21:39:15] Workspace loaded in 7755ms
[info - 21:39:15] Graph loaded in 660ms
[info - 21:39:15] Tags loaded in 4ms
[info - 21:39:15] Loaded 10001 resources
```

Using the current master the numbers are:

```
Foam Logging: info
[info - 21:46:53] Starting Foam
[info - 21:46:53] Loading from directories:
[info - 21:46:53] - /Users/paulderaaij/projects/interconnected-markdown/Markdown/10000s/10000
[info - 21:46:53]   Include: **/*
[info - 21:46:53]   Exclude: **/.foam/**,**/.vscode/**/*,**/_layouts/**/*,**/_site/**/*,**/node_modules/**/*,**/.git,**/.svn,**/.hg,**/CVS,**/.DS_Store,**/Thumbs.db
[info - 21:47:01] Workspace loaded in 8147ms
[info - 21:47:27] Graph loaded in 25513ms
[info - 21:47:27] Tags loaded in 2ms
[info - 21:47:27] Loaded 10001 resources
```

A delta of almost 25 seconds for loading the graph. Not only a performance improvement for loading the workspace, but also for some other operations that uses `listByIdentifier` directly or via `find`. This should make several UI components work better and more responsive.

The downside is that it adds a new stateful object, but I don't see an other way to increase the performance of the foreach loop. 

@riccardoferretti curious to your thoughts on this.


packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -70,17 +109,39 @@ export class FoamWorkspace implements IDisposable {
...
+    // Get all resources matching the needle(s)
+    if (this._identifiers.has(needle)) {
+      for (const resource of this._identifiers.get(needle)) {
+        if (resource.endsWith(normalize('/' + identifier))) {
+          resources.push(this._resources.get(resource));
+        }
       }
     }
+
+    if (needle === '/page b.md') {
```
pderaaij, line 128: I need to remove this ;-)

riccardoferretti:
This is amazing @pderaaij ! I am back from holidays (you know, italy + august = holiday) so will look a bit more into the PR this week

riccardoferretti:
Sharing some high level thoughts here.
As I was doing some research in the approach, I came across the `trie` data structure, which fundamentally feels like what we are kinda implementing here.
As I looked into it, I realized that using the reversed path (similar to what we do in `getShortestIdentifier`) and saving in in a trie would be the cleanest solution, as it would out of the box replace the existing indexing while giving us the features and performance we want.

What do you think? Again, this was at a cursory look, I might be missing something

pderaaij:
Never really looked into the trie datastructure. Will do some research and fiddle around!

pderaaij:
I need a bit more time to figure out the right approach. Have been looking into tie data structures, but most implementation I found are immutable. A characteristic we don't want for our use case here. So, I need to find an implementation that does allow mutations. Did a quick checkout of https://rimbu.org, but couldn't find a right data structure off the bat for our use case. So, I require some more time to look around and find a solution. Ideas welcome of course.

pderaaij:
I got a long way tonight using https://yomguithereal.github.io/mnemonist/trie-map. Need to finish some things and get the last use cases working. Will do a new performance test when things are working. Hopefully I can finish it later this week.

pderaaij:
@riccardoferretti I've updated the implementation with my WIP so far. 

It has a little impact on performance. Compared to the earlier version, I now load the `10000s` workspace in 2 to 3 seconds. Which still is a massive improvement compared to the current implementation. I do feel the performance impact is acceptable given the ease of maintenance this implementation brings in the code.

I have one remaining issue and I hope you can help. Originally, a Map holds the insertion order. TrieMap obviously does not do that. But, this means that `resources()` needs to return a sorted Iterator. I can't find a correct way to sort the values and then return an `IteratableIterator`. Perhaps you have an idea here?

If we can resolve this issue, all tests will be green so we can do some further testing and enhancing before shipping it.

pderaaij:
@riccardoferretti it seems I was on the right track, but simply not sharp after a long day of work. Got it working now. Please review and check accordingly.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -57,7 +62,11 @@ export class FoamWorkspace implements IDisposable {
   }
 
   public resources(): IterableIterator<Resource> {
-    return this._resources.values();
+    const resources: Array<Resource> = Array.from(
+      this._resources.values()
+    ).sort(Resource.sortByPath);
+
+    return resources.values();
```
riccardoferretti, line 69: do we need the `.values()` again on this line?

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -70,17 +79,24 @@ export class FoamWorkspace implements IDisposable {
   }
 
   public listByIdentifier(identifier: string): Resource[] {
-    const needle = normalize('/' + identifier);
+    let needle = this.getReversedIdentifier(identifier);
```
riccardoferretti, line 82: I see you moved the addition of `\` from here into the `getReverseIdentifier` fn. I wonder if there is any side effect from doing that?

Tbh, I am kinda relying on the tests passing here, I find this code (the existing one, that I wrote :) ) hard to properly process, I think at some point I will need a round to make it clearer for my future self and others :)\

I think fundamentally the semantics we want here is that resources can be found by either the identifier as is and by the identifier + default extension. Maybe we save both keys in the Trie.

In any case, just sharing some thoughts, I am ok with moving forward with the PR before sorting this out.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -119,9 +135,32 @@ export class FoamWorkspace implements IDisposable {
     return identifier;
   }
 
+  /**
+   * Returns a note identifier in reversed order. Used to optimise the storage of notes in
+   * the workspace to optimise retrieval of notes.
+   *
+   * @param reference the URI path to reverse
+   */
+  private getReversedIdentifier(reference: URI | string): string {
```
riccardoferretti, line 144: somewhat minor, but I think it makes the code easier to read: can we call this either `getTrieIdentifier`?

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -135,7 +174,9 @@ export class FoamWorkspace implements IDisposable {
           : isSome(baseUri)
           ? baseUri.resolve(candidate).path
           : null;
-        resource = this._resources.get(normalize(searchKey));
+        resource = this._resources.get(
+          normalize(searchKey).split('/').reverse().join('/')
```
riccardoferretti, line 178: why don't you use `getReversedIdentifier` here?

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -6,6 +6,7 @@ import { Emitter } from '../common/event';
 import { ResourceProvider } from './provider';
 import { IDisposable } from '../common/lifecycle';
 import { IDataStore } from '../services/datastore';
+import { TrieMap } from 'mnemonist';
```
riccardoferretti, line 9: minor: should we import from `mnemonist/trie-map` to help with tree shaking?

riccardoferretti (commented):
Yes this is looking like the right approach, I left a few comments. Let me know what you think

packages/foam-vscode/src/core/model/workspace.ts
pderaaij: It seems so. The only way I can see to return an IterableIterator

packages/foam-vscode/src/core/model/workspace.ts
pderaaij: So far, I haven't found any side effect. I guess it was done originally to always have a start of a path.

I agree with optimising the logic on a next round. Also to solve #1399.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -70,17 +79,24 @@ export class FoamWorkspace implements IDisposable {
...
+        ? this.getReversedIdentifier(identifier + this.defaultExtension)
         : undefined;
+
     const resources: Resource[] = [];
-    for (const key of this._resources.keys()) {
-      if (key.endsWith(mdNeedle) || key.endsWith(needle)) {
-        resources.push(this._resources.get(normalize(key)));
+
+    this._resources.find(needle).forEach(elm => {
+      if (elm[0].indexOf(getExtension(normalize(identifier))) > 0) {
```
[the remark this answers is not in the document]
pderaaij: I guess this was a left over from reverse engineering how this worked :) Could be removed indeed.

pderaaij:
@riccardoferretti somehow I managed to break the build in CI. Locally all works, but can't get it going on Actions. Could you have a look?

riccardoferretti:
@pderaaij I think it's related to the markdown-it version change

pderaaij:
Yeah, I was hoping that would fix it. Had the problems after rebasing the lock files. Which was a hassle on its own....

pderaaij:
Well... I found the little bugger:

https://github.com/markdown-it/linkify-it/issues/111

pderaaij:
I am giving up on this PR. Going to do a fresh branch on master and open a new PR. Not trusting all the mess this has created.

PR #1395, opened 2024-09-16
Foam Web Extension

riccardoferretti:
This PR is **heavily** inspired and influenced by #1290 - thanks @pderaaij for spearheading and working through this complex feature.
See that PR for more context.

Compared to #1290 , this PR:
- does not introduce webpack, mocha, and double TS settings
- removes (for now) support for testing the web extension (basically, we are still, only, testing the node extension). I know, that's not great, but given the limited testing we were able to do anyways, it feels like a reasonable trade off for the reduced complexity
- avoids some extra dependencies for the migration where possible (e.g. using a sha library that works in both node and browser)

This way:
- we have a working web extension (although a bit brittle as it has no testing of its own - but then again I don't expect many changes that would impact the web extension without affecting the native one)
- we have the code in a place where improvements on this side (including moving to webpack etc) can be done incrementally, as the feature is already there

The only thing I would have liked to add was running a basic test in the web extension (without the need for `mocha`) that would only test if the extension is present. Then again, this can be added as a second step.


packages/foam-vscode/src/features/panels/utils/grouped-resources-tree-data-provider.ts
riccardoferretti, line 1: this is not strictly about this PR - should have not been here 😅

packages/foam-vscode/src/features/panels/notes-explorer.ts
riccardoferretti, line 1: this is not strictly about this PR - should have not been here 😅

pderaaij:
Looks good to me!

riccardoferretti:
Great @pderaaij ! Again, I want to empathise that you basically did all the hard work here, I just repackaged it to avoid those extra dependencies 🙏

PR #1406, opened 2024-10-11
Improve performance via Triemap in workspace

pderaaij:
The compromised version of #1391.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24025,6 +24025,14 @@
         "node": ">=10"
       }
     },
+    "node_modules/mnemonist": {
+      "version": "0.39.8",
+      "resolved": "https://registry.npmjs.org/mnemonist/-/mnemonist-0.39.8.tgz",
+      "integrity": "sha512-vyWo2K3fjrUw8YeeZ1zF0fy6Mu59RHokURlld8ymdUPjMlD9EC9ov1/YPqTgqRvUN9nTr3Gqfz29LYAmu0PHPQ==",
+      "dependencies": {
+        "obliterator": "^2.0.1"
+      }
+    },
     "node_modules/modify-values": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/modify-values/-/modify-values-1.0.1.tgz",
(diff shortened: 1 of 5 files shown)


pderaaij:
I am not sure why this test case is suddenly failing. I have to look at this with some fresh brains later this week. Always open for suggestions.

pderaaij:
@riccardoferretti On my local machine I have a passing test, but in CI I can't make it work. Could you have a look in your environment and see what happens? Perhaps you can spot a problem, I am missing it right now.

riccardoferretti:
(from Discord)

yeah I saw that. Weird thing was that I was able to reproduce it locally, then made a small change to the content of the definition of `fileC` (a few lines above) as I was wondering if it was some odd caching issue. (the reason was that the test fails for 5 chars, which is the difference between the wikilink in fileB and fileC - long shot but I tried, and it "kinda" fixed the immediate issue, the puzzling question became what is causing that). After the change the test passed, and the oddest thing was that after reverting the change it continued to pass. 
basically in the definition of `fileC` in the test, I made the string a bit longer ("File C is also connected to" -> "File C is also connected also to")

riccardoferretti:
Odd, but do you wanna try the same thing and see if it passes CI?

pderaaij:
Well... adding the text didn't help, unfortunately. But, it did help me to think about the problem differently. What I found is that the references are returned in no particular order. Depending on the random file name, the first wikilink might come in as first reference or as second. If the latter happens, that will fail the test. The random file names explain the flakiness of the test.

I expect this is an issue since this change. As TrieMap does not hold the insertion order, the issue manifests during the build. For other places I expect this not to be a problem as they are sorted in their view/UI panel.

Other than that, this PR is ready for review and testing. Just did a test with the 10000s folder, the activation log states:

```
Foam Logging: info
[info - 19:51:05] Starting Foam
[info - 19:51:07] Loading from directories:
[info - 19:51:07] - /Users/paulderaaij/projects/interconnected-markdown/Markdown/10000s/10000
[info - 19:51:07]   Include: **/*
[info - 19:51:07]   Exclude: **/.foam/**,**/.vscode/**/*,**/_layouts/**/*,**/_site/**/*,**/node_modules/**/*,**/.git,**/.svn,**/.hg,**/CVS,**/.DS_Store,**/Thumbs.db
[info - 19:51:14] Workspace loaded in 7253ms
[info - 19:51:17] Graph loaded in 2227ms
[info - 19:51:17] Tags loaded in 106ms
[info - 19:51:17] Loaded 10001 resources
```

Compared to 25+ seconds that the current release needs for loading that workspace.

riccardoferretti:
Ahh makes perfect sense, good find!

riccardoferretti (approved):
Amazing, thanks @pderaaij !

Issue #1404, opened 2024-10-09
Error creating file in .dev version: No file system handle registered


markschaver:
### Describe the bug

On a Mac, when I attempt to create a new file or open a daily note in the .dev version of Foam, I get an error message: No file system handle registered. I am using v0.26.1 of the extension. 

Errors: 

Command 'Foam: Create New Note' resulted in an error
Unable to write file '/markschaver/foamtest/Test.md' (Unavailable (FileSystemError): Error: No file system handle registered (/markschaver))

Command 'Foam: Open Daily Note' resulted in an error
Unable to write file '/markschaver/foamtest/2024-10-09.md' (Unavailable (FileSystemError): Error: No file system handle registered (/markschaver))

### Small Reproducible Example

https://github.dev/markschaver/foamtest/blob/master/readme.md

### Steps to Reproduce the Bug or Issue

1. Open repository in github.dev by clicking .
2. Use Cmd-shift-p to create a new file or open a daily note.

### Expected behavior

I expect the files to be created. They are not. I am also unable to create new files by clicking on a wiki link.

### Operating System Version

macOS (though experienced same problem on Windows 11 with v26.)

### Visual Studio Code Version

Latest .dev version.

pderaaij:
Which browser are you using? And could you see if vscode.dev works?

markschaver:
Chrome. And yes, vscode.dev works otherwise, including other extensions. I have used vscode.dev for a long time. It does not work either in Safari, where I get this error message:

Command 'Foam: Create New Note' resulted in an error
command 'foam-vscode.create-note' not found

pderaaij:
During my development work on this I only tested this on local folders. Doing some research it seems we need to add some capabilities to support remote repositories. See: https://code.visualstudio.com/blogs/2021/06/10/remote-repositories#_ensuring-your-extension-works-in-a-virtual-workspace

I will have a look at this in the upcoming weeks.

pderaaij:
I've made some progress in this area. Will document my findings here. (@riccardoferretti FYI).

Ultimately, a re mote repository works as a virtual file system. This differs to the normal behaviour of Foam which in some areas looks at absolute paths or simply core functions that work differently. This requires us to handle some paths differently. 

For example, the behaviour of `workspace.findFiles()` is different for local systems or virtual systems. For my test, I need to change the implementation as following:

```ts
const uris = await workspace.findFiles(
        folder.uri.scheme === 'vscode-vfs'
          ? '**/*'
[...]

PR #1409, opened 2024-11-08
Improved URI handling for virtual FS

riccardoferretti:
Fixes https://github.com/foambubble/foam/issues/1404

Based on what I mentioned in the ticket, the approach I went for was to never leave the "uri space" of the workspace folder.
This fixes the loading of the resources, the graph and the creation of notes in my environment.
Can you please check it makes sense to you and it works on your system as well?


pderaaij:
Looks good to me. Better solution indeed.

I only needed to add the `{ language: 'markdown', scheme: 'vscode-vfs' }` config to the mdDocSelector. Otherwise the hover completion doesn't work. But, with this change. It worked perfectly.

pderaaij (approved):

PR #1455, opened 2025-04-11
Fix parsing of tags when colon exists in other properties

daniellowtw:
The parsing of tags relies splitting on colon which does not work when other properties has colon. E.g. having a date property that includes time.

This commit fixes it by rewriting the logic to get tags -> line number.

Previously, having frontmatter like 
```
---
title: 2025 04 11
date: 2025-04-11T00:01:00+01:00
tags: 
    - new
---
```
causes the parser to silently error out and therefore doesn't parse the tags defined in the yaml at all.

### DIFF
--- a/packages/foam-vscode/src/core/services/markdown-parser.test.ts
+++ b/packages/foam-vscode/src/core/services/markdown-parser.test.ts
@@ -320,6 +320,28 @@ this is some #text that includes #tags we #care-about.
       ]);
     });
 
+    it('can find tags when other fields has colon', () => {
+      // This is a regression test. Previously a colon in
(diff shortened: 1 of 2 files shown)


riccardoferretti:
Thanks @daniellowtw - given the work in #1482 which completely changes the way we parse markdown, I will wait to merge this PR. Maybe @ryanncode can add your test case to the PR to make sure colons are properly handled

daniellowtw:
Thanks. I look forward to having this fixed so I don't have to run my own branch anymore :)

riccardoferretti:
This should be now fixed in 0.32.0 - let me know if you have any problem

PR #1456, opened 2025-04-16
support tag graph

Tenormis:
Add support for 'Tag Relationship Graph' on the basis of 'Hierarchical Tags'

### DIFF
--- a/packages/foam-vscode/static/dataviz/graph.js
+++ b/packages/foam-vscode/static/dataviz/graph.js
@@ -251,23 +251,32 @@ function initDataviz(channel) {
 }
 
 function augmentGraphInfo(graph) {
+  const tagNodes = {};
   Object.values(graph.nodeInfo).forEach(node => {
+    if (node.type === 'tag') {
+      tagNodes[node.title] = node;
+    }
     node.neighbors = [];
     node.links = [];
+  });
+  Object.values(graph.nodeInfo).forEach(node => {
     if (node.tags && node.tags.length > 0) {
       node.tags.forEach(tag => {
         subtags = tag.label.split('/');
         for (let i = 0; i < subtags.length; i++) {
           const label = subtags.slice(0, i + 1).join('/');
-          const tagNode = {
-            id: label,
-            title: label,
-            type: 'tag',
-            properties: {},
-            neighbors: [],
-            links: [],
-          };
-          graph.nodeInfo[tagNode.id] = tagNode;
+          if (!tagNodes[label]) {
+            const tagNode = {
+              id: label,
+              title: label,
+              type: 'tag',
+              properties: {},
+              neighbors: [],
+              links: [],
+            };
+            graph.nodeInfo[tagNode.id] = tagNode;
+            tagNodes[tagNode.id] = tagNode;
+          }
           if (i > 0) {
             const parent = subtags.slice(0, i).join('/');
             graph.links.push({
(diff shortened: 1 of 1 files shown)


Tenormis:
@riccardoferretti Now both ‘Hierarchical Tags’ and ‘Tag Relationship Graph’ are supported.

> https://github.com/foambubble/foam/discussions/1442#discussion-8119317

riccardoferretti (commented):
Thanks for the PR @Tenormis 
I am not convinced by the idea of a "tag note", so I won't be merging this until I get more conviction around that.
Also I feel introducing this concept this way might not be the best approach, as e.g. if you have a tag note I guess you'd want to see it in the tag explorer, and this PR wouldn't do that.
Basically, if this were to be done, it needs to be done holistically, and I don't believe it's as trivial as it might sound.

Also, I am not sure I believe in the solution in the first place, partially because I am not super familiar with the problem - and I don't want to end up having to support/navigate a feature that I don't believe in.

Tenormis:
@riccardoferretti 

It's actually not a "tag," nor even a "note." Perhaps it's more aptly called "tag-rel" (tag relationships).

It utilizes markdown metadata to declare the relationships between tags within the metadata. For example, when you mention web, you might associate it with javascript, html, css, and so on. In various notes, there might be tags like javascript, css, etc., but you usually wouldn't add an additional web tag to the notes. The "tag-rel" provides a way to establish connections between tags like javascript, html, and css.

So, this should answer your first question, "if you have a tag note I guess you'd want to see it in the tag explorer" — I don't want to see "tag-rel" in the tag explorer, nor do I want "tag-rel" to appear as a node in the graph.

Regarding your second question, "I don't believe it's as trivial as it might sound" — it really is that simple, not as complicated as you might imagine.

Moreover, this submission has no impact/side effects on those who do not need to use this feature.

Looking forward to your reply.

Tenormis:
Certainly, the code implementation here is a bit of a hack — mainly the `tagNodes` variable (which is a set/map object), it keeps track of all "tag notes" (notes of `type: tag`, I will explain the reason for this design later).

When creating nodes for all tags, it first checks whether there is an existing "tag note" with the same name in `tagNodes`. If it exists, then the "tag note" will replace the tag node and fulfill its purpose:

1. Use the existing "tag note" as the tag node, without creating another tag node;
2. Let other notes that declare this tag connect to the "tag node";

Finally, here's the reason for this design:

1. In the graph, you won't see two nodes with the same name — the "tag note" and the "tag node";
2. The style of the "tag note" will remain consistent with the "tag node";

riccardoferretti:
> It's actually not a "tag," nor even a "note." Perhaps it's more aptly called "tag-rel" (tag relationships).

This is exactly what I am referring to. It's a new concept and before introducing a new concept I want to be super sure that it's necessary, there is no good workaround, and that the use case is common enough to justify the effort to add it and (more importantly) to maintain it.

> I don't want to see "tag-rel" in the tag explorer.

I understand, but then some other user might say "hey, it's a tag related concept, I want to see it there". And it comes down to style, philosophy, and so on. 
Part of my job in Foam is to curate such personal preferences, sometimes saying no to possibly good ideas that I don't have a clear view on (what I am doing right now ;) )

> this submission has no impact/side effects on those who do not need to use this feature

It doesn't matter in this specific case, because this PR introduces a new concept.
The side effect is the introduction of a new concept.

Notice the difference with the font style you added in the other PR. That change is self-contained,  doesn't introduce new ideas and at worst someone might want to customize another aspect of the graph style. I can easily remove it without major consequences if it turns out to not be useful or a net benefit.
Here we are introducing something new, something that as you said is 'actually not a "tag," nor even a "note.' - not the type of change I want to introduce lightly just because in its first introduction the "submission has no impact/side effects".

I really appreciate the initiative and the submission of the PR. I encourage it and support people that want to contribute.
I need to balance this with the needs of the project, and my needs as maintainer (contributors come and go, as you can imagine).

I hope this clarifies and puts my decision in context.

Issue #1478, opened 2025-06-10
Foam uses different glob parser than VSCode, so setting files.exclude can crash Foam


s-jacob-powell:
### Describe the bug

Setting `files.exclude` in VSCode settings like this:
```
    "files.exclude": {
      "**/.git": true,
      "**/.svn": true,
      "**/.hg": true,
      "**/.DS_Store": true,
      "**/Thumbs.db": true,
    }
```
Works:
```
Foam Logging: info
[info - 4:12:05 PM] Starting Foam
[info - 4:12:25 PM] Loading from directories:
[info - 4:12:25 PM] - /home/jacob/dev/project
[info - 4:12:25 PM]   Include: **/*
[info - 4:12:25 PM]   Exclude: **/.foam/**,**/*.code-workspace,**/.git,**/.svn,**/.hg,**/.DS_Store,**/Thumbs.db,**/.*cache*,**/_*cache*
[info - 4:12:25 PM] - /home/jacob/dev/project2
[info - 4:12:25 PM]   Include: **/*
[info - 4:12:25 PM]   Exclude: **/.foam/**,**/*.code-workspace,**/.git,**/.svn,**/.hg,**/.DS_Store,**/Thumbs.db,**/.*cache*,**/_*cache*
[info - 4:12:25 PM] - /home/jacob/dev/project3
[info - 4:12:25 PM]   Include: **/*
[info - 4:12:25 PM]   Exclude: **/.foam/**,**/*.code-workspace,**/.git,**/.svn,**/.hg,**/.DS_Store,**/Thumbs.db,**/.*cache*,**/_*cache*
[info - 4:12:37 PM] Workspace loaded in 11860ms
[info - 4:12:37 PM] Graph loaded in 326ms
[info - 4:12:37 PM] Tags loaded in 83ms
[info - 4:12:37 PM] Loaded 3711 resources
```

---

However, setting something like this:
```
    "files.exclude": {
      "**/.git": true,
      "**/.svn": true,
      "**/.hg": true,
      "**/.DS_Store": true,
      "**/Thumbs.db": true,
      "**/subfolder/{A,aa,ac,ag,ak,ap,arc,at,aw,b,ce,cl,cn,co,cr,cx,cu,d,e,f,g,im,in,j,k,le,lm,lo,m,n,o,p,q,r,s,t,v,w}*": true
    }
```
Causes Foam to not start (although, setting this after startup does not crash foam):
```
[info - 4:19:20 PM] Starting Foam
[...]

PR #1479, opened 2025-06-13
Add expandAlternateGroups to support ignoredFiles globs like "**/ignore{1,2}.txt"

s-jacob-powell:
Fixes https://github.com/foambubble/foam/issues/1478

VScode setting `files.exclude` supports glob alternates like `file{1,2}.txt`. This excludes files `file1.txt` and `file2.txt`.

On startup, Foam will use this exclude setting along with some others (`settings.ts`). This causes Foam to crash on startup due to not supporting the alternates syntax `{ }`. At worst it should simply warn instead of crash, but it's not too hard to support the syntax by expanding the alternates in `{ }`.

I added this file and function to utils, and a bunch of tests to showcase what it does.
I'm happy to simplify anything as needed, for example right now it is recursive, so supports multiple and nested alternates, which might be too much.

I am also not sure if files can have `{` or `}` in them, but I don't know if we would want to support that anyway, like supporting escaping them or something. We can add that if we feel like we need that.

I am not super TS/JS oriented so hopefully everything looks ok; I used copilot for some help in especially the tests and the regex. 

I tested it in an example workspace where it errored without this change, and then of course ensured the unit tests I added pass. I'm also kind of unseasoned in contributing to open source so hopefully this is workable. Let me know if I really need anything else.

Thanks, I ❤️ Foam!


riccardoferretti:
@allcontributors add @s-jacob-powell for code

s-jacob-powell:
Wow awesome! Hopefully I can add more as I continue to use foam everyday!

My only question is how does the release schedule/process work, when will I/can I already use the latest extension release? Maybe there's like pre-release versions or something?

riccardoferretti:
I will do a release between today and tomorrow :)

PR #1480, opened 2025-06-13
Sync links on directory move/rename

### LINKED ISSUE #1143: Links do not sync after renaming folder in file explorer
### Describe the bug

Foam correctly syncs links when a folder name is changed using f2 on a wikilink or when dragging a file to a different folder in the file explorer, however, if a folder is renamed by pressing f2 on the folder in the file explorer, Foam does not update links to files in the directory.

_No response_

1. Great a Foam workspace with a file in the root directory and a file in a subdirectory.
2. In the file in the root directory put a link to the file in the subdirectory.
3. In the file explorer select the subdirectory and press f2 to change the subdirectory name
4. The link in the file in the root directory will now be broken. 

Problem also occurs when using reference links.

### Expected behavior

Links should sync when renaming directory in file explorer.

### Screenshots or Videos

_No response_

### Operating System Version

Windows 10, Windows 11

1.75-insider

### Additional context

_No response_

Alb-O:
This PR implements link integrity on folder renaming: when a folder is renamed via the VS Code explorer, intelligently update all affected Markdown links (wikilinks, standard markdown links, and relative paths).

*   User Confirmation Dialog:
    *   Displays a dry run summary of files and links to be updated before proceeding.
    *   Provides options "Update Links," "Skip Link Updates," "Abort Rename," and "Settings."
          *   "Settings" option takes the user to the relevant setting below:
*   Settings:
    *   `foam.folderRename.updateLinks`: Allows users to set their preference to "always" update, "never" update, or "confirm" each time.
    *   `foam.folderRename.maxFilesForSummary`: Configures the maximum number of files to list in the confirmation dialog summary.
*   Workspace Integrity:
    *   Refreshes the Foam workspace cache and URI matcher after a rename operation.
    *   Ensures that only affected files are saved, files dirtied previously by the user and that are not relevant to the operation are untouched.
*   Link updates are consolidated into a single `WorkspaceEdit` for performance.

Fixes #1143


packages/foam-vscode/src/features/folder-rename-dialog.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Handles the user confirmation dialog for folder renames
+ */
+export class FolderRenameDialog {
+  
+  /**
+   * Helper function for intelligent pluralization
+   */
+  public static pluralize(count: number, singular: string, plural?: string): string {
```
riccardoferretti: I like this function, I would put it in `core/utils`

packages/foam-vscode/src/features/folder-rename.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+import * as vscode from 'vscode';
+
+describe('Folder rename with link updates', () => {
+  beforeAll(async () => {
+    await closeEditors();
+    await cleanWorkspace();
+  });
+  afterAll(closeEditors);
+
+  describe('folder rename operations', () => {
```
riccardoferretti, line 24: Can we add a test case for:
```
folder-a/
  note-a.md
  note-b.md
folder-b/
  note-b.md
````

with:
- `note-a` points to `folder-a/note-b`
- `folder-a` being renamed

File: packages/foam-vscode/src/features/folder-rename-dialog.ts [hunk 3]
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    if (selected.label.includes('Update Links') || selected.label.includes('Proceed')) {
+      return { action: 'proceed' };
+    } else if (selected.label.includes('Skip Link Updates')) {
+      return { action: 'skip' };
+    } else if (selected.label.includes('Settings')) {
+      return { action: 'settings' };
+    } else if (selected.label.includes('Abort Rename')) {
+      return { action: 'abort' };
```
riccardoferretti: Minor: have you consider adding an `action` field to the quickpick items, and when one is selected to return the value of the field? will spare you all the if/else cases here.

packages/foam-vscode/src/features/folder-rename-dialog.ts
```diff
@@ -0,0 +1,153 @@
...
+  edits: vscode.WorkspaceEdit;
+  linksUpdatedCount: number;
+  filesToEditCount: number; // Number of files with edits
+  filesProcessedCount: number; // Total markdown files scanned
+  calculationErrors: string[];
+}
+
+export interface FolderRenameDialogResult {
+  action: 'proceed' | 'skip' | 'cancel' | 'settings' | 'abort';
+}
```
riccardoferretti: Minor: is there a reason why this is an object instead of a string union (i.e. `type X = 'a' | 'b' | 'c'`)

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+        ) {
+          suppressedErrorCount++;
+          Logger.debug(
+            `Suppressed expected FileDataStore error for rename paths: ${fullMessage}`
+          );
+          return;
+        }
+      }
+      originalLoggerError(message, ...params);
+    };
```
riccardoferretti, line 96: Can you help me understand what is happening here?

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+export class FolderRenameHandler {
+  private foam: Foam;
+  constructor(foam: Foam) {
+    this.foam = foam;
+  }
+
+  /**
+   * Main entry point for handling folder rename events
+   */
+  async handleFolderRename(
```
riccardoferretti, line 46: this function is a bit long and does a few things, do you think it would be possible to break it down into sub-functions for the various steps?

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+    const folderPath = fromVsCodeUri(folderUri).path;
+
+    // Get all resources from Foam workspace that are in the renamed folder
+    const markdownFilesInFolder = this.foam.workspace
+      .list()
+      .filter(resource => {
+        const resourcePath = resource.uri.path;
+        return (
+          resourcePath.startsWith(folderPath + '/') ||
+          resourcePath === folderPath
```
riccardoferretti: I would suggest putting this as a utility fn in `URI`, for the following reasons:
- it's been used multiple times in this file
- I can see how it could be useful in other places
- the function name can describe semantically what you want to achieve
- you would then be working at the level of `URI` instead of path (so if you need to take into account OS-specific elements, you would put that in the URI abstraction, which is where it belongs)

Thoughts?

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+                  );
+                } else {
+                  Logger.debug(
+                    `  → Simple wikilink "${originalTarget}" has ${candidatesAfterRename.length} candidates, will update to avoid ambiguity`
+                  );
+                }
+              } else {
+                Logger.debug(
+                  `  → Path-based wikilink "${originalTarget}" will be updated for folder rename`
+                );
```
riccardoferretti, line 594: I am not sure I understand this section. If it's a simple wikilink, renaming the folder shouldn't affect the link (it was `[[filename]]` before and it will be `[[filename]]` after.
In the case where there is more than one candidate, I don't think you have enough information to know which candidate should be kept, which I guess means we would still want to leave the original wikilink.
Am I missing or misunderstanding something?

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+                    vscode.workspace.getWorkspaceFolder(newFileUri);
+                  if (wsFolder) {
+                    const rel = path
+                      .relative(wsFolder.uri.fsPath, newFileUri.fsPath)
+                      .replace(/\\/g, '/');
+                    newLinkContent = rel.replace(path.extname(rel), '');
+                  } else {
+                    newLinkContent =
+                      this.foam.workspace.getIdentifier(targetFoamUri);
+                  }
```
riccardoferretti, line 619: I wonder if there is a way where you can work at the level of the `FoamWorkspace` instead of having to manipulate paths and extensions. 
E.g. if (this is off the top of my head, not thought through) you "clone" the Foam workspace object, update the resource objects in there (basically, simulating the renaming was completed), and the using `getIdentifier` to retrieve the new identifier

The good thing about that approach is that it should be pretty standard, but the devil might be in the details (including performance)

What do you think?

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+        for (const outgoingConnection of outgoingConnections) {
+          const { target } = MarkdownLink.analyzeLink(outgoingConnection.link);
+
+          Logger.debug(
+            `Checking outgoing link: "${outgoingConnection.link.rawText}" (type: ${outgoingConnection.link.type}, target: "${target}")`
+          );
+
+          // Process relative links that start with ../ for both regular links and wikilinks
+          // But skip links that point to files within the same renamed folder (those are handled by backlink processing)
+          if (!isAbsolute(target) && target.startsWith('../')) {
```
riccardoferretti, line 697: Good point, I hadn't thought about this case - we should probably add a test case for this if you haven't already

(feels like a candidate for its own function)

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+   * Calculates an adjusted relative path when a file is moved to a different folder depth.
+   * This is needed for relative links in moved files that point outside the moved folder.
+   *
+   * @param originalRelativePath The original relative path (e.g., "../../other.md")
+   * @param oldFileUri The original file URI
+   * @param newFileUri The new file URI after the folder move
+   * @param workspaceRootPath The workspace root path for resolving absolute paths
+   * @returns The adjusted relative path or null if adjustment is not needed/possible
+   */
+  private calculateAdjustedRelativePath(
```
riccardoferretti, line 826: would be good to have some testing about this function

packages/foam-vscode/src/features/folder-rename-handler.ts
```diff
@@ -0,0 +1,1112 @@
...
+          );
+          // Open VS Code settings and search for the foam folder rename setting
+          await vscode.commands.executeCommand(
+            'workbench.action.openSettings',
+            'foam.folderRename.updateLinks'
+          );
+
+          // Abort the current rename operation by reverting the folder
+          try {
+            await this.revertFolderRename(newUri, oldUri);
```
riccardoferretti, line 268: why do you need to revert the folder rename? isn't this called _before_ the renaming happens? (`onWillRenameFiles`)

riccardoferretti (commented):
Thanks for taking on such a big chunk of work!
I have left a few comments, let me know what you think

Alb-O:
> Thanks for taking on such a big chunk of work! I have left a few comments, let me know what you think

No worries, I'll get to your comments when I can, thanks for the feedback.

packages/foam-vscode/src/features/folder-rename-dialog.ts
Alb-O: Done, made a `utils/string.ts` file.

packages/foam-vscode/src/features/folder-rename-handler.ts
Alb-O: Agreed, done

packages/foam-vscode/src/features/folder-rename-dialog.ts
Alb-O: Agreed, done

packages/foam-vscode/src/features/folder-rename-dialog.ts
Alb-O: Nope, you're right, fixed

PR #1482, opened 2025-06-24
Block Level Identifiers

ryanncode:
This feature adds ^abc123 style block level identifiers. We can use block identifiers generally in every way that you could use a header section for backlinks and transclusions, like ![[doc#^block-id]]. The block identifier must be placed either in-line at the end of a paragraph or list item, or on its own line following a markdown node (full list, block quote, code block, table). 

I have also found this was previously requested in this issue:
https://github.com/foambubble/foam/issues/294

Some conditions we are faced with:
- In Markdown Preview, block id backlinks link to their nearest header section. Across many attempts, it has been found that VS Code sandboxed default Markdown Preview window is heavily restricted from setting custom html id tags or querying them. There may be a solution I'm not familiar with. In every other case including editor view, hover view, transclusion, autocomplete, etc., we link directly to the block id as needed. To me this seems like a fair tradeoff, since the main interest would be block identifier transclusions.
- This is not compatible with Prettier since it is very opinionated about how Markdown nodes must be ordered, and will not allow the block id to be placed on the last line. My suggestion would be to pair Markdown All in One with Markdown Linter, which is trade off between strict automation of the layout vs more user control over which errors they want to catch
- For simplicity we choose to only recognize the last block id on the line, it could be improved to allow multiple block ids on the same line (to be linked to for separate contexts)

Other notes:
- The parser has been reworked to be AST-based instead of line # and indentation level heuristic based, which should be an overall improvement in performance. This could use some testing in large documents or workspaces with many documents and so on
- Since this does require heavy changes, I can follow up until we confirm every part of the feature is well tested. I did add many tests to start off with, and they should be currently passing.


packages/foam-vscode/package.json
```diff
@@ -680,6 +680,7 @@
     "test": "yarn test-setup && node ./out/test/run-tests.js",
     "test:unit": "yarn test-setup && node ./out/test/run-tests.js --unit",
     "test:e2e": "yarn test-setup && node ./out/test/run-tests.js --e2e",
+    "test:tdd": "yarn build:node && jest --runInBand",
```
riccardoferretti: What does this allow?

packages/foam-vscode/src/core/model/graph.ts
```diff
@@ -28,6 +29,10 @@ export class FoamGraph implements IDisposable {
    * Maps the connections arriving to a URI
    */
   public readonly backlinks: Map<string, Connection[]> = new Map();
+  /**
+   * Maps the block identifiers to the notes that contain them
+   */
+  public readonly blockBacklinks: Map<string, Set<URI>> = new Map();
```
riccardoferretti: was there a reason to not leverage `backlinks`, and `Connection`? It would be nice to have a consistent API on these

packages/foam-vscode/tsconfig.json
```diff
@@ -3,11 +3,13 @@
...
     "moduleResolution": "node",
     "esModuleInterop": true,
+    "allowJs": true,
     "outDir": "out",
     "lib": ["ES2019", "es2020.string", "DOM"],
     "sourceMap": true,
     "strict": false,
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "module": "CommonJS"
```
riccardoferretti: why did you need to change this?

packages/foam-vscode/tsconfig.json
```diff
@@ -3,11 +3,13 @@
   "compilerOptions": {
     "moduleResolution": "node",
     "esModuleInterop": true,
+    "allowJs": true,
```
riccardoferretti: do we really want that? :)

packages/foam-vscode/src/core/model/note.ts
```diff
@@ -39,8 +40,11 @@ export interface Alias {
 }
 
 export interface Section {
+  id?: string; // A unique identifier for the section within the note.
   label: string;
   range: Range;
+  blockId?: string; // The optional block identifier, if one exists (e.g., '^my-id').
+  isHeading?: boolean; // A boolean flag to clearly distinguish headings from other content blocks.
```
riccardoferretti: A few points/questions:
- it's an interesting idea "computing" the `id` of a section at parsing time, so if we need to reference it afterwards we can easily do so, and it allows us to slightly "decouple" label from id - is this the reason why you introduced that?
- given `id` and `isHeading`, do we have enough to identify a block? Isn't `blockId === id` for a block Section?
- from glancing at where `isHeading` is used, do we need it explicitly or can we just check if the `id` starts with `^` as a convetion?
- also, `isHeading` refers to the anchor, but this object describes a section. I wonder if we should call this something like `isBlock` (in which case we are inverting the semantics though) or something like `isHeaderSection` (which I think it's a bit uglier but "compatible" with the existing PR)

packages/foam-vscode/src/core/model/note.ts
```diff
@@ -9,6 +9,7 @@ export interface ResourceLink {
 }
 
 export interface NoteLinkDefinition {
+  type?: string; // 'block' for block identifiers
```
riccardoferretti: I don't think I am seeing this being used, probably I am missing it, can you point me to usage?

packages/foam-vscode/src/core/services/markdown-blockid-html-plugin.ts
```diff
@@ -0,0 +1,74 @@
+import MarkdownIt from 'markdown-it';
```
riccardoferretti: this file should probably live in `features/preview`

packages/foam-vscode/src/core/services/markdown-parser.ts
```diff
@@ -399,107 +402,430 @@ const definitionsPlugin: ParserPlugin = {
...
-    // have exited the trailing definition block, and should bail
-    const start = def.range!.start.line;
-    if (start < previousLine - 2) {
-      break;
-    }
+  return {
+    name: 'block-id',
+    onWillVisitTree: () => {
+      processedNodes.clear();
+      slugger.reset();
```
riccardoferretti: why are you resetting the slugger?

packages/foam-vscode/src/core/services/markdown-parser.ts
```diff
@@ -399,107 +402,430 @@ const definitionsPlugin: ParserPlugin = {
...
-    previousLine = def.range!.end.line;
-  }
+      // NEW: Special Case for Full-Line Block IDs on Lists
+      if (node.type === 'list') {
+        const listText = getNodeText(node, markdown);
+        const listLines = listText.split('\n');
+        const lastLine = listLines[listLines.length - 1];
+        const fullLineBlockId = getLastBlockId(lastLine.trim());
+
+        if (fullLineBlockId && /^\s*(\^[\w.-]+\s*)+$/.test(lastLine.trim())) {
```
riccardoferretti: can you put the regex in a variable so it's clearer from the name what it is?

packages/foam-vscode/src/core/services/markdown-parser.ts
```diff
@@ -399,107 +402,430 @@ const definitionsPlugin: ParserPlugin = {
...
+          ) {
+            block = previousSibling;
+            blockId = fullLineBlockId;
+            idNode = node; // This paragraph is the ID node
+          } else {
+            // If it's a full-line ID paragraph but not correctly linked,
+            // mark it as processed so it doesn't get picked up as an inline ID later.
+            processedNodes.add(node);
+            return; // Skip further processing for this node
+          }
```
riccardoferretti: I am not gonna pretend I am fully understanding this bit, maybe on second/deeper review

packages/foam-vscode/src/core/services/markdown-parser.ts
```diff
@@ -399,107 +402,430 @@ const definitionsPlugin: ParserPlugin = {
...
+ *
+ * This plugin adheres to the following principles:
+ * - Single-pass AST traversal with direct sibling analysis (using `unist-util-visit-parents`).
+ * - Distinguishes between full-line and inline IDs.
+ * - Applies the "Last One Wins" rule for multiple IDs on a line.
+ * - Prevents duplicate processing of nodes using a `processedNodes` Set.
+ *
+ * @returns A `ParserPlugin` that processes block identifiers.
+ */
+export const createBlockIdPlugin = (): ParserPlugin => {
```
riccardoferretti, line 433: why are you using this factory function? (as opposed to a simple object)
is it to keep the helper fns private?

packages/foam-vscode/src/core/services/markdown-parser.ts
```diff
@@ -399,107 +402,430 @@ const definitionsPlugin: ParserPlugin = {
...
+  };
 
-function getFoamDefinitions(
-  defs: NoteLinkDefinition[],
-  fileEndPoint: Position
-): NoteLinkDefinition[] {
-  let previousLine = fileEndPoint.line;
-  const foamDefinitions = [];
+  // Gets the raw text of a node from the source markdown
+  const getNodeText = (node: Node, markdown: string): string => {
```
riccardoferretti: this could be helpful as a generic parsing utility fn right?

packages/foam-vscode/src/core/services/markdown-provider.ts
```diff
@@ -35,7 +35,7 @@ export class MarkdownResourceProvider implements ResourceProvider {
       if (isSome(section)) {
         const rows = content.split('\n');
         content = rows
-          .slice(section.range.start.line, section.range.end.line)
+          .slice(section.range.start.line, section.range.end.line + 1)
```
riccardoferretti, line 38: is this addressing an existing bug or is some semantic changing here?

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -101,11 +146,54 @@ export class HoverProvider implements vscode.HoverProvider {
...
+      if (isSome(content)) {
+        // Using vscode.MarkdownString allows for rich content rendering in the hover.
+        // Setting `isTrusted` to true is necessary to enable command links within the hover.
+        const markdownString = new vscode.MarkdownString(content);
+        markdownString.isTrusted = true;
+        mdContent = markdownString;
+      } else {
+        // If no content can be loaded, fall back to displaying the note's title.
+        mdContent = targetResource.title;
+      }
```
riccardoferretti: why are you not using `getNoteTooltip` here?
it adds some extra checks (e.g. handling YAML), and also takes care of using `MarkdownString`

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -119,17 +130,86 @@ export class SectionCompletionProvider
...
-        item.sortText = String(b.range.start.line).padStart(5, '0');
-        item.range = replacementRange;
-        item.commitCharacters = sectionCommitCharacters;
-        item.command = COMPLETION_CURSOR_MOVE;
-        return item;
+      });
+      // Provide completion for all sections: headings, block IDs (including list items), and header IDs
+      const items = resource.sections.flatMap(section => {
+        const sectionItems: vscode.CompletionItem[] = [];
+        if (section.isHeading) {
```
riccardoferretti: this is why I am a bit hesitant about the `isHeading` and treating blocks differently from sections - basically everywhere you need to handle the two cases. is there a way to consolidate them?

packages/foam-vscode/src/features/panels/utils/tree-view-utils.ts
```diff
@@ -188,24 +188,58 @@ export const groupRangesByResource = async (
...
 
+/**
+ * Creates backlink items for a resource, optionally scoped to a section/block (by fragment).
+ * If fragment is provided, only backlinks to that section/block are included.
+ */
 export function createBacklinkItemsForResource(
   workspace: FoamWorkspace,
   graph: FoamGraph,
   uri: URI,
+  fragment?: string,
```
riccardoferretti: the `uri` already has a fragment attribute, what is the extra parameter for?

packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
```diff
@@ -121,21 +199,21 @@ This is the third section of note E
 
     await withModifiedFoamConfiguration(
       CONFIG_EMBED_NOTE_TYPE,
-      'full-card',
+      'full-inline',
```
riccardoferretti: why changing this element of the test?

packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
```diff
@@ -121,21 +199,21 @@ This is the third section of note E
...
       () => {
         const md = markdownItWikilinkEmbed(MarkdownIt(), ws, parser);
 
         const res = md.render(
           `This is the root node. ![[note-e-container#Section 3]]`
         );
         expect(res).toContain('This is the root node');
-        expect(res).toContain('embed-container-note');
+        expect(res).not.toContain('embed-container-note');
```
riccardoferretti: same here

packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
```diff
@@ -411,7 +480,195 @@ content-card![[note-e#Section 2]]`);
       }
     );
 
-    await deleteFile(noteA);
-    await deleteFile(noteB);
+    await deleteFile(noteA.uri);
+    await deleteFile(noteB.uri);
+  });
+
+  describe('Block Identifiers', () => {
```
riccardoferretti, line 484: I like the new tests for the block identifiers. shouldn't all the other tests "just work" as I would assume the change to backwards compatible

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -112,35 +129,23 @@ function getNoteContent(
       const { noteScope, noteStyle } = retrieveNoteConfig(noteEmbedModifier);
 
       const extractor: EmbedNoteExtractor =
-        noteScope === 'full'
-          ? fullExtractor
-          : noteScope === 'content'
-          ? contentExtractor
-          : fullExtractor;
+        noteScope === 'content' ? contentExtractor : fullExtractor;
```
riccardoferretti, line 136: good point lol

packages/foam-vscode/src/features/wikilink-diagnostics.ts
```diff
@@ -114,141 +139,260 @@ export default async function activate(
...
-            });
-          }
-        }
-      }
+  const resource = parser.parse(
+    fromVsCodeUri(document.uri),
+    document.getText()
+  );
+
+  const diagnostics = resource.links.flatMap(link => {
```
riccardoferretti, line 166: interesting approach as opposed to a filter/reduce

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -208,22 +217,53 @@ export function retrieveNoteConfig(explicitModifier: string | undefined): {
...
+      // extract the content precisely using the range from the parser.
+      const rows = noteText.split('\n');
+      noteText = rows
+        .slice(section.range.start.line, section.range.end.line + 1)
+        .join('\n');
+    }
+  } else {
+    // No fragment: transclude the whole note (excluding frontmatter if present)
+    // Remove YAML frontmatter if present
+    noteText = noteText.replace(/^---[\s\S]*?---\s*/, '');
```
riccardoferretti: minor: this could be in `utils/md.ts` as it might be helpful somewhere else

riccardoferretti (commented):
First of all, thanks for taking on this big task!
I am still getting myself familiar with the PR, so my comments are meant as a conversation at this stage more than just feedback.
I have left a few comments, questions and suggestions.
I admit the PR is massive, if there were also a way to breaking it down into smaller ones it would make it easier to review.
Finally, can you tell me what AI support you used if any?
Thanks again!

packages/foam-vscode/src/core/services/markdown-parser.ts
ryanncode: Factory function use in markdown-parser.ts is definitely to keep the state and helper functions private to that plugin.

packages/foam-vscode/src/core/services/markdown-provider.ts
ryanncode: We add + 1 to the range slice in markdown-provider.ts to avoid needing to do this in multiple places later on, avoiding the outcome of missing a place where that was needed to be done. JS slice includes the start index but excludes the end index, even though we want returned content to represent the full section as defined by the range.

packages/foam-vscode/package.json
ryanncode: The new test:tdd command allows individual testing of any parsing focused test.ts file, which was mainly used for heavy testing of the block id parsing test file. For now I will rename it better unless it's not wanted.

packages/foam-vscode/src/core/services/markdown-provider.ts
ryanncode: We add + 1 to the range slice in markdown-provider.ts to avoid needing to do this in multiple places later on, preventing the outcome of missing a place where that was needed to be done. JS slice includes the start index but excludes the end index, even though we want returned content to represent the full section as defined by the range.

packages/foam-vscode/src/features/preview/wikilink-embed.ts
ryanncode: I have used stripFrontMatter instead, it seems to be usable on any string

packages/foam-vscode/src/core/services/markdown-parser.ts
ryanncode: Overall we are finding block ids and creating sections for them, but we now do this by walking through the AST. Along with sectionsPlugin, we perform a single, unified traversal of the document after initial parsing is complete. We call each plugin's visit function at every node.
- block holds the AST node that the ID applies to
- idNode tracks the node that contains a full line id, it might be separate depending on AST oddities for each different markdown node type
- nodeText extracts plain text of the node being visited

We first check for full line block IDs which happens with the regex on line 505. Here we do check for one or more block IDs on the line, although we only use the last one. On lines 521-523, if the conditions are met we link the ID to the previousSibling node. This is where block gets set to the node before the ID, and blockId gets set to the ID found on the line.

On line 538 we begin checking for inline block IDs, it only runs if the full-line ID was not found. On line 547 we handle list items by only forming a list item with its nested subitems block if we find the blockId on the parent list item, we then mark the children list items as passed.

On line 563 we create the section if we find a block and blockId, and skip heading sections which are handled by sectionsPlugin. On line 571 we have a switch block for the variety of markdown node types, and I have to say here again that the AST has extremely specific differences about the way each markdown node type gets handled. We then calculate the range, if the block has a last line ID it gets excluded.

packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
ryanncode: I had my personal setting to inline and forgot about the other styles as user settings, fixed.

packages/foam-vscode/src/features/preview/wikilink-embed.spec.ts
ryanncode: The primary changes in expected test data in this file are a result of improved transclusion parsing to remove incorrect double paragraph elements. There is also updated expected test data in markdown-parser.test.ts that is a result of improving getBlockFor range selection accuracy.

ryanncode:
> First of all, thanks for taking on this big task! I am still getting myself familiar with the PR, so my comments are meant as a conversation at this stage more than just feedback. I have left a few comments, questions and suggestions. I admit the PR is massive, if there were also a way to breaking it down into smaller ones it would make it easier to review. Finally, can you tell me what AI support you used if any? Thanks again!

I'm not sure about splitting the PR, the main thing is that the AST parser was introduced not just for performance but that I couldn't get the line and indent level heuristics to cooperate. The rest is pretty tightly coupled after making that choice. I can definitely give explanation for specific changes that were made to each file.

- Primary AI dev tools are RooCode for architecture planning and task orchestration, and Copilot for smaller editing tasks.

- Removed a few leftover artifacts from dev process, for example stuff related to trying to use javascript injections that I ended up not using.

packages/foam-vscode/src/core/model/note.ts
ryanncode: I had these questions also about the simplicity of the data model.

We do extend Foam's heading section label for a different use for block ids instead of changing the section object. Always using the computed id instead of findSection would be a performance improvement, but I also wanted to maintain familiarity of the object and not change every existing reference to label across the project.

After some experiments, it has been found that id, blockId, and isHeading are each fairly essential additions to the section object. There is not a great way to get rid of any one of them without making downstream code on the consumers end more complicated.

We could rename isHeading to hasHeader, but that's ambiguous about containment. The meaning of isHeading for me, is that it is a markdown heading section as opposed to an html header or yaml frontmatter, and also as opposed to a block identifier section. isHeading as a value contained in the Section object has more clarity than checking for the ^ of the block identifier, which is a complicated string comparison.

We are generally forced to check for isHeading in many places on the consumer end. Heading sections and block id sections (and each kind of block id) simply have many different ways of displaying across different interfaces. It would not be a good idea to try to pack all of those displays into the Section object itself, the current separation of concerns across different renderers is the correct approach.

Let's say from this point that what the consumers really need more than anything is a unified section results list, and this is exactly what we construct among the collection of heading sections and block id sections.

blockId is used in places like link-completion.ts and wikilink-diagnostics.ts. Were it removed, we would have to check isHeading, and then reconstruct the string from the ^ and id each time it needed to be used in the consumers. It is a good addition to the Section object. Of course, having a blockId value is enough to confirm being a block id, which suggests the use of isHeading instead.

For interest, the Section object I was giving consideration to before I encountered a lot of problems with implementation was this:

> label: string;
> range: Range;
> canonicalId: string; // used when creating new links, slug for a heading, blockid with carat for a block
> linkableIds: string[]; // list of all valid ids that resolve to this section, including canonicalId, also could be used for multiple block ids on a line

edit: 

isHeading acts as a good reminder to need to act on its presence in the Section object. We might also make use of discriminated unions or explicit type fields for more strict enforcement.

id can be used for a heading and a heading block id to refer to the same section, like when we reconstruct heading block id backlink hrefs to point to the correct heading section.

packages/foam-vscode/src/core/utils/md.ts
```diff
@@ -1,4 +1,28 @@
 import matter from 'gray-matter';
+import { Position } from '../model/position'; // Add Position import to the top
+
+/**
+ * Gets the raw text of a node from the source markdown.
+ * @param node The AST node with position info.
+ * @param markdown The full markdown source string.
+ * @returns The raw text corresponding to the node.
+ */
+export function getNodeText(
```
riccardoferretti: this to me feels like should live with the parser. it feels quite specific to that logic (e.g. the shape of the `node` param)

packages/foam-vscode/src/features/preview/blockid-preview-removal.ts
```diff
@@ -0,0 +1,54 @@
...
+// Matches a block ID at the end of a block (e.g., "^my-block-id")
+const blockIdRegex = /\s*(\^[-_a-zA-Z0-9]+)\s*$/;
+
+/**
+ * Markdown-it plugin for Foam block IDs (inline ^block-id syntax).
+ *
+ * - Removes block IDs from the rendered text for all block types.
+ * - For paragraphs and list items, cleans the block ID from the text.
+ */
+export function markdownItblockIdRemoval(
```
riccardoferretti, line 13: interesting idea, i like it

packages/foam-vscode/src/core/model/note.ts
riccardoferretti: This is probably the part of the PR I want to pay the closest attention to, as the data model for me is extremely important, so bear with me as I am trying to fully understand these choices and their implications.

> We do extend Foam's heading section label for a different use for block ids instead of changing the section object. 

Can you please help me understand that? In which way are you changing that? 
At a high level, I agree that `label` and `id` are not necessarily the same. At the same time, right now `findSection` does a simple label comparison, and in the case of a block, I could see how it would simply be a section where the label is the ID of the block (e.g. `^block-53`, notice I am including the `^`).
Does it feel semantically incorrect to you?

> After some experiments, it has been found that id, blockId, and isHeading are each fairly essential additions to the section object. There is not a great way to get rid of any one of them without making downstream code on the consumers end more complicated

I would like to understand why.
I am asking for a simple purpose: I would like to avoid a sub-optimal change to the model that is driven by "practical" concerns like how many downstream refactorings it requires, because the refactoring is a one off - whereas a change to the core model will be carried over time (accumulates tech debt, and especially on the "public model", which is a sort of API, feels like a heavy burden)

> We could rename isHeading to hasHeader, but that's ambiguous about containment. The meaning of isHeading for me, is that it is a markdown heading section as opposed to an html header or yaml frontmatter, and also as opposed to a block identifier section. isHeading as a value contained in the Section object has more clarity than checking for the ^ of the block identifier, which is a complicated string comparison

What about having a `type` property, which could be `'section' | 'block'`. It's semantically closer to the object semantics ("what type of section is this?") and it's extensible shall we need new types in the future (I don't think we need, but it was interesting what you were saying about html/yaml sections)

> Heading sections and block id sections (and each kind of block id) simply have many different ways of displaying across different interfaces

In what way they are different? The only thing I can think of is that a section has a header whereas a block will not, beside that I don't see any difference. Am I missing something?

> Let's say from this point that what the consumers really need more than anything is a unified section results list, and this is exactly what we construct among the collection of heading sections and block id sections

I agree, this is why I am trying to keep them as "aligned" as possible. My goal would be that from the user's POV you can treat the two in (almost) the same way. This is why I am a bit hesitant about having e.g. custom fields (like `blockId`) that only really apply to only one type of section.

> id can be used for a heading and a heading block id to refer to the same section, like when we reconstruct heading block id backlink hrefs to point to the correct heading section.

Sorry I didn't understand this, can you add an example?

ryanncode:
I have added a unified section object according to the requested parameters. Note that this does not allow us to avoid forking code in multiple files for header sections and block ID sections, but it should now be more obvious that this check is required in new files we might want to add.

While working on this solution, I did catch a few more edge case problems that were not noticed previously. One of the main changes I needed to make was to give up trying to clean the transclusion duplicate paragraph tags (so I returned to the original Foam expected test data), it's technically invalid but it's pretty complicated to match the open and closing tags given different inputs and it renders the same either way.

> id can be used for a heading and a heading block id to refer to the same section, like when we reconstruct heading block id backlink hrefs to point to the correct heading section.
> Sorry I didn't understand this, can you add an example?

This occurs in wikilink-navigation.ts, lines 78-117.

For the header:

## Header Text ^header-id

We want both "Header Text" and "^header-id" to be separate usable sections for the user that point to the same header section.

We did additionally add a new section level parameter to the unified section object. This is primarily used in wikilink-embed.ts for determining the end of a section for correct transclusion.

Beyond this immediate requirement, making the level part of the data model opens up possibilities for future features and would be beneficial in other potential downstream consumers, even if they don't strictly require it now. These include:

* Outline and Tree Views: Any feature that displays a document's structure (like the "Outline" view in VS Code or a panel showing a note's headings) could use the level to render the hierarchy correctly with proper indentation, without having to re-parse the file. The current backlink tree view in tree-view-utils.ts doesn't do this, but it's a logical extension.

* Navigation Features: A "Go to Next/Previous Heading" command could be made more intelligent. Instead of just jumping to any heading, it could jump to the next heading of the same level, which is a common navigation pattern in structured documents.

* Refactoring Tools: A tool that refactors a document's headings (e.g., promoting or demoting all subheadings under a specific parent) would need the level information to perform its task correctly.

Lastly, we could consider writing a Pandoc lua script file that would allow publishing for Foam in general and with the new block ID feature. There are existing examples that would make the script pretty simple to write.

packages/foam-vscode/src/core/model/markdown-parser-block-id.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        {
+          id: 'my-heading',
+          blockId: '^heading-id',
+          type: 'heading',
+          label: 'My Heading',
+          level: 2, // Add level property
+          range: Range.create(1, 0, 2, 0),
+        },
+      ]);
+    });
```
riccardoferretti, line 46: This is interesting, very good example you picked here.
In my mind, this markdown should actually produce 2 "sections", one triggered by the heading, and one triggered by the block-id.
tbh I am not sure why anyone would write something like this, but being valid markdown I agree we should handle it.

Also, I am seeing that in the case of the heading, the `label` doesn't include the `block-id`, whereas in the test case above of the paragraph, the label includes it. I am wondering whether I like this difference in behavior, what od you think?

packages/foam-vscode/src/core/model/markdown-parser-block-id.test.ts
```diff
@@ -0,0 +1,372 @@
...
+          id: 'parent-id',
+          blockId: '^parent-id',
+          type: 'block',
+          label: `- Parent item ^parent-id
+  - Child item 1
+  - Child item 2`,
+          range: Range.create(1, 0, 3, 16),
+        },
+      ]);
+    });
```
riccardoferretti, line 205: So the `label` is really the content? I wonder if for lists this should just be the name of the item the block is declared on (maybe not including the block id itself?). In this test case it would be `Parent item`.
The `range` can be used to extract the content (as we would do for a heading section).

Notice a lot of my comments are about unifying as much as possible the semantics of block sections with heading sections.
In fact, in my view they are exactly the same thing, only defined by different mechanisms (which is why I wrote that comment about `# my section ^section-id` generating 2 sections. it's kinda having the same block with two anchors pointing at it)

riccardoferretti: (notice that I am not sure how I would "generate" the label for tables or code blocks, any ideas? maybe the de-slugified block id? just thinking out loud)

packages/foam-vscode/src/core/model/note.ts
```diff
@@ -38,11 +39,29 @@ export interface Alias {
...
+interface BaseSection {
+  id: string; // The stable, linkable identifier (slug or blockId w/o caret)
+  label: string; // The human-readable or raw markdown content for display/rendering
+  range: Range; // The location of the section in the document
 }
 
+// A section created from a markdown heading
+export interface HeadingSection extends BaseSection {
+  type: 'heading';
+  level: number;
```
riccardoferretti, line 52: I like the level property. In theory we could also use it for list items (maybe on a different number-space (e.g. starting at 10) then headers), which i think we can also easily get from the md parser, but I don't consider it a blocker for moving forward.
Actually we could even have it for paragraphs (which i guess would just be level `10` or `0` depending how we want to do the numbering for non-heading sections)

packages/foam-vscode/src/core/model/note.ts
```diff
@@ -38,11 +39,29 @@ export interface Alias {
...
+  blockId?: string; // A heading can ALSO have a block-id
```
riccardoferretti, line 53: echoing what i wrote above, if a heading has a block id, I wonder if we should consider that text to generate 2 sections, one of type `type: 'block'` and the other of type `type: 'section'`.

packages/foam-vscode/src/core/model/workspace.ts
```diff
@@ -100,8 +100,13 @@ export class FoamWorkspace implements IDisposable {
    * Returns the minimal identifier for the given resource
    *
    * @param forResource the resource to compute the identifier for
+   * @param section the section of the resource to link to (optional)
    */
-  public getIdentifier(forResource: URI, exclude?: URI[]): string {
+  public getIdentifier(
+    forResource: URI,
+    exclude?: URI[],
+    section?: string
```
riccardoferretti, line 108: why are you adding this `section` parameter? I am not seeing it being used anywhere

packages/foam-vscode/src/core/utils/links.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Parses a wikilink target into its note and fragment components.
+ * @param wikilinkTarget The full string target of the wikilink (e.g., 'my-note#my-heading').
+ * @returns An object containing the noteTarget and an optional fragment.
+ */
+export function parseWikilink(wikilinkTarget: string): {
```
riccardoferretti, line 6: does it overlap with `MarkdownLink.analyzeLink`? Or should we put this function in that file for co-location/

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -77,10 +106,26 @@ export class HoverProvider implements vscode.HoverProvider {
...
-    const sources = uniqWith(
-      this.graph
+
+    // --- Start of Block ID Feature Changes ---
+
+    // Extract the fragment (e.g., #my-header or #^my-block-id) from the link.
+    // This is crucial for handling links to specific sections or blocks within a note.
+    const { section: linkFragment } = MarkdownLink.analyzeLink(targetLink);
+
+    let backlinks: import('../core/model/graph').Connection[];
```
riccardoferretti, line 116: why this dynamic import?

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -101,11 +146,44 @@ export class HoverProvider implements vscode.HoverProvider {
...
+          linkFragment
+        );
+        if (isSome(section)) {
+          if (section.type === 'block') {
+            // For block IDs, show the block label (e.g., the list item or paragraph)
+            content = section.label;
+          } else if (section.type === 'heading') {
+            // For headings, show the content under the heading (sliceContent)
+            const noteText = await this.workspace.readAsMarkdown(targetFileUri);
+            content = sliceContent(noteText, section.range);
```
riccardoferretti, line 167: You should be able to consolidate these by using the `range` field, right?

```
const noteText = await this.workspace.readAsMarkdown(targetFileUri);
content = sliceContent(noteText, section.range);
```

In fact, even if the `label` turns out to contain the content, I think for consistency of the semantics of a `Section` object it would be better to consolidate into "We use the range to get the content of a section"

packages/foam-vscode/src/features/hover-provider.ts
```diff
@@ -101,11 +146,44 @@ export class HoverProvider implements vscode.HoverProvider {
...
+      } else {
+        // If there is no fragment, show the entire note content, minus frontmatter.
+        content = await this.workspace.readAsMarkdown(targetFileUri);
+      }
 
-      mdContent = isSome(content)
-        ? getNoteTooltip(content)
-        : this.workspace.get(targetUri).title;
+      if (isSome(content)) {
+        content = content.replace(/---[\s\S]*?---/, '').trim();
```
riccardoferretti, line 182: I don't think you need to replace the YAML here, as it's already done by `getNoteTooltip`

packages/foam-vscode/src/features/link-completion.ts
```diff
@@ -119,17 +130,68 @@ export class SectionCompletionProvider
...
-          resource.uri.with({ fragment: b.label })
-        );
-        item.sortText = String(b.range.start.line).padStart(5, '0');
-        item.range = replacementRange;
-        item.commitCharacters = sectionCommitCharacters;
-        item.command = COMPLETION_CURSOR_MOVE;
-        return item;
+      const items = resource.sections.flatMap(section => {
+        const sectionItems: vscode.CompletionItem[] = [];
+        switch (section.type) {
```
riccardoferretti, line 135: I do wonder whether we are actually converging to a consistent treatment of the two types of sections, especially if we don't do special treatment for headings with block ids

packages/foam-vscode/src/features/navigation-provider.ts
```diff
@@ -122,7 +122,7 @@ export class NavigationProvider
       ? section.range
       : Range.createFromPosition(Position.create(0, 0), Position.create(0, 0));
     const targetSelectionRange = section
-      ? section.range
+      ? (section as any).labelRange || section.range // Use labelRange for headings, fallback to full section range
```
riccardoferretti, line 125: not sure the special casing is necessary, but happy to go with how you feel works best

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -208,60 +224,143 @@ export function retrieveNoteConfig(explicitModifier: string | undefined): {
...
+    ? Resource.findSection(note, linkFragment)
+    : null;
+
   if (isSome(section)) {
-    const rows = noteText.split('\n');
-    noteText = rows
-      .slice(section.range.start.line, section.range.end.line)
-      .join('\n');
+    if (section.type === 'heading') {
+      // For headings, extract all content from that heading to the next.
```
riccardoferretti, line 251: don't you already get this from the `range` property? Why are you re-computing it?
the `range` property should also already take into account the right level

packages/foam-vscode/src/features/preview/wikilink-embed.ts
```diff
@@ -208,60 +224,143 @@ export function retrieveNoteConfig(explicitModifier: string | undefined): {
...
+  if (!linkFragment) {
+    // If no fragment is provided, default to the first section (usually the main title)
+    // to extract the content of the note, excluding the title.
     section = note.sections.length ? note.sections[0] : null;
   }
-  let rows = noteText.split('\n');
+
   if (isSome(section)) {
-    rows = rows.slice(section.range.start.line, section.range.end.line);
+    if (section.type === 'heading') {
```
riccardoferretti, line 318: same here, you could use the range and add `1` to the row in the start position

packages/foam-vscode/src/features/preview/wikilink-navigation.ts
```diff
@@ -11,64 +11,135 @@ import { isEmpty } from 'lodash';
...
+ *
+ * @param md The markdown-it instance.
+ * @param workspace The Foam workspace to resolve links against.
+ * @param options Optional configuration.
+ */
 export const markdownItWikilinkNavigation = (
   md: markdownit,
-  workspace: FoamWorkspace
+  workspace: FoamWorkspace,
+  options?: { root?: vscode.Uri }
```
riccardoferretti, line 26: I can't find where this is being used and where it's being set?

riccardoferretti (commented):
I am really loving the direction this is going, and amazing job on the testing!
I have left various comments, looking forward to your thoughts

tctlrd:
Incredible work on this! I am excited to see this merged.

riccardoferretti:
hi @ryanncode - just wanted to check in and see your thoughts on the PR, thanks!

TomLucidor:
This is starting to vibe like LogSeq and SiYuan ngl

riccardoferretti:
@TomLucidor not sure I understand, can you elaborate?

TomLucidor:
Obsidian is traditionally a page-based PKM (link by file so it is easier to code), while LogSeq and SiYuan are block-based PKM (and linking is more modular with outline-like structure). The way they design how the latter works is kinda weird, but often useful when people cannot stand long articles.

ryanncode:
> hi @ryanncode - just wanted to check in and see your thoughts on the PR, thanks!

Trying to get time to get back on working on this. The first problem I encountered is that I can't fulfill both of your first two requests. I spent a good deal of time working on the first request, only to find that it conflicts with the more important requirements of the second request. In general, I don't see a way to get the VS Code interface to cooperate with a perfectly logical and semantic structure to the code. The best I can do is to clean up at least some of the output on the consumers end. Will submit something new within 1-2 weeks.

riccardoferretti:
Sorry, just to be clear I understand, can you tell me what the first and second requests you are referring to are? 
And in which way they conflict?

riccardoferretti:
@ryanncode I am closing this as in the meantime the code has diverged, and I haven't had news on the development of the branch. 
Thanks for starting the effort, I have now created another branch and this feature will soon be released, and some ideas and checks are coming from the work you did.

ryanncode:
@riccardoferretti That's good to hear. Sorry for bailing on the project, I ended up deciding to apply to graduate programs last fall, that started taking up all of my free time, and the process is still ongoing now.

I do love using Foam and I hope to be able to come back with some more targeted pull requests that will be less of a hassle to integrate.

PR #1497, opened 2025-08-03
Support resolving absolute path

wdhongtw:
Consider absolute path to be related to current workspace root, instead of actual path in the filesystem.

Current root is decided by base URI, if any.
Since that we can have multi-root workspace, the root for each resolving process may be different, according to the base URI.

---

The implement try to use one `FoamWorkspace` instance, and inject a `RootChecker` for resolving root folder for some base URI.
This is probably the minimal possible implementation for this feature, but will have some problem like

- The Foam graph feature will merge nodes from all workspace roots into one single graph.

Another potential implementation, is to have multiple `FoamWorkspace` instance, and dispatch resolving process to
corresponding instance based on current context.

Also, the implementation looks a little dirty, by exposing and using `resolveRoot` during placeholder generation and file creation.
If there is any concern around this, please let me know. :D

### DIFF
--- a/packages/foam-vscode/src/core/model/foam.ts
+++ b/packages/foam-vscode/src/core/model/foam.ts
@@ -1,6 +1,6 @@
 import { IDisposable } from '../common/lifecycle';
 import { IDataStore, IMatcher, IWatcher } from '../services/datastore';
-import { FoamWorkspace } from './workspace';
+import { FoamWorkspace, RootChecker, DummyRootChecker } from './workspace';
 import { FoamGraph } from './graph';
 import { ResourceParser } from './note';
 import { ResourceProvider } from './provider';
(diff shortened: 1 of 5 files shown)


wdhongtw:
This PR is related to #791

wdhongtw:
Hi @riccardoferretti might you have a look at this one? 🙂

riccardoferretti:
Hi @wdhongtw thanks for the contribution and sorry for not getting back to you earlier!
This PR is based off some old code, and in particular now on `main` we do have the concept of root folder in `MarkdownProvider`, which I think addresses #791 

Can you please let me know if the problem is still there?
If so, can you please rebase off the latest `main`?

wdhongtw:
Yes, the issue is already resolved in latest version. 😂
Thanks for this good extension and the ecosystem.

Close this PR as there is no action needed.

PR #1512, opened 2025-09-18
update variable-resolver.ts FOAM_DATE_DAY_ISO

ChThH:
Adds
FOAM_DATE_DAY_ISO
to
const knownFoamVariables

This allows use of the Foam function ${FOAM_DATE_DAY_ISO} in templates


riccardoferretti:
@allcontributors add @ChThH for code

PR #1542, opened 2025-10-31
Graph: added option to automatically start graph

meestahp:
Start the foam graph on extension startup.
Add "foam.graph.onStartup": true,  to settings.json

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -701,6 +701,11 @@
           "type": "object",
           "description": "Custom graph styling settings. An example is present in the documentation.",
           "default": {}
+        },
+        "foam.graph.onStartup": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether to open the graph on startup."
         }
       }
     },
--- a/packages/foam-vscode/src/features/panels/dataviz.ts
+++ b/packages/foam-vscode/src/features/panels/dataviz.ts
@@ -3,6 +3,7 @@ import { Foam } from '../../core/model/foam';
 import { Logger } from '../../core/utils/log';
 import { fromVsCodeUri } from '../../utils/vsc-utils';
 import { isSome } from '../../core/utils';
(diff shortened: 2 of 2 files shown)


packages/foam-vscode/package.json
```diff
@@ -701,6 +701,11 @@
           "type": "object",
           "description": "Custom graph styling settings. An example is present in the documentation.",
           "default": {}
+        },
+        "foam.graph.onStartup": {
+          "type": "boolean",
+          "default": true,
```
riccardoferretti:
```suggestion
          "default": false,
```

packages/foam-vscode/src/extension.ts
```diff
@@ -86,6 +87,11 @@ export async function activate(context: ExtensionContext) {
     const foam = await foamPromise;
     Logger.info(`Loaded ${foam.workspace.list().length} resources`);
 
+    const shouldOpenGraphOnStartup = getFoamVsCodeConfig('graph.onStartup');
+    if (shouldOpenGraphOnStartup) {
+      commands.executeCommand('foam-vscode.show-graph');
+    }
+
```
riccardoferretti: The pattern is good, but it should live in the corresponding feature (the `activate` fn of `dataviz.ts`)

meestahp: Whoops, sorry about the package.json boolean, forgot to set that back. I just did a push, moved the startup check to dataviz.ts, not sure if I need to redo the PR or if that's all.

riccardoferretti: I see you copied the code to `dataviz.ts` - now you probably wanna remove it from here ;)

meestahp: ok, I did a rebase and should be cleaned up, unless I'm missing something. My head is screwed on backwards.

riccardoferretti:
@allcontributors add @meestahp for code

PR #1559, opened 2025-12-02
Feature/training notes

Karl-Paul:
# Training Notes Feature

This PR introduces the Training Notes feature to Foam.
It provides a structured way to create, manage, and process notes for efficient learning.

For a full explanation of how the feature works, please refer to the Training Notes [documentation](docs/user/features/training-notes.md).

## Draft Pull Request

I’m sharing it early to get feedback before finalizing the implementation and cleanup. I’ve been working on this feature for quite some time and would appreciate a quick confirmation that the approach aligns with the project’s needs before investing further effort.

## What’s Included So Far

-Core implementation of the Training Notes feature
-Creation of TrainNotes using the template variable
-Object-oriented MarkdownParser<T> structure to support both Notes and TrainNotes
-Raise and Retry Phase commands
-Phase Explorer: tree view with top-level entries for Today, Late, and each phase (1–6), with
 their respective Training Notes listed underneath
-Note Explorer: includes Training Notes

## Remaining Work / Open Points

- Unit tests: fix the failing test issue `TypeError: (0 , test_utils_1.randomString) is not a function` that occurs when using the readFileFromFS or getRandomURI from `test/test-utils.ts` in test files, as well as any other issues that may arise during testing
- Investigate the failing test issue:
Initial research suggests this error can sometimes be caused by circular dependencies.
I already refactored most code paths to rule this out, except for the remaining dependency between foamWorkspace and the provider, which still needs closer examination. However, this did not resolve the error.
I haven't done a full restructuring yet, as this dependency also exists in the main project where the tests run without issues. Still, I’ll revisit this if needed.
- Full feature testing: ensure all parts of the Training Notes workflow function correctly

## Feedback

I’d appreciate any feedback on this draft PR — what works well, what could be improved, or any concerns you see before finalizing the feature.
Additionally, I’d love your input on the feature itself, for example if you notice any issues or points that could be refined.

### DIFF
--- a/packages/foam-vscode/src/core/model/phase.ts
+++ b/packages/foam-vscode/src/core/model/phase.ts
@@ -66,7 +66,7 @@ export class Phases {
     return this.phases;
   }
 
-  public find(name: string, days: number): Phase {
+  public find(name?: string, days?: number): Phase {
     return this.phases.find(
       phase => phase.name === name && phase.days === days
     );
--- a/packages/foam-vscode/src/core/services/markdown-parser.ts
+++ b/packages/foam-vscode/src/core/services/markdown-parser.ts
@@ -131,8 +131,8 @@ export const trainrehydration = (uri: URI, trainnote: TrainNote): TrainNote => {
   trainNote.phases = phases;
   trainNote.type = 'training-note';
   trainNote.currentPhase = phases.find(
-    trainNote.currentPhase.name,
-    trainNote.currentPhase.days
+    trainNote.currentPhase?.name,
+    trainNote.currentPhase?.days
   );
   trainNote.nextReminder =
     trainnote.nextReminder instanceof Date


riccardoferretti:
Hi @Karl-Paul, thanks for taking the time to look into the code and implement this feature.
I understand what you are trying to achieve here, but in all honesty I don't believe this concept should live within Foam, I consider it to be out of scope. Possibly this could be another extension built on top of Foam's API (which can be accessed via VS Code API, see https://code.visualstudio.com/api/references/vscode-api#extensions).
I believe the API should already expose everything you need to implement it as a separate extension, but if there are hooks that might be helpful I am open to discussing exposing them.
It would also be the first extension built on top of Foam, so that makes for an interesting use case!
Let me know your thoughts, and thanks again for looking into this!

Karl-Paul:
Hejo @riccardoferretti,

thanks for your response!
I admit it’s not exactly what I was hoping for to hear that you see this feature as out of scope for Foam, but I appreciate the clarification.
Just to make sure I understood correctly: is one of the reasons that you want to keep Foam focused and avoid it becoming too large or complex? I can imagine that adding a feature like this to the Project might increase maintenance effort and the risk of bugs — is that roughly what you had in mind?

Your suggestion about building this as a separate extension on top of Foam’s API sounds interesting. It could indeed be a nice project — although I’m not entirely sure how much time I’ll realistically have to work on it in the near future. Since most of the source code already exists, I’d mainly need to adapt it to the extension API, but I’m unsure how big the effort would be.

Do you have a rough idea of the complexity or how long such an extension might take to build based on your experience with Foam’s API?
That would help me decide whether it’s feasible for me to continue down that route.

Thanks again for your feedback!

riccardoferretti:
> is one of the reasons that you want to keep Foam focused and avoid it becoming too large or complex?

Yes, you have exactly right. My goal with Foam is to build a solid foundation for a networked knowledge graph data model and exposure in VS Code, and an API that makes building anything on top of it "easy".
Of course a knowledge management system can be used and tailored for a lot of different use cases (zettlekasten, interval training, PARA, ...). Everyone has their own workflow, and in fact I encourage anyone who asks me to eventually develop their own.
I have in fact decided against making my own way of managing my notes first class citizen, and it's instead supported "on top of Foam".

In the case of this feature, the building blocks IMO are available:
- YAML properties are exposed and available in the data model
- via the API it's easy to filter the notes based on those
- and of course VS Code has everything you need to surface those

The one thing that might be missing is better support for updating notes (but actually I think it's mostly there, if not exposed).

So one thing is reducing the number of core features, because I don't want Foam to be opinionated _in that way_.
The second is about maintaining those features. Writing the initial code is about 30% of the effort for the average feature, for 3 reasons:
- any bug discovered in a feature needs to be addressed
- any update to the code will need to be propagated to, and compatible with, all features
- features have a way of demanding more pulling in for users that want their version of those features (e.g. extra options or workflows, ...)

It's hard for me to give you an assessment because the PR also does some refactoring in the core Foam files, which causes a ton of diffs all over the place. In truth my take is that it shouldn't be too complicated to implement just the feature itself, at least V1. I could be wrong because I didn't go through the PR in detail.

Karl-Paul:
Hello @riccardoferretti,
thanks a lot for the detailed clarification — that makes your perspective very clear.
I understand much better now why you want to keep Foam unopinionated and focused on providing solid building blocks and an extensible API. I also think I underestimated the long-term maintenance aspect of this feature.
At the moment I don’t realistically have the time to extract the feature cleanly and rebuild it on top of the API, so I’ll pause on this for now.
Thanks again for taking the time to explain your reasoning so thoroughly — I really appreciate it.

Issue #835, opened 2021-11-19
Graph visualisation: Ability to filter out or apply different decoration (css) to notes from subfolders  


michael-guimet:
On the **Foam Graph**:

A. Filter out subdirectories: 
I would like to filter out (remove) the nodes for notes in particular subdirectories. Ideally with the same behaviour as tags today (UI checkbox activation / deactivation using the name of the folder f.e.)

B. Decorate subdirectory nodes: 
If the above is not possible, I would like to differentiate the nodes for notes in particular subdirectories.

---

**A. Filter out subfolders** 

If we have the folder structure:

```filetree
foamDirectory
--folderA
----note1
--folderB
----note2
----note3
--note4
``` 

When a user runs the command Foam: "Show Graph":
- Initially all the notes and their relationships are shown
- In the Graph filters, in addition to "tags" and "placeholders", users would also see checkboxes for "folderA" and "folderB", initially checked
- The user could uncheck folderA and the graph would remove note1, in the same way it removes tags
- If all checkboxes where unchecked by the user, the note4 would still show, as it is not a tag, placeholder and is not in a subdirectory 

**B. Decorate subdirectory nodes** 

The idea is to be able to define types of notes just by the folder they are in, instead of using the syntax:

```YAML
---
type: feature
---
```
Once the type is defined, css can be applied as it currently is for note types

---
 
I'm happy to elaborate further if this is of interest.

Thanks!

riccardoferretti:
I am curious to see your script @Casyfill, sounds interesting.
The way I was originally thinking about it was to expose that via some sort of configuration/API, so that an actual piece of JS code can "decorate" the nodes (either the notes themselves, or when they are converted into the graph). 
Basically it would be similar to what the python script does, but on the fly (and ephemerally)

clux:
This seems like a very nice feature to have. Have played around with mass modifications through `sed` on all files in a folder to prepend a `type: whatever`, but it causes a lot of header pollution (given that you generally don't need headers in foam), and it feels a bit redundant. If the graphing lib had **some kind of setting** to populate types from `$foldername` then this would be very powerful zero configuration categorisation system (at least for people who use more than one folder).

### Workaround with `sed`
style node type in `foam.graph.style` and mass modify with a one-off script:

```sh
cd recaps
sed -i '1 i\---\ntype: recap\n---\n' *.md
```
..then create your docs from templates with the `type` header set.

janbechtold:
I would also be very interested in this feature.
[...]

Issue #452, opened 2021-01-17
Feat: graph visualization for directory structure


davidde:
I recently discovered foam and I'm really liking it, especially since I have already settled on local plaintext markdown notes. The Roam Research "anti-hierarchy" approach might be great for proprietary solutions that abstract away the file system, but I'm not convinced it is the best approach for working with local files. As such, I am very happy that Foam aims to support both approaches.

I think the graph visualization is really cool, but less useful for people with well-structured note archives, where linking is of secondary importance. My first instinct was to add "toc" files in every directory. However, on second thought this might not be a great idea because of the maintenance burden. I should be able to just write about the stuff that matters without too much boilerplate maintenance.

Thinking some more about this, I believe this might be an ideal feature to support at the extension level; most people with local notes likely already have a directory structure of some sort, since things become unmanageable quickly without. It would be really nice to have built-in support for visualizing this structure. I feel this would bring utility beyond even what current proprietary solutions have to offer.

riccardoferretti:
Hi @davidde, I think there are some interesting ideas there, a few people have mentioned some sort of "index" file per directory, and this feels like something that could be useful, although I am unclear about the shape or how to get to it.

For what regards a graph based on file path, in what way do you consider it to be better than the file explorer that comes with VsCode?

riccardoferretti:
#421 is where we discuss having an index page

davidde:
You're totally right that this kind of overlaps in functionality with the file explorer, and this is also why I was initially hesitant to propose it. Why I eventually did propose it, is because I currently feel like I'm "missing out" on this wonderful visualization tool that is already available. With this feature request I'm trying to gauge interest before spending too much time brainstorming an unwanted feature ;)

Why I feel this would be useful, is because directory hierarchy is actually very similar to linking; it indicates what is related. I usually do not bother to create explicit links where files are already organized by directory, and this means my current graph looks a bit meagre. Having both visualizations in a single graph, with maybe a checkbox to toggle the directory visualization on/off, feels like a winning combination to me.

One idea I had was simply a function that auto-generates these "toc" files I mentioned. A single toc file per directory, which itself would link to all its sibling files, as well as the toc files of its sibling directories. (Alternatively, this data could also be kept in `foam.json` or any other centralized location instead of individual toc files.) This visualization could have a different color to clarify the difference, and the "auto-generate" function could have a boolean config, making it easy to ignore if you don't care for it.

[...]

PR #1567, opened 2026-01-07
[graph] Randomly color nodes by directory

meestahp:
foam colorByDirectory
When checkmark is clicked, the node text will be assigned a random color based on the directory of the node's file. A new setting is exposed to allow the setting default to be changed. 

Relevant Issues: #835, #452

### DIFF
--- a/packages/foam-vscode/package.json
+++ b/packages/foam-vscode/package.json
@@ -756,6 +756,11 @@
           "type": "boolean",
           "default": false,
           "description": "Whether to open the graph on startup."
+        },
+        "foam.graph.appearance.colorByDirectory": {
+          "type": "boolean",
+          "default": false,
+          "description": "When true, colours the node name in the graph with a random colour based on directory"
         }
       }
     },
--- a/packages/foam-vscode/src/features/panels/dataviz.ts
+++ b/packages/foam-vscode/src/features/panels/dataviz.ts
@@ -204,5 +204,9 @@ async function getWebviewContent(
 }
 
 function getGraphStyle(): object {
-  return vscode.workspace.getConfiguration('foam.graph').get('style');
+  const config = vscode.workspace.getConfiguration('foam.graph');
+  return {
+    style: config.get('style'),
+    colorByDirectory: config.get('appearance.colorByDirectory', false),
+  };
 }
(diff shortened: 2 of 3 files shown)


riccardoferretti (commented):
Thanks for the proposal @meestahp, I am open to the option of coloring nodes by directory, but I would apply the following changes to the PR:
- do not expose it as a setting, that part needs to be reworked and I don't want to add tech debt (it means the value will be forgotten between runs, but I am ok with the tradeoff)
- instead of exposing the option as a checkbox, I would add a configuration called "Color by". At first the options can be `none` and `directory`, but I can also see how we could use `type`, `tag`, ... down the line
- Instead of the random color + map, have you considered using a hash of the directory path? it would provide consistency without the need to keep a map in memory
- I am not sure about `__forceUpdate`, why are you doing that?

meestahp:
Ah yes, I anticipated the reluctance to expose it as a setting, the second commit adds that functionality on top of the basic coloring feature. So that should a simple fix.  I will rework the PR. 
The next 2 points: ColorBy dropdown and Directory hashes will take a little time but I'll look into them.
The last one, the force update is because the AFAIK only some node changes trigger a graph update. I'll let the AI explain it.

✦ The line node.__forceUpdate = !node.__forceUpdate; is there to force the graph visualization to re-render when the
  "Directory coloring" option is changed.

  Here's a breakdown of why it's needed:
   1. How the graph library updates: The force-graph library, which appears to be in use here, optimizes rendering by
      only redrawing parts of the graph when it detects changes in the data (nodes and links).
   2. The problem with style changes: When you toggle the "Directory coloring" checkbox, the line
      model.style.colorByDirectory = v; updates a style setting in the model object. However, this change doesn't alter
      the actual node objects within the model.data.nodes array that the graph library is monitoring.
   3. The "dummy property" trick: Because the graph library doesn't see any changes in the nodes, it doesn't know it
      needs to re-render them with the new color setting. The code forces this re-render by looping through each node
      and changing a "dummy" property (__forceUpdate). This modification makes the graph library believe the nodes have
      changed, triggering a redraw.
   4. Applying the new style: During the redraw, the nodeCanvasObject function is executed for each node. This function
      reads the updated model.style.colorByDirectory value and applies the correct coloring.

  In short, it's a workaround to tell the rendering engine that the nodes are "dirty" and need to be redrawn, even
  though their fundamental data (like ID or position) hasn't changed. This ensures the visual style is updated
  immediately when the user interacts with the GUI control.

PR #1639, opened 2026-05-05
Fixing graph focus when using markdown preview

Owen-Exon:
This changes the handleActiveEditorChange trigger and method of opening markdown preview to allow the graph focus to be consistent across files.

### DIFF
--- a/packages/foam-vscode/src/vscode/features/graph-webview/index.ts
+++ b/packages/foam-vscode/src/vscode/features/graph-webview/index.ts
@@ -33,8 +33,8 @@ export default async function activate(
       updateGraph(p, foam);
     };
     const noteUpdatedListener = foam.graph.onDidUpdate(onFoamChanged);
-    const editorListener = vscode.window.onDidChangeActiveTextEditor(e => {
-      handleActiveEditorChange(p, foam, e);
+    const editorListener = vscode.window.tabGroups.onDidChangeTabGroups(() => {
+      handleActiveEditorChange(p, foam);
     });
     p.onDidDispose(() => {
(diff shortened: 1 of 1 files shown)


riccardoferretti:
Thanks for the PR @Owen-Exon 

I am running it locally but it's not working. My steps:
- open foam graph
- open any note
- show the preview for that note
- go back to the graph, make sure no note is selected (click on the background)
- go back to the preview window

I would expect the graph to focus on the note, but it doesn't.
Instead if I go to an editor the graph will focus on the note.

I also tried then navigating from the preview window to another note (staying in preview) and the graph doesn't pick that up either.

Am I missing something?

Owen-Exon:
Sorry, I've just realised that the way that I've "fixed" this has been for the specific way I've been browsing the files.

The reason for this is because I've been attempting to use the graph as a way to navigate the files in a kind of "reading mode" when I don't need to edit. (since the regular edit mode updates as you click through links anyway)

Unfortunately, I couldn't it to work with the default markdown preview (where the source URI seems to be hidden). If you instead open the file from the graph view with `foam.graph.navigateToPreview` set to true, it will update accordingly (and continue updating as you click through links). This also works if you set the `workbench.editorAssociations` for markdown files to `vscode.markdown.preview.editor`

I still feel that it's a partial fix (as it now allows you to open the files in _a_ preview and still update the graph accordingly) but it isn't perfect.

riccardoferretti:
Thanks for the inspiration @Owen-Exon - I have committed a fix that supersedes this PR and it will be available in the next release of the extension

PR #1652, opened 2026-05-22
Updated documentation: Embeds.md

qwloh:
Added a section to indicate the CSS class to target for others who'd like to customize their card embeds like me.

Let me know if the section placement is alright; open to moving my addition further up or down the docs as the maintainers deem fit.

I could also open a PR on the /foam-template repo to add this class in the custom css file that comes with the template. Should I go ahead? And if I so I'll probably rename that file from `custom-tag-style.css` to `custom-styles.css` and come back and update the docs on Tags.md to reflect the updated file name. Checking in before proceeding in case the maintainers team has different preferences/concerns. Thank you!

### DIFF
--- a/docs/user/features/embeds.md
+++ b/docs/user/features/embeds.md
@@ -60,6 +60,27 @@ card![[my-note]] # Bordered container
 inline![[my-note]] # Seamless integration
 ```
 
+#### Custom Card Styling
+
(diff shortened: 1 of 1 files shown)


riccardoferretti (approved):
thanks for this!

riccardoferretti:
@allcontributors add @qwloh for docs

PR #1677, opened 2026-07-09
Fix: Eliminate Startup and Typing Lag for Large Workspaces

ryanncode:
Users with large workspaces can experience prolonged lag (freezing the VS Code UI for several seconds) on startup or while typing.

This occurs because `VsCodeBasedParserCache` persists its data to VS Code's `workspaceState`. Because the cache is stored under a single key, any `cache.set()` operation (triggered by modifying just a single file) causes the debounced `delayedSync` function to dump the entire `LRUCache` (up to 10,000 parsed files). VS Code then serializes this massive JSON payload (which can exceed 50MB) and sends it over the IPC channel to the main UI thread. This chokes the main thread, blocking basic typing and editor functionality.

Foam is optimized to skip `cache.set()` if a file's checksum hasn't changed. For users with small or medium workspaces, the payload is tiny and transfers instantly. Furthermore, if a user rarely modifies files outside of VS Code, startup produces 100% cache hits, meaning `delayedSync` is never even triggered on load. It only becomes a bottleneck for users with massive workspaces, where modifying a file forces the extension to dump and transfer the entire 10k-file cache during the typing debounce.

This PR migrates the parser cache storage away from `workspaceState` and instead uses VS Code's virtual file system API (`vscode.workspace.fs`) to save the payload to `parser-cache.json` in the extension's dedicated `storageUri`.

This is the correct solution because:
1. **Zero IPC Bottleneck:** It offloads massive blob storage to asynchronous background file I/O, completely decoupling the save operation from the VS Code UI thread.
2. **Virtual Workspace Compliant:** By using `vscode.workspace.fs` instead of Node's native `fs`, it adheres to Foam's linting rules and ensures full compatibility with VS Code for the Web and virtual workspaces.
3. **Safe Initialization:** The instantiation was cleanly refactored to an `async` factory method that seamlessly integrates into the existing asynchronous `activate` flow.

All unit tests pass, and this eliminates cache-related freezes.

Feature was built with Antigravity IDE/Gemini Pro 3.1.

### DIFF
--- a/packages/foam-vscode/src/extension.ts
+++ b/packages/foam-vscode/src/extension.ts
@@ -68,7 +68,7 @@ export async function activate(context: ExtensionContext) {
       workspace.createFileSystemWatcher('**/*'),
       workspace.onDidSaveTextDocument
     );
-    const parserCache = new VsCodeBasedParserCache(context);
+    const parserCache = await VsCodeBasedParserCache.create(context);
     const parser = createMarkdownParser([], parserCache);
 
     const notesExtensions = Config.getNotesExtensions();
--- a/packages/foam-vscode/src/vscode/services/cache.ts
+++ b/packages/foam-vscode/src/vscode/services/cache.ts
@@ -1,6 +1,8 @@
 import { debounce } from 'lodash';
 import { LRUCache } from 'lru-cache';
 import { ExtensionContext } from 'vscode';
+import { promises as fs } from 'fs';
+import * as path from 'path';
 import { URI } from '@foam/core';
 import {
   ParserCache,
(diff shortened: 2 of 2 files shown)


riccardoferretti:
Hi @ryanncode - thanks for the PR and great job with diagnosing the problem.
I had a couple of questions/reservations about the PR, in particular about not cleaning the old cache, and having a single cache file that ends up (although in an async way) still with the whole management of the cache. 
So I am going to use your PR as inspiration (will credit you and add you as contributor) and add bucketing/sharding on top of that, and address the other issues.
That should take care of the use case you fixed and also avoid big fs dumps.

Thanks a lot! Will keep this open until I push mine

ryanncode:
Sounds good, I'll probably come back with some more stuff in the next few weeks since I have some free time right now

riccardoferretti:
superseded by #1681

PR #1681, opened 2026-07-09
perf(vscode): shard parser cache storage to fix large-workspace lag

riccardoferretti:
The parser cache was persisted under a single workspaceState key, so every debounced sync serialized the entire cache (up to 10k entries, ~50MB) and shipped it over IPC to the main thread, freezing the UI on large workspaces during startup and while typing.

The cache now persists to 64 bucket files under the extension storageUri. A note maps to a stable bucket via a hash of its URI and only dirty buckets are rewritten, so persistence cost is proportional to the change, not the workspace size. The legacy workspaceState blob is migrated to the new layout on first activation and removed. Pending writes are flushed on shutdown, and clear() can no longer be undone by an in-flight sync.

Supersedes #1677.

### DIFF
--- a/packages/foam-core/src/utils/cache.ts
+++ b/packages/foam-core/src/utils/cache.ts
@@ -3,5 +3,9 @@ export interface ICache<K, V> {
   has(key: K): boolean;
   set(key: K, data: V): void;
   del(key: K): void;
-  clear(): void;
+  /**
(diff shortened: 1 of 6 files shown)


riccardoferretti:
@allcontributors add @ryanncode for code

PR #1687, opened 2026-07-17
Fix extra line breaks between link definitions in vscode plugin

Strepto:
Remove unnecessary line breaks when updating wikilink definitions and add comments for clarity.

AI assisted, human reviewed. Feel free to close if the fix is in the wrong place, but I feel this works well.

### DIFF
--- a/packages/foam-core/src/lint/generate-link-references.ts
+++ b/packages/foam-core/src/lint/generate-link-references.ts
@@ -86,6 +86,7 @@ export const generateLinkReferences = (
         def.range?.start.line === insertLineIndex
     );
     const prefix = isKeptFoamDefinitionAtInsert ? eol : eol + eol;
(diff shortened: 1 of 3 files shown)


riccardoferretti:
Thanks for this - the fix itself is correct.

Before I merge: the second test, orders existing definitions by wikilink occurrence when configured, is failing on the branch and asserts behavior that doesn't exist.
Could you drop that test? The first one is what actually exercises the fix and it passes. 

Once that's done I will be happy to merge, thanks!

Strepto:
Ah. I thought I rolled the sorting back before I committed but I left the test. Sorry. I looked into sorting all the references, but it may add more churn in git conflicts etc, so I dropped it.

riccardoferretti (approved):

Issue #1675, opened 2026-07-01
dose not respect workbench.editorAssociations in vscode user setting


bigheadghost:
### Describe the bug

in vscode user settings.json: 

```json
"workbench.settings.applyToAllProfiles": ["workbench.editorAssociations"],

"workbench.editorAssociations": {
     "*.pdf": "xxx"
}
```

In normal vscode workspce, when pdf file link is clicked,  xxx is started to open it, which is expected.

But  in foam workspace, it is not.

But when the same setting `workbench.editorAssociations` is copied in foam workspace setting `.vscode\settings.json`, it works. It seeems that foam does not respect this key in vscode user settings.json.

### Steps to Reproduce the Bug or Issue

the same as above

### Expected behavior

respect vscode user setting

### Operating System Version

win10 x64

### Visual Studio Code Version

1.126

riccardoferretti:
Thanks for reporting the issue. Could you share a simple repo to reproduce it? 
is this happening with wikilinks or regular links? with or without definitions?
It's very odd that it works if the setting is in the workspace but not at the user level as Foam doesn't distinguish those.

riccardoferretti:
I might have found the culprit. Please verify against the latest version if you still have the problem, thanks!

bigheadghost:
The issue still remains.

It applies to any file extension. let's just take `jpg` files for example, with foam v0.44.2 on vscode v1.127

# preparation

(1) test.md
```markdown
[test](a.jpg)
```

(2) vscode user settings.json:
```json
"workbench.settings.applyToAllProfiles": ["workbench.editorAssociations"],

"workbench.editorAssociations": {
     "*.jpg": "default"
}
```

# test result
1. When foam is disabled, open `test.md` file, control click the link, vscode uses **default** (the editor itself) to open `a.jpg`.
1. When foam is enabled, open `test.md` file, control click the link, vscode dose **not use** default, but the previewEditor.
   - But when `"*.jpg": "default"` setting is copied into foam workspace setting `.vscode\settings.json`, it **uses** default.

bigheadghost:
The issue still remains in foam v0.44.5.  Same test result.

riccardoferretti:
I am very surprised, because for regular markdown links Foam no longer generates the link definition, so what you are getting are the links generated by VS Code. 
Maybe there is a bug in the code, can you double check and confirm that:

```
[test](a.jpg)
```

- only `a.jpg` is a link (not the full text, including the label)
[...]

Issue #1294, opened 2023-10-10
Go to Definition mouse gesture on wikilink doesn't open linked note


irnc:
### Describe the bug

After VS Code upgrade to 1.83.0 behaviour of Go to Definition mouse gesture (cmd+click or option+click) on wikilinks changed.

Prior it opened linked note, now it moves focus to the link definition at Wikilink definitions section.

Issue could be reproduced when `foam.edit.linkReferenceDefinitions` setting is set to `withExtensions`, i.e. Wikilink definitions section are generated. Setting it to `off` and removing the Wikilink definitions restores desired behaviour, Go to Definition now opens the note file.

### Steps to Reproduce the Bug or Issue

1. Open note file with Wikilink defintions section.
2. Use Go to Definition mouse gesture (cmd+click or option+click) on link with a reference in Wikilink definitions section.
3. See that focus moves to the Wikilink definitions section instead of opening the linked note file.

### Expected behavior

As a user, I expect that Go to Definition mouse gesture would open the linked note, even when Wikilink definitions section are present.

### Operating System Version

macOS

### Visual Studio Code Version

1.83.0

riccardoferretti:
Thanks for reporting this @irnc.

What I believe is happening is that VS Code in 1.83.0 is adding a definition to the wikilink (you can see that by putting the cursor is on a wikilink and then running the `Go to definition` command, using the `F12` key).
- Before (and now when there is no wikilink definition section) the only definition was the one added by Foam.
- Now VS Code adds as definition the one of the link at the bottom of the file.

I am not sure why they decided to do so. 
One might say it's correct, as the wikilink is "defined" in that line. On the other hand when you `Go to definition` on a type in a typescript file, you don't got to the `import` statement, but to the resource that defines the type. So if anything I would have expected VS Code to navigate to the file associated to the wikilink definition.

I don't know if there is much that can be done on Foam's side. Maybe there is a way to increase the priority of our definition (`cmd+click` navigates to the first definition). 
This needs a bit of research. Might also be worth raising an issue on the VS Code project.

t2hv33:
- [ ] Is there any configuration or workaround for this issue? @riccardoferretti @irnc

irnc:
I don't know of any configuration that would bring previous behaviour back. As Riccardo said, this needs a bit of research to investigate and resolve the issue.

[...]

PR #1693, opened 2026-07-26
Honor user-level workbench.editorAssociations for attachment links (#1675)

riccardoferretti:
## Problem

#1675: with `workbench.editorAssociations` set in **user** settings (e.g. `"*.jpg": "default"`), ctrl+clicking a link to that file type in a Foam workspace opens the wrong editor (e.g. the built-in image preview). The same setting at **workspace** scope works, and with Foam disabled the user-level setting works too. The fix shipped in 0.44.2 (DocumentLinks with plain file URI targets) did not resolve it — the reporter reproduced the bug on 0.44.2 itself.

## Root cause

- A DocumentLink with a **plain file URI** target is opened via `openerService` → `editorService.openCodeEditor`, which does not go through the same editor resolution as the `vscode.open` command; user-level (profile-propagated) `workbench.editorAssociations` is ignored on that path.
- The built-in Markdown extension opens its links through the **`vscode.open` command** (`extensions/markdown-language-features/src/util/openDocumentLink.ts`), which resolves the editor correctly at every settings scope — that's why everything works with Foam disabled.
- Foam's DocumentLink over the whole `[text](file.jpg)` range also **shadowed** the built-in link (overlapping links are deduped), replacing the working native behavior.

## Changes

- **Wikilinks and reference-style links** to attachments/images now emit a DocumentLink targeting `command:vscode.open` (via the existing `commandAsURI` helper) — the same mechanism the built-in Markdown extension uses. Verified that VS Code's `CommandOpener` revives `$mid`-marshalled `Uri` command arguments, so the target round-trips correctly.
- **Direct markdown links** (`[text](file.pdf)`) emit no Foam DocumentLink at all: the built-in Markdown extension's link is no longer shadowed and handles the click natively (also honoring `markdown.links.openLocation`). Foam still suppresses its Definition for these targets so nothing hijacks ctrl+click.
- **Clickable ranges now match VS Code's markdown link styling**: wikilinks cover only the inner content of `[[...]]` (also for resolved-reference wikilinks, the #1294 case), full reference links `[text][ref]` cover only the reference label, and collapsed (`[text][]`) / shortcut (`[text]`) forms cover the text. Multi-line links fall back to the full range.
- Test infrastructure: added a `DocumentLink` class to the vscode mock and marked `navigation-provider.spec.ts` as `@unit-ready` so it runs under the mock as well as in the extension host.
- Changeset: `foam-vscode` patch (no `@foam/core` changes).

## Testing

- Followed the test-first workflow for each behavior change: updated/added specs (`vscode.open` command targets, no-link-for-direct-markdown-links, bracket-free wikilink ranges, reference-label ranges for all three reference forms), confirmed each failed against the previous implementation, then implemented.
- Full unit suite green; the only failures are two pre-existing environmental ones in `cache.spec.ts` (chmod-based readonly simulation is a no-op when running as root) that reproduce on a clean checkout.
- The real-VS Code e2e run could not execute in the dev environment (VS Code download blocked by proxy) — CI should be the gate for the extension-host run.
- **Recommended manual check**: for the shortcut form `[test]` (+ collapsed/full reference forms), Foam's link range is now identical to the built-in Markdown extension's, so overlap dedup depends on provider ordering — verify in a real window that ctrl+click still opens the target file rather than jumping to the `[ref]:` definition line. Also worth re-running the reporter's matrix (user-level vs workspace-level setting, with a profile + `applyToAllProfiles`) for `[test](a.jpg)`, `[[a.jpg]]`, and a PDF.

Since #1675 was closed with the 0.44.2 fix that turned out to be insufficient, it may be worth reopening it until this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VNA7nU1yjJVsDmo2bbj5h8

---
_Generated by [Claude Code](https://claude.ai/code/session_01VNA7nU1yjJVsDmo2bbj5h8)_


packages/foam-vscode/src/vscode/features/navigation/navigation-provider.spec.ts
```diff
@@ -372,10 +378,19 @@ describe('Document navigation', () => {
...
 
   // Attachments (PDFs, docs, etc.) and images must be surfaced as DocumentLinks
-  // and NOT as definitions, so clicks route through VS Code's URI-opening
-  // pipeline (which honors `workbench.editorAssociations`) rather than the
-  // text-editor pipeline. See: https://github.com/foambubble/foam/issues/1675
+  // targeting the `vscode.open` command, and NOT as definitions. Both the
+  // text-editor (definition) pipeline and plain-URI DocumentLinks bypass the
+  // editor resolution that honors user-level `workbench.editorAssociations`;
+  // only the `vscode.open` command (the same mechanism used by the built-in
+  // Markdown extension) resolves the editor correctly at every settings scope.
```
riccardoferretti:
```suggestion
  // targeting the `vscode.open` command, and NOT as definitions, to honor 
  // user-level `workbench.editorAssociations`.
```

packages/foam-vscode/src/vscode/features/navigation/navigation-provider.spec.ts
```diff
@@ -424,10 +444,14 @@ describe('Document navigation', () => {
       const { doc } = await showInEditor(noteA.uri);
       const provider = new NavigationProvider(ws, graph, parser, tags);
 
+      // No Foam DocumentLink: it would shadow the built-in Markdown link,
+      // which already opens via `vscode.open` and honors
+      // `markdown.links.openLocation`.
```
riccardoferretti:
```suggestion
```

packages/foam-vscode/src/vscode/features/navigation/navigation-provider.spec.ts
```diff
@@ -449,14 +473,81 @@ describe('Document navigation', () => {
 
       const links = await provider.provideDocumentLinks(doc);
       expect(links.length).toEqual(1);
-      expect(links[0].target).toEqual(toVsCodeUri(img.uri));
+      expect(links[0].target).toEqual(openCommandURI(img.uri));
+      // The link range excludes the surrounding brackets, consistent with
+      // how VS Code styles markdown links
```
riccardoferretti:
```suggestion
```

packages/foam-vscode/src/vscode/features/navigation/navigation-provider.ts
```diff
@@ -262,36 +262,94 @@ export class NavigationProvider
...
+    // Wikilinks and reference-style links to attachments (PDFs, docs, etc.)
+    // and images are surfaced as DocumentLinks that invoke the `vscode.open`
+    // command. A DocumentLink with a plain file URI target is opened via the
+    // openerService, which does not honor user-level
+    // `workbench.editorAssociations`; the `vscode.open` command (the same
+    // mechanism used by the built-in Markdown extension) resolves the editor
+    // correctly at every settings scope.
+    // Direct markdown links ([text](file.pdf)) are not covered here: the
+    // built-in Markdown extension already provides a working link for them,
+    // and a Foam link over the same range would shadow it.
```
riccardoferretti:
```suggestion
    // Wikilinks and reference-style links to attachments (PDFs, docs, etc.)
    // and images are surfaced as DocumentLinks that invoke the `vscode.open`
    // command (the same mechanism used by the built-in Markdown extension) 
    // resolves the editor correctly at every settings scope.
    // Direct markdown links ([text](file.pdf)) are handled by the built-in Markdown 
    // extension
```

PR #1704, opened 2026-09-08
Omit embedded frontmatter and preserve content without a title

Sherlock290:
# Omit embedded frontmatter and preserve content without a title

Whole-note embeds currently display YAML metadata as Markdown. `content-*` embeds can discard the first body line of a titleless note or truncate the note after its first section. Setext section titles can leave their underline in the rendered output.

This change omits leading, closed, valid mapping/empty YAML from whole-note embeds and removes only an actual leading ATX/Setext title in content mode. It keeps subsequent sibling sections. Fragment ranges are applied to the original source before any display-only cleanup; block semantics and link rewriting are retained. Unclosed/invalid YAML and prose between thematic breaks are preserved.

For example, embed this child with `full-inline![[Child]]` or `content-inline![[Child]]`:

```markdown
---
aliases: [example]
---
# Child

First body line

# Second

Last body line
```

Expected: metadata is absent, both body paragraphs remain, and only content mode omits the initial title. A titleless child retains its first line.

## Validation

- Before the fix: 21 new regression tests failed; 714 tests passed.
- Final extension unit suite: 737 passed, 5 skipped, 1 todo. The final assertion-only cleanup also passed all 25 content cases.
- Real host: 46 passed (21 existing embed tests plus 25 new content cases).
- CLI: 389 unit tests and 18 end-to-end tests passed.
- Combined with the independent math fix: 53 real-host checks passed, including nested notes with YAML plus inline/display math.

Validation on Linux x64, Node 22.23.2, Yarn 1.22.22; base `97b82e4ed7f363ff9cdbbfab0bb4dd90c7c4d487`.

- `yarn build` passes (including the Node and web extension bundles).
- `yarn workspace foam-vscode compile` passes.
- `yarn workspace foam-vscode lint` completes with 0 errors; the 77 remaining warnings are pre-existing.

The stock `test:e2e` command currently reports **no tests**, with “Runner ... vitest-pool-vscode.ts is not supported” under the locked Vitest 4.1.11, and nevertheless exits 0. I did **not** count that as a pass. For these host checks I used an uncommitted local entrypoint that calls the existing pool directly, adapting its removed `@vitest/spy` `mocks` export in generated output only. Tests ran in the real VS Code 1.110.0 extension host, with an isolated user-data directory and actual VS Code APIs. The runner workaround is not part of this PR.

Includes patch changesets for `foam-vscode` and `@foam/cli`, which bundles this shared preview code. The math integration is a separate, independent PR; this branch targets `main` directly. No private notes or local installation patches are included.

## Submission status

Independent companion PR: #1703. Both branches target `main` directly and can be reviewed or merged separately.

[Upstream CI](https://github.com/foambubble/foam/actions/runs/34244975374) currently reports `action_required` with no jobs started. The local results above are separate from upstream CI; a green CI result has not yet been obtained.

### DIFF
--- a/.changeset/tidy-embeds-preserve.md
+++ b/.changeset/tidy-embeds-preserve.md
@@ -1,0 +1,6 @@
+---
+"foam-vscode": patch
+"@foam/cli": patch
+---
+
+Omit leading YAML metadata from whole-note embeds. Content-only embeds now preserve the first body line of titleless notes and all subsequent sections, and remove complete ATX or Setext titles without shifting section or block references.
--- a/packages/foam-vscode/src/vscode/features/preview/embed-content.test.ts
+++ b/packages/foam-vscode/src/vscode/features/preview/embed-content.test.ts
@@ -79,8 +79,9 @@ describe('Embedded note content', () => {
         expect(html).not.toContain('<hr');
         expect(html).not.toContain('aliases:');
         expect(html).not.toContain('id: 1');
-        if (source.includes('First body line'))
-          expect(html).toContain('First body line');
+        expect(html.includes('First body line')).toBe(
+          source.includes('First body line')
+        );
       });
     });
   }
@@ -92,8 +93,10 @@ describe('Embedded note content', () => {
     it(`keeps titleless content ${JSON.stringify(source)}`, () => {
       withNote(source, render => {
         const html = render('content-inline![[Child]]');
-        if (source) expect(html).toContain('First body line');
-        if (source.includes('Tail')) expect(html).toContain('Tail');
+        expect(html.includes('First body line')).toBe(
+          source.includes('First body line')
+        );
+        expect(html.includes('Tail')).toBe(source.includes('Tail'));
       });
     });
   }
--- a/packages/foam-vscode/src/vscode/features/preview/embed-content.test.ts
+++ b/packages/foam-vscode/src/vscode/features/preview/embed-content.test.ts
@@ -1,0 +1,188 @@
+import MarkdownIt from 'markdown-it';
+import { mkdtempSync, writeFileSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import path from 'path';
+import { FoamGraph, URI, createMarkdownParser } from '@foam/core';
+import { createTestWorkspace } from '../../../test/test-utils';
+import { createFoamMarkdownIt } from './foam-markdown-it';
+
+function withNote(
+  source: string,
+  check: (render: (input: string) => string) => void
+) {
+  const dir = mkdtempSync(path.join(tmpdir(), 'foam-embed-content-'));
+  const uri = URI.file(path.join(dir, 'Child.md'));
+  writeFileSync(uri.toFsPath(), source);
+  const parser = createMarkdownParser();
+  const note = parser.parse(uri, source);
+  const workspace = createTestWorkspace([URI.file(dir)]).set(note);
+  for (const [name, body] of [
+    ['Parent', '---\naliases: [parent]\n---\n![[Child]]'],
+    ['Outer', '![[Parent]]'],
+  ]) {
+    const nestedUri = URI.file(path.join(dir, name + '.md'));
+    writeFileSync(nestedUri.toFsPath(), body);
+    workspace.set(parser.parse(nestedUri, body));
+  }
+  const graph = FoamGraph.fromWorkspace(workspace, false);
+  try {
+    const md = createFoamMarkdownIt(
+      {
+        workspace,
+        graph,
+        parser,
+        linkResolver: ({ resource }) => ({ href: resource.uri.path }),
+        getCurrentResource: () => note,
+      },
+      MarkdownIt({ html: true })
+    );
+    check(input => md.render(input));
+  } finally {
+    graph.dispose();
+    rmSync(dir, { recursive: true, force: true });
+  }
+}
+
+describe('Embedded note content', () => {
+  for (const style of [
+    'full-inline',
+    'full-card',
+    'content-inline',
+    'content-card',
+  ]) {
+    it(`omits metadata and preserves the body in ${style}`, () => {
+      withNote(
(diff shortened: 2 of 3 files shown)


riccardoferretti:
Thanks for this — the bugs are real (YAML rendered in whole-note embeds, `content-*` dropping the wrong line for setext titles and titleless notes), and the test cases you added genuinely fail on `main`. I ran them against both branches. But I can't merge it in its current shape: the new title/frontmatter detection regresses cases the old code handled, and it adds parallel implementations of helpers the repo already has. Details below, all reproduced locally with the harness from your test file.

### Regressions in content mode

**1. The title survives whenever anything precedes it** (`stripLeadingTitle`, `wikilink-embed.ts:316`)

The old code found the note's first heading via `note.sections[0]`. The new code only inspects markdown-it's *first* block token. So any note with something above the title keeps the title:

```md
<!-- note -->
# Title

Body
```

`content-inline![[Child]]` — before: `<p>Body</p>`; now: the comment, `<h1>Title</h1>`, and the body. Same for a `#tag` line, a badge image, a blockquote, or a BOM-prefixed file (`﻿# Title` now renders as a literal paragraph `# Title`).

Related: a section heading nested in a blockquote or list (`> # Quoted heading`) is a real `Section` starting at the container line, so `content-inline![[Child#Quoted heading]]` now renders the heading; the old `rows.shift()` removed it.

**2. Frontmatter that `yaml` rejects, or that isn't a mapping, is kept — and then the title is kept too** (`wikilink-embed.ts:334`)

Foam's parser (remark-frontmatter) treats any leading `---` block as frontmatter and removes it from `sections`/`title` regardless of YAML validity. The preview now disagrees with the parser. For tab-indented YAML, duplicate keys, `title: A: B`, a sequence, or plain prose between fences:

```md
---
tags:
	- a
---
# Title

Body
```

the parser reports `sections=[Title@line 4]`, but `content-inline![[Child]]` renders `<hr>`, a setext `<h2>` of the YAML text, *and* `<h1>Title</h1>` — because once the frontmatter is kept, the first token is an `hr` and `stripLeadingTitle` bails. Before the PR content mode rendered just `Body`.

**3. An unresolved fragment still renders the note with its YAML** (`wikilink-embed.ts:354`)

`![[Child#Missing]]` or `![[Child#^deleted]]` falls through both branches: the fragment is truthy so the new `else` never runs, and `contentExtractor` skips the title strip too. Output is `<hr><h2>id: 1</h2><h1>Title</h1>…` while `![[Child]]` of the same file hides the YAML. Pre-existing, but the changeset promises whole-note embeds omit metadata and this fallback *is* a whole-note embed.

To make these concrete, `main` now has six regression tests for #1 and #2 in `wikilink-embed.test.ts` (`content mode strips the title using the parsed sections`): comment and badge before the title, YAML the parser rejects, non-mapping YAML, a heading nested in a blockquote, and a BOM-prefixed note. They pass on `main` and all six fail on this branch, so a rebase will show them directly.

### Structural issues

**4. The fix is at the wrong layer.** The actual gap is that the inner markdown-it instances built by `createFoamMarkdownIt`/`buildInner` have no front-matter rule — the outer preview only works because VS Code's own `md` is passed in as `target`. Stripping inside one branch of `fullExtractor` leaves every other inner render broken: the HTML export report (`render-report.ts:178`) renders raw note markdown through the same pipeline and shows `<hr>` + a setext `<h2>` of the YAML + a duplicated H1 today. A front-matter block rule in `createFoamMarkdownIt` (or a core "body without frontmatter" accessor used by every consumer) fixes embeds, the report, and #3 at once.

**5. Duplicated helpers.** `@foam/core` already exports `stripFrontMatter` (`foam-core/src/utils/md.ts:28`, used by `services/editor.ts` and the query engine's `body`/`content` fields). Title removal already exists as `stripH1Title` (`query/index.ts:403`), `stripTitleHeading` (`render-report.ts:211`) and `stripLeadingH1` (CLI starlight exporter). This PR adds a fourth of each with different rules, so a `foam-query` `content` cell and a `content![[…]]` embed of the same note now produce different bodies. See AGENTS.md: "Reuse existing helpers and constants instead of adding parallel ones."

**6. The `@foam/cli` changeset entry isn't justified.** The CLI doesn't bundle any foam-vscode preview code (`packages/foam-cli/esbuild.js` bundles only `src/index.ts` with `@foam/core` and `@foam/mcp`). The paired-bump rule applies to `@foam/core` changes only. As written it would publish a CLI version whose changelog describes behaviour the binary doesn't have. Please drop that line.

### Smaller points

- `contentExtractor` (`:380`) now does two full-document parses per embed (remark in `withLinksRelativeToWorkspaceRoot`, then markdown-it in `stripLeadingTitle`) plus a YAML parse, where the old code parsed only the section slice once. This is the hot path #1375 tracks.
- `contentExtractor` (`:385`) re-runs `Resource.findSection` to guess which branch `fullExtractor` took; the two conditions have to be kept in sync by hand. The `!first.map` guard is unreachable (markdown-it always sets `map` on `heading_open`), and the comment about avoiding recursive embed rendering doesn't apply to `parse()`.
- Tests: loop-generated names are the stringified fixture (`keeps titleless content ""`), and several expectations are computed from the input (`expect(html.includes(x)).toBe(source.includes(x))`), so the empty-source case asserts nothing. `withNote` writes Parent/Outer notes and builds a `FoamGraph` for all 25 cases though one uses them. Literal `toContain`/`not.toContain` under behaviour names, in `wikilink-embed.test.ts`, would be clearer.

### Suggested direction

I have added tests on `main` that will fail on your branch. 
Rebase, keep your tests, and add a case for #3. Then, instead of a second Markdown parser in the extractor:

- strip frontmatter once at the pipeline level (or via one core helper) so all inner renders agree with the parser's notion of frontmatter;
- remove the title using the heading/section ranges the `Resource` already carries, in original-source coordinates, before link rewriting;
- apply the same fallback to unresolved fragments.

That's a smaller change than this one and it removes the regressions rather than adding special cases. Happy to review a revised version.

PR #1708, opened 2026-09-19
Fix the foam-vscode e2e suite, and what it was hiding

riccardoferretti:
The `foam-vscode` integration suite stopped running on 2026-07-09 and CI stayed
green for ten weeks. This turns it back on, adds the guard that makes the next
silent breakage loud, and fixes what it had been hiding.

## What broke

`a8d74abb` (#1680) widened **only** foam-vscode's vitest range to `^4.1.10`; the
other four packages kept `^3.0.0`. That commit's own metadata records
`dependency-version: 3.2.6`, so the advisory it resolved is satisfied by `^3.x` —
the major bump was gratuitous. Vitest 4 removed `pool: <path to a .ts file>` and
`poolOptions`, so the custom extension-host pool was rejected for all 38 spec
files.

## Why nobody noticed

`suite.ts` decided success by filtering `getFiles()` for `result.state === 'fail'`.
With no file run at all that array is empty, so nothing threw and `test:e2e`
exited 0.

## The guard

`getE2eRunFailure` now owns the verdict and also fails on an empty file list, on
files that collected no test case, and on `state.getUnhandledErrors()` — where
the pool rejection actually landed. It is a pure function, unit tested without an
extension host.

The runner also prints `Foam e2e: ran N tests across M spec files`. That line
earned its place during this work: two runs looked green only because the
extension host had died partway through and `vscode-test` still exited 0. The
guard cannot catch that — it never executes — but the missing count is obvious.

## Vitest

Pinned to `^3.2.6`, and the other four packages raised to match: one range, one
lockfile entry, one hoisted install. Vitest 4 replaced the pool API with
`pool: PoolRunnerInitializer`, whose `PoolWorker` is an RPC channel;
reimplementing that in-process is heavy machinery for a pool whose whole purpose
is to avoid workers, and it would leave foam-vscode as the only package on v4.

## What the suite was hiding

17 failing tests across 4 spec files. Most were real product bugs.

**`instanceof URI` across module copies** — `asAbsoluteUri` and three
`FoamWorkspace` methods take `URI | string` but narrowed on `instanceof URI`.
That fails for a URI made by another copy of the module, so a valid URI fell down
the string branch: `path.startsWith is not a function` in `create-note`,
`reference.split is not a function` in folder renames. The discriminator belongs
on the string. This alone fixed 10 of the 17.

**Folder renames** (`refactor.ts`):

- The rename removed each note from the index and re-read it from disk. The read
  is async and nothing awaited the handler, so the note was indexed under neither
  path for a while — the disappearance #1699 set out to fix. A directory move
  leaves contents untouched, so the entry is now re-keyed in place.
- The files whose wikilinks a rename rewrites were never re-indexed, so renaming
  the same folder twice found no backlinks the second time.
- `onWillRenameFiles` and `onWillDeleteFiles` did async work VS Code never
  awaited: a listener's own promise is not a participant, only what it passes to
  `waitUntil` is. Both are now registered through `waitUntil`.
- `onWillDeleteFiles` statted a path without catching, so deleting something
  already gone became an unhandled rejection that took the extension host down
  mid-run.

**Four test defects.** Flagging these explicitly, since three involved changing
an expectation:

- `create-note` asserted that VS Code snippet syntax survives untouched. It does
  not and never did — content is inserted with `insertSnippet`, and
  `docs/user/features/templates.md` documents snippet-variable support. It was a
  characterization test written while the suite was dark, and it mischaracterized
  the behavior it meant to pin. Corrected to the documented behavior rather than
  changing the product.
- `scoped-watcher` asserted on filesystem-watcher events with no wait, so it only
  proved the event had not arrived — which also made its negative assertion pass
  for free.
- `convert-links` used `cleanWorkspace()` without a timeout, so files were deleted
  but not de-indexed and a stale `note-a.md` made `[[note-a]]` ambiguous.
- The `deleteFile` helper removed directories with `fs.delete`, which fires no
  file-operation event, leaving every note under them indexed for the rest of the
  run.

`waitForEmptyFoamWorkspace` now names the resources still indexed; the bare count
gave nothing to go on.

## Verification

- `test:e2e` on Vitest 4 with the guard: exit 1 naming the 38 pool rejections
  (was exit 0).
- `test:e2e` on Vitest 3: **355 tests across 38 spec files**, exit 0, stable over
  four consecutive full runs, zero unhandled rejections. Previously 0 tests ran.
- `yarn test:unit` across all five packages: green.
- `yarn lint`: 0 errors, warning counts unchanged.

## For the reviewer

- `refactor.ts` looks larger than it is: two handlers were extracted to named
  functions so they could be passed to `waitUntil`, so most of the diff is
  re-indentation. `git diff -w` reduces it to ~65 lines.
- The convert-links isolation problem was order- and timing-dependent. Stability
  was confirmed over four consecutive full runs, but the shared-extension-host
  design (`fileParallelism: false`, one workspace) is where to look first if CI
  shakes out a different interleaving.
- The last commit only touches comments the earlier commits introduced, and is a
  reasonable squash candidate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


riccardoferretti:
@claude review

riccardoferretti:
@claude implement the fixes

riccardoferretti:
@claude review

riccardoferretti:
@claude look at the review and decide what to address right away

riccardoferretti:
@claude review

packages/foam-vscode/src/vscode/features/editing/refactor.ts
```diff
@@ -32,143 +33,198 @@ export default async function activate(
...
-
-        if (!syncLinks) {
-          continue;
-        }
+  /**
+   * Rewrites the wikilinks that point at what is being renamed, and records
+   * what a directory rename is about to move.
+   *
+   * Registered through `waitUntil`: VS Code does not await a listener's own
+   * promise, so otherwise this would not finish before the rename lands.
```
riccardoferretti:
```suggestion
```

packages/foam-core/src/model/workspace.ts
```diff
@@ -422,11 +417,13 @@ export class FoamWorkspace implements IDisposable {
   }
 
   public find(reference: URI | string, baseUri?: URI): Resource | null {
-    if (reference instanceof URI) {
+    // Narrowed on the string so a URI from another copy of this module does
+    // not fall through to the identifier branch below.
```
riccardoferretti:
```suggestion
```

packages/foam-core/src/model/uri.ts
```diff
@@ -469,7 +473,9 @@ export function asAbsoluteUri(
   if (baseFolders.length === 0) {
     throw new Error('At least one base folder needed to compute URI');
   }
-  const path = uriOrPath instanceof URI ? uriOrPath.path : uriOrPath;
+  // Narrowed on the string: `instanceof URI` rejects a structurally valid URI
+  // that came from another copy of this module.
```
riccardoferretti:
```suggestion
```

packages/foam-core/src/model/uri.ts
```diff
@@ -101,7 +103,9 @@ export class URI {
   }
 
   resolve(value: string | URI, isDirectory = false): URI {
-    const uri = value instanceof URI ? value : URI.parse(value, 'file');
+    // Narrowed on the string, as in `asAbsoluteUri`: `instanceof URI` rejects a
+    // structurally valid URI that came from another copy of this module.
```
riccardoferretti:
```suggestion
```

packages/foam-core/src/model/uri.ts
```diff
@@ -69,11 +69,13 @@ export class URI {
     if (!match) {
       return new URI();
     }
-    defaultScheme =
-      defaultScheme instanceof URI
-        ? defaultScheme.scheme
-        : (defaultScheme as string | null);
-    const scheme = match[2] || defaultScheme;
+    // Narrowed on the string, as in `resolve`: `instanceof URI` rejects a
+    // structurally valid URI that came from another copy of this module.
```
riccardoferretti:
```suggestion
```

packages/foam-core/src/model/workspace.test.ts
```diff
@@ -96,6 +96,26 @@ describe('Workspace resources', () => {
   });
 });
 
+describe('URIs from another copy of the module', () => {
```
riccardoferretti: instead of talking about another copy of the module I wonder if we should just say that we support URI-like objects (that have a `path` field). Thoughts?

packages/foam-core/src/model/uri.test.ts
```diff
@@ -64,6 +64,29 @@ describe('Foam URI', () => {
     });
   });
 
+  it('resolves a URI that is not an instance of this copy of the module', () => {
+    // Same shape and behavior, different class identity — what a URI built by a
+    // second copy of this module looks like.
```
riccardoferretti: Same comment as below, talk about URI like objects instead of a specific case?

```suggestion
```

riccardoferretti:
@claude check out the two reviews since your last update

